PR #1, opened 2026-04-30
chore: sync upstream - CONFLICTS need manual review

TerrysPOV:
Automated sync from moazbuilds/claudeclaw:master has merge conflicts. Resolve before merging.

PR #2, opened 2026-05-01
docs: introduce ClaudeClaw+ identity

TerrysPOV:
## Summary

- Full README replacement: ClaudeClaw+ identity, upstream attribution (moazbuilds + contributor credits), daily-sync explanation, 3-way comparison table, feature descriptions for PRs #71-75 #77 #144 with upstream PR links, FAQ, contributors section
- New CONTRIBUTING.md: welcomes large PRs and opinionated changes, issue-first workflow, plugin bump requirements, Sentrux note, sync-PR explanation
- Updated .github/pull_request_template.md: open scope, no size limit, CI-required bump block
- Plugin manifest rebrand: `claudeclaw` → `claudeclaw-plus`, owner `moazbuilds` → `TerrysPOV`, version `1.0.4` → `2.0.0`

## Test plan

- [ ] View README on GitHub and confirm formatting, badges render, contrib.rocks images load
- [ ] Confirm no broken links (upstream PR links, docs links)
- [ ] Confirm `sync-upstream.yml` still runs cleanly after merge (it touches src/, not docs)

PR #21, opened 2026-05-03
feat(memory): session-history search via FTS5 + sentence-transformers (#19)

### LINKED ISSUE #19: feat(memory): FTS5 + embedding semantic search over session history
## Background

ClaudeClaw-Plus has a persistent memory system. A complementary layer — searchable session history — would give Claude recall over past work without loading entire sessions into context.

## Proposed approach (from @Nibbler1250)

Index all session JSONL files into SQLite with two search layers:
1. **FTS5** — fast exact keyword match
2. **Sentence-transformer vectors** — semantic similarity (`all-MiniLM-L6-v2`, ~80MB)

Combined weighted hybrid score: `alpha * vector_score + (1-alpha) * fts5_score`

For Claude Max users (no separate API key), summaries can pipe through the `claude --print` CLI rather than the SDK:
```python
result = subprocess.run(
    ["claude", "--print", "--model", "claude-haiku-4-5"],
    input=summary_prompt, capture_output=True, text=True
)
```

@Nibbler1250 has a working ~200-line Python implementation ready to contribute.

## Why ClaudeClaw-Plus

- Orthogonal to (but complementary with) Plus's existing MEMORY.md system
- Provides unlimited searchable history without blowing up context window
- The infrastructure complexity (SQLite, embeddings) fits Plus's scope

## Source

Requested by @Nibbler1250 in moazbuilds/claudeclaw#14 (Mega-Post) with a working proof-of-concept.

Nibbler1250:
Resolves #19. Adds hybrid keyword + semantic search over Claude Code session JSONL history, exposed as deterministic read/write entry points on `src/memory.ts` (per your option **(c)** preference in #19).

## Architecture (option C from #19)

```
src/memory.ts          ← TS extension, deterministic R/W entry points
  ├─ indexSessions()   ← spawn Python indexer, return counts
  └─ searchSessions()  ← spawn Python search CLI, return parsed JSON

tools/memory-search/   ← Python core (heavy lifting kept here)
  ├─ memory_search/db.py     SQLite + FTS5 + sentence-transformers
  ├─ memory_search/recall.py search + claude --print summarizer
  ├─ memory_search/cli.py    argparse CLI used by the TS layer
  ├─ memory_search/config.py env-var-driven configuration
  └─ tests/test_db.py        pytest smoke tests (6/6 passing)

skills/session-recall.md     ← skill markdown, plug-and-play
```

## Determinism

Every `searchSessions()` call shells out to a fresh Python subprocess that reads from the on-disk SQLite DB. No in-process cache — a write from a concurrent `indexSessions()` is visible immediately on the next read.

`indexSessions()` is idempotent: the indexer skips JSONL files whose mtime hasn't changed since the last index pass. Cheap to call on every boot.

## Wired to your asks from #19

| Your answer | This PR |
|---|---|
| Option (c) — extension to `src/memory.ts` | ✅ |
| Boot trigger | ✅ `src/commands/start.ts` calls `indexSessions()` after `ensureUserSymlinks()`. Wrapped in try/catch — never blocks startup. |
| Periodic re-index every N turns | ⏳ Saved for follow-up PR (touches `runner.ts` / `agents.ts`, kept out of this PR per your scope guidance) |
| Configurable embedding model | ✅ `MEMORY_SEARCH_MODEL` env var, defaults to `all-MiniLM-L6-v2` |
| First PR scope: indexer + skill | ✅ |

## Configuration

All knobs are env-var driven so the tool works on any project layout. Defaults match Claude Code's default `~/.claude/projects/` location:

| Env var | Default |
|---|---|
| `MEMORY_SEARCH_DB` | `~/.claude/claudeclaw/memory-search.db` |
| `MEMORY_SEARCH_SESSIONS_DIR` | All subdirs of `~/.claude/projects/` |
| `MEMORY_SEARCH_MODEL` | `all-MiniLM-L6-v2` |
| `MEMORY_SEARCH_ALPHA` | `0.5` (hybrid blend) |
| `MEMORY_SEARCH_CHUNK_SIZE` | `5` |
| `MEMORY_SEARCH_SUMMARY_MODEL` | `claude-haiku-4-5` |
| `CLAUDECLAW_CLAUDE_BIN` | `which claude` |

The `Settings` interface accepts an optional `memorySearch` block that maps to the same knobs (intentionally typed loosely in this PR — happy to add a strict TS type to `config.ts` in a follow-up if you want it baked in).

## Install

```bash
cd tools/memory-search
pip install -e .
```

That installs the `memory-search` console script and registers the package. The TS layer in `src/memory.ts` uses `spawnSync` and falls back from `tools/memory-search/.venv/bin/memory-search` → PATH → graceful no-op.

## Tested locally

- 41 sessions / 427 chunks indexed against my own `~/.claude/projects/` history
- Hybrid search returns relevant sessions (e.g. `"voice pipeline asterisk"` correctly surfaces the May 2 setup session as #1 hit)
- 6/6 pytest tests green (mocked vectorize so tests don't load the real model)
- `bunx tsc --noEmit` is clean for both `src/memory.ts` and the new lines in `src/commands/start.ts`. Two pre-existing TS errors in `start.ts` (lines 499, 504) are unrelated to this PR — they're present on `main` (their line numbers just shift by +15 due to the boot-trigger insertion).

## Out of scope (intentionally — for follow-ups)

- N-turn periodic re-index trigger (touches `runner.ts`)
- Strict TS type for `memorySearch` settings in `config.ts`
- Agent-system-prompt integration so Claude proactively recalls
- Multi-file MEMORY.md index pattern (separate issue you said you'd assign)

Marked as draft so you can shape it before final review. Happy to slice differently if any of the above should land in this PR instead.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.4",
+      "version": "2.0.5",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.3",
+      "version": "2.0.4",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.0.3",
+  "version": "2.0.4",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.0.4",
+  "version": "2.0.5",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/skills/session-recall.md
+++ b/skills/session-recall.md
@@ -1,0 +1,75 @@
+---
+name: session-recall
+description: Hybrid keyword + semantic search over past Claude Code session history. Use when the user asks "what did we do about X", "you remember when", "was there a fix for Y", "earlier we talked about Z", or any callback to a prior conversation that isn't in the current session.
+trigger:
+  - "qu'est-ce qu'on avait fait"
+  - "tu te souviens"
+  - "c'était quoi le fix"
+  - "on en avait parlé"
+  - "retrouve-moi"
+  - "what did we do about"
+  - "was there a discussion"
+  - "earlier we"
+  - "remember when"
+---
+
+# session-recall
+
+Hybrid FTS5 + sentence-transformer search over the JSONL session history that
+Claude Code persists under `~/.claude/projects/`. Surfaces past sessions
+relevant to the current question without loading their full text into context.
+
+## When to use
+
+- The user references something they remember discussing but you don't have
+  it in the current context window.
+- You need to check whether a problem was already solved in a prior session.
+- The user asks for a status summary that spans multiple past sessions.
+
+## How to use
+
+```bash
+memory-search recall "<query>" --top 5
+```
+
+Add `--no-summary` for a faster keyword/vector match without the LLM
+summarization step.
+
+For programmatic use inside skills:
+
+```python
+from memory_search import recall
+result = recall("yfinance crumb fix", top_k=5)
+for hit in result["hits"]:
+    print(hit["score"], hit["timestamp"], hit["excerpt"])
+print(result["summary"])
+```
+
+## Examples
+
+```bash
+memory-search recall "BREAKOUT_52W backtest results"
+memory-search recall "fix yfinance crumb"
+memory-search recall "Asterisk SRTP codec negotiation"
+memory-search recall "IBKR Gateway zombie session"
+```
+
+## Reindex (when needed)
+
+The indexer is idempotent (skips files whose mtime is unchanged) so it's safe
+to run on every daemon start or on a cron.
+
+```bash
+memory-search index           # incremental
+memory-search index --force   # full reindex (e.g. after model change)
+memory-search stats           # quick health check
+```
+
+## Notes
+
+- The first invocation downloads the sentence-transformer model (~80 MB).
+- Vector search alone returns up to 50 candidates; FTS5 is layered on top
+  for keyword anchoring and the two are blended with `alpha` (default 0.5).
+- Search results are deduplicated per session: the top chunk wins.
+- The `--summary` step uses `claude --print`, so it works without an
+  Anthropic API key (relies on Claude Code's existing OAuth session).
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -17,7 +17,7 @@ import { initializeJobSystem } from "../orchestrator/resumable-jobs";
 import type { Job } from "../jobs";
 import { isWizardTrigger, hasActiveWizard, handleWizardInput } from "./plugin-wizard";
 import { PluginManager, setPluginManager } from "../plugins";
-import { indexSessions } from "../memory";
+import { indexSessionsBackground } from "../memory";
 
 const CLAUDE_DIR = join(process.cwd(), ".claude");
 const HEARTBEAT_DIR = join(CLAUDE_DIR, "claudeclaw");
@@ -340,20 +340,6 @@ export async function start(args: string[] = []) {
   // dev. Idempotent and non-destructive.
   try {
     const links = await ensureUserSymlinks();
-  // memory-search: incremental session re-index on boot (issue #19).
-  // Idempotent (skips files whose mtime is unchanged), so it is cheap to
-  // run every time. Runs synchronously to keep the deterministic R/W
-  // contract; failures are logged but never block daemon startup.
-  try {
-    const memSettings = (await loadSettings()) as { memorySearch?: any };
-    const idx = indexSessions(memSettings.memorySearch);
-    if (idx.indexed > 0) {
-      console.log(`[memory-search] indexed ${idx.indexed} new session(s) (${idx.skipped} up-to-date)`);
-    }
-  } catch (e) {
-    console.log(`[memory-search] index skipped on boot: ${(e as Error).message}`);
-  }
-
     const now = new Date().toLocaleTimeString();
     if (links.created.length > 0) {
       console.log(`[${now}] Installed ${links.created.length} user symlink(s): ${links.created.join(", ")}`);
@@ -365,6 +351,18 @@ export async function start(args: string[] = []) {
     }
   } catch (err) {
     console.error(`[${new Date().toLocaleTimeString()}] ensureUserSymlinks failed:`, err);
+  }
+
+  // memory-search: incremental session re-index on boot (issue #19).
+  // Async fire-and-forget — must NOT block the Bun event loop on first boot
+  // (model download + indexing of N sessions could otherwise freeze Telegram /
+  // Discord polling for tens of seconds). Top-level try so failures here are
+  // never confused with symlink-installation failures.
+  try {
+    const memSettings = (await loadSettings()) as { memorySearch?: any };
+    indexSessionsBackground(memSettings.memorySearch);
+  } catch (e) {
+    console.log(`[memory-search] background index could not be scheduled: ${(e as Error).message}`);
   }
 
   // Phase 17: migrate any Phase 16 single-job agents from flat dir into agents/<name>/jobs/default.md
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -17,6 +17,7 @@ import { initializeJobSystem } from "../orchestrator/resumable-jobs";
 import type { Job } from "../jobs";
 import { isWizardTrigger, hasActiveWizard, handleWizardInput } from "./plugin-wizard";
 import { PluginManager, setPluginManager } from "../plugins";
+import { indexSessions } from "../memory";
 
 const CLAUDE_DIR = join(process.cwd(), ".claude");
 const HEARTBEAT_DIR = join(CLAUDE_DIR, "claudeclaw");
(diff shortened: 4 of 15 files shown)


TerrysPOV (changes requested):
Found 4 issues:                              
                                                                                                                                                                                                                                              
  ---                                                    
  1. Summarizer prompt is hardcoded in French — all English users get French output
                                                                                                                                                                                                                                              
  recall.py was lifted almost verbatim from @Nibbler1250's French-language setup. The summarizer prompt and format_for_humans output strings are in French:
                                                                                                                                                                                                                                              
  prompt = (                                             
      f'Voici des extraits de sessions passées pertinentes pour la requête : "{query}"\n\n'                                                                                                                                                   
      ...                                                
      + "\n\nRésume en 3-5 points concis..."                                                                                                                                                                                                  
  )                                                                                                                                                                                                                                           
                                               
  And in format_for_humans: '🔍 Recherche : "{result["query"]}"'                                                                                                                                                                              ─
                                                         
  Any English-speaking user gets a French narrative summary. Must be fixed to English (or made locale-configurable) before merge.                                                                                                             
  
  https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/620152ae90c830e7681ba56c64b3e796a524a1a6/tools/memory-search/memory_search/recall.py#L41-L51                                                                                              
                                                         
  ---                                                                                                                                                                                                                                         
  2. spawnSync with a 10-minute timeout blocks the Bun event loop on boot
                                                                                                                                                                                                                                              
  indexSessions() uses spawnSync — synchronous, blocking the entire event loop while it runs. On first boot (or after a model change), this downloads the ~80MB embedding model + indexes potentially hundreds of sessions. The daemon will be
   completely unresponsive to Telegram and Discord messages for the duration.

  The "deterministic R/W" justification in the PR description doesn't apply to the boot trigger — no concurrent reader is waiting for it.

  Fix: use spawn (async) at boot, fire-and-forget, log when done:
  // boot trigger — non-blocking
  spawn(resolveBin(opts), ["index"], { env: buildEnv(opts), stdio: "ignore" })
    .then(() => console.log("[memory-search] boot index complete"))
    .catch(e => console.log("[memory-search] boot index failed:", e.message));

  searchSessions() can remain spawnSync since it's explicitly invoked by a skill and the caller expects to wait.

  https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/620152ae90c830e7681ba56c64b3e796a524a1a6/src/memory.ts#L131-L150

  ---
  3. Memory-search try block is nested inside the ensureUserSymlinks() try block

  In start.ts, the memory-search boot trigger is indented inside the outer try that wraps ensureUserSymlinks(). If that outer try-catch has different error-handling semantics (it does — the symlinks failure path differs from the index failure path), exceptions from inner code can be swallowed by the wrong catch.

  The memory-search block should be a separate top-level try block immediately after the symlinks one, not nested inside it.

  https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/620152ae90c830e7681ba56c64b3e796a524a1a6/src/commands/start.ts#L340-L358

  ---
  4. reindexEveryNTurns is in the TS interface but completely unwired

  MemorySearchSettings.reindexEveryNTurns is documented as "Re-index after every N runner turns. Default: 10. 0 disables." — but there is zero wiring in runner.ts or anywhere else. A user who sets this config value gets no effect. Either remove the field from the interface until the follow-up PR lands, or mark it /** NYI — follow-up PR */ so it doesn't mislead integrators.

  https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/620152ae90c830e7681ba56c64b3e796a524a1a6/src/memory.ts#L98-L99

  ---
  Shaping / steering notes

  These aren't blocking but worth raising with the contributor:

  - uv not pip — install instructions say pip install -e .. TerrysPOV tooling preference is uv. Change to uv pip install -e . or better, provide a uv-managed lockfile under tools/memory-search/.
  - Loose typing — as { memorySearch?: any } is acknowledged as intentional. A follow-up to add a proper MemorySearchSettings block to config.ts should be tracked as a concrete TODO, not left open-ended.
  - Installation UX — right now the user has to manually cd tools/memory-search && uv pip install -e . before the feature works. Consider adding a check in indexSessions() that detects the binary is missing and logs a clear setup hint rather than a generic error.
  - The withSummary flag defaulting to false in searchSessions() but the skill telling users to call memory-search recall without --no-summary — these should be consistent. Decide whether summary-on-by-default or summary-off-by-default is the right UX and align both layers.

  ---
  Overall verdict: The architecture is excellent and the test isolation (monkeypatched vectorizer, tmp_path DB) is exactly right. Fix the two P0s (French strings, blocking spawnSync at boot) and the structural nesting bug in start.ts before marking ready for review. The reindexEveryNTurns field is a minor cleanup. Once these changes are made, I'm happy for you to move this out of Draft and submit for review. Thanks for your work!

Nibbler1250:
Thanks for the careful review @TerrysPOV — all four bumped to fixed in `210a541`.

## Walk-through

| # | Issue | Fix |
|---|-------|-----|
| 1 | French strings in `recall.py` | Translated `summarize()` prompt and all `format_for_humans()` user-facing strings to English. Locale-configurable variant deferred — happy to add as a follow-up if `MEMORY_SEARCH_LOCALE` env var feels worth the surface. |
| 2 | `spawnSync` blocking the Bun event loop on boot | Added `indexSessionsBackground()` using `child_process.spawn` (fire-and-forget). Original `indexSessions()` (sync) kept for `searchSessions` callers that explicitly wait. Boot path uses the async variant. |
| 3 | memory-search nested in `ensureUserSymlinks` try | Moved out — now a top-level `try` immediately after the symlinks try-catch. Errors here can't be swallowed by the symlinks catch anymore. |
| 4 | `reindexEveryNTurns` documented but unwired | JSDoc now explicitly says `NYI — wired in a follow-up PR`. Kept the field on the interface so the planned follow-up doesn't have to re-add it. |

## Shaping notes also addressed

- ✅ `README` now uses `uv pip install -e .` (and `uv pip install -e ".[dev]"`) per your tooling preference. The setuptools-based `pyproject.toml` is uv-compatible out of the box; happy to add a uv-managed lockfile in a follow-up if you want lockfile reproducibility shipped.

## Still open from the shaping notes (not addressed in this push)

- **Strict `MemorySearchSettings` type in `config.ts`** — explicitly tracked as a follow-up. Will open a small PR right after this one lands.
- **Binary-missing setup hint** — agreed, will add to `indexSessionsBackground` (and the sync variant) so callers get a one-line "run `cd tools/memory-search && uv pip install -e .`" message instead of a generic spawn error. Can fold into this PR if you'd prefer.
- **`withSummary` default consistency** — agreed it's inconsistent. Right now TS defaults to `false`, the skill instructs `--no-summary` is opt-in. Easy fix: flip the skill to call `recall` without the flag, since users on a normal install have the `claude` CLI available. Will fold into this PR if you'd prefer it not slip to a follow-up.

## Verification

- 6/6 pytest still green.
- `bunx tsc --noEmit` clean for the two files this PR touches. The two pre-existing `forwardToTelegram`/`forwardToDiscord` arity errors at `start.ts:497` and `:502` live on `main` already — left untouched.
- Smoke-tested `memory-search recall "voice" --top 1` against the local 41-session DB; output is now English (`🔍 Search: ...`, `✅ 1 session(s) found:`, etc.).

Re-marking ready for review.

Nibbler1250:
Follow-up: the three shaping notes are now addressed. Pushed as three small, self-contained commits on top of the Round 2 fixes so they're easy to revert independently if you'd rather take any of them as a separate PR.

| # | Shaping note | Commit | Approach |
|---|--------------|--------|----------|
| 1 | Strict `MemorySearchSettings` type in `config.ts` instead of `as { memorySearch?: any }` | [`0120706`](https://github.com/TerrysPOV/ClaudeClaw-Plus/pull/21/commits/0120706) | New `parseMemorySearchSettings()` exported from `memory.ts`, mirrors the `parseWatchdogConfig` pattern. Per-key validation drops invalid fields (booleans, non-empty strings, `alpha ∈ [0,1]`, integers ≥ 0). `Settings.memorySearch` is now non-optional (`{}` when unset) — caller in `start.ts` drops the cast. |
| 2 | Binary-missing setup hint instead of generic spawn ENOENT | [`9ebab21`](https://github.com/TerrysPOV/ClaudeClaw-Plus/pull/21/commits/9ebab21) | Replaced `resolveBin()` with `checkMemorySearchBin()` returning a discriminated union `{ ok: true; bin } \| { ok: false; reason }`. Detection order preserved (explicit `binPath` → project venv → `which memory-search`). Missing binary now logs: `[memory-search] binary not installed (looked at <path> and PATH). Run: cd tools/memory-search && uv pip install -e .` — `indexSessions`/`searchSessions` throw with that hint, `indexSessionsBackground` logs and returns (boot path must not crash). |
| 3 | `withSummary` default consistency between TS and skill | [`8ed6a31`](https://github.com/TerrysPOV/ClaudeClaw-Plus/pull/21/commits/8ed6a31) | The Python CLI defaults to summary-ON (`--no-summary` is the opt-out) and `skills/session-recall.md` documents that. Flipped the TS wrapper to match: `withSummary` defaults to `true`, `--no-summary` is appended only when callers explicitly pass `false`. No call-sites in the repo pass `withSummary` today, so existing behaviour is unchanged — just the contract of future TS callers is brought back in line. |

## Verification

- `tsc --noEmit` clean for the three touched files (`memory.ts`, `config.ts`, `start.ts`). The two pre-existing `forwardToTelegram`/`forwardToDiscord` arity errors at `start.ts:497`/`502` still live on `main`, untouched.
- 6/6 pytest still green (no Python changes in the follow-ups).
- Smoke-tested the missing-binary path locally by pointing `binPath` at a non-existent file: log line is `[memory-search] binary not found at configured binPath '/tmp/nope'. Run: cd tools/memory-search && uv pip install -e .` as expected.

Happy to split any of these out into their own PR if you'd prefer to land them separately, or to drop the `which memory-search` PATH lookup in #2 if you'd rather keep the binary search strictly in-repo.

TerrysPOV:
@Nibbler1250 — two bugs were caught by Codex autoreview after we merged your work into PR #23. Flagging them here in case they affect your personal fork or any copies you're running.

---

**Bug 1 (P1) — FTS5 rank normalization is inverted in `db.py`**

In `search()`, the FTS5 phase normalises scores with:
```python
fts_scores[rowid] = (rank - lo) / span
```

SQLite FTS5 `rank` is BM25-derived — **lower (more negative) values mean a better match**. With the original formula, the best-matching chunk gets score `0.0` and the worst gets `1.0`, which flips the hybrid blend and pushes relevant sessions *down*, especially in keyword-heavy mode (low `alpha`).

Fix (applied in #23):
```python
fts_scores[rowid] = 1.0 - (rank - lo) / span
```

---

**Bug 2 (P2) — Binary discovery is POSIX-only in `memory.ts`**

The default venv path and PATH fallback are hardcoded for POSIX:
```typescript
const MEMORY_SEARCH_BIN = join(MEMORY_SEARCH_DIR, ".venv", "bin", "memory-search");
// ...
const which = spawnSync("which", ["memory-search"], ...);
```

On Windows the venv uses `.venv\Scripts\memory-search.exe` and the shell command is `where` (which can also return multiple lines). The fix:
```typescript
const MEMORY_SEARCH_BIN =
  process.platform === "win32"
    ? join(MEMORY_SEARCH_DIR, ".venv", "Scripts", "memory-search.exe")
    : join(MEMORY_SEARCH_DIR, ".venv", "bin", "memory-search");

const whichCmd = process.platform === "win32" ? "where" : "which";
const which = spawnSync(whichCmd, ["memory-search"], { encoding: "utf8" });
if (which.status === 0 && which.stdout.trim()) {
  return { ok: true, bin: which.stdout.trim().split("\n")[0].trim() };
}
```

---

Both fixes are in the combined PR #23. Thank you again for a really solid implementation — the architecture (Python tool + TS entry points) held up well under review.

Nibbler1250:
@TerrysPOV — thanks for the heads-up. Both bugs applied to my fork on `feat/memory-search` ([`c33c826`](https://github.com/TerrysPOV/ClaudeClaw-Plus/pull/21/commits/c33c826)) since I am running this branch personally.

| # | Bug | Fix |
|---|-----|-----|
| 1 (P1) | FTS5 rank normalization inverted (`db.py`) — BM25 lower-is-better was mapped to 0.0 instead of 1.0, flipping the hybrid blend | Changed to `1.0 - (rank - lo) / span` so best chunk → 1.0, matching the vector cosine convention used downstream. |
| 2 (P2) | Binary discovery POSIX-only (`memory.ts`) — `.venv/bin/memory-search` + `which`, neither works on Windows | Branched on `process.platform === "win32"` for both the binary path (`.venv\Scripts\memory-search.exe`) and the PATH lookup command (`where` instead of `which`), and take the first line of `where` output. |

Glad the architecture held up — that's exactly the trade-off I was hoping for by isolating the heavy lifting in Python. And appreciate the Codex catch on bug 1 in particular; that flip would have been a nightmare to track down via search-quality regression alone.

Let me know if you want me to close this PR (since #23 carries the merged + fixed version), or leave it open as a record.

PR #25, opened 2026-05-05
feat(memory): wire reindexEveryNTurns periodic re-index in runner (#19 follow-up)

Nibbler1250:
The MemorySearchSettings.reindexEveryNTurns option was parsed and exposed
in the public interface but marked NYI. This wires it into the runner so
the session DB stays current as conversations grow.

Changes:
- src/runner.ts: after each successful turn (in the existing turn-tracking
  block, post-incrementTurn), if turnCount % reindexEveryNTurns === 0,
  fire indexSessionsBackground(memSettings). Fire-and-forget — never
  blocks the runner. Logs the trigger at the same point as the existing
  Turn count log line.
- src/memory.ts: drop the 'NYI - wired in a follow-up PR' marker from
  reindexEveryNTurns JSDoc; document where it is wired and the
  fire-and-forget semantic.

Behavior:
- Default 10 turns (per Terry's spec in #19).
- 0 disables.
- Single getSettings() call per turn-end (no double-read).
- Re-uses the existing indexSessionsBackground() async variant; the Python
  indexer is mtime-aware so repeated calls are cheap (skips unchanged
  session files).

Tested via tsc --noEmit: no new errors in runner.ts or memory.ts.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.7",
+      "version": "2.0.8",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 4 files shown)


Nibbler1250:
Hey @TerrysPOV — just a gentle ping in case this slipped through. This PR is the per-turn re-index follow-up I committed to in #19 (the `reindexEveryNTurns` wiring we discussed). It's small (+15/-6), mergeable/clean, and completes the last piece of the #19 scope. Let me know if you'd like any changes before merging.

PR #28, opened 2026-05-07
feat(telegram): opt-in live tool-call streaming progress

### LINKED ISSUE #13: feat(telegram): opt-in live tool-call streaming progress view
## Background

Currently Telegram shows nothing while Claude is working, then delivers the complete response. Users have no visibility into whether Claude is actively doing work or stuck.

## Proposed solution

Add an opt-in config key `telegram.streamToolCalls: true` (default: `false`).

When enabled:
1. As Claude executes tool calls, send/edit a "working" message showing the current tool name (e.g. `⚙️ Running: Bash`, `⚙️ Reading: src/runner.ts`)
2. When Claude finishes, delete/replace the working message with the final clean response
3. Tool call details are ephemeral — they don't clutter chat history

Builds on the existing `stream-json` execution mode already in ClaudeClaw-Plus which exposes tool call events.

## Config

```json
{
  "telegram": {
    "streamToolCalls": true
  }
}
```

**Must be opt-in** — streaming adds Telegram API calls and some users explicitly prefer clean output only.

## Source

Requested by @mrcharlesiv in moazbuilds/claudeclaw#14 (Mega-Post) with working screenshots.

Nibbler1250:
## Summary

Implements #13 — opt-in live tool-call progress indicator in Telegram.

When `telegram.streamToolCalls: true` is set in `settings.json`, each tool call Claude makes is surfaced as a live Telegram message that updates in-place:

1. First tool call → sends `⚙️ ToolName…` as a new message
2. Subsequent tool calls → edits that message to show the current tool
3. Claude finishes → deletes the progress message, sends the final response

### Changes

**`src/runner.ts`**
- `runClaudeStream`: add optional `onToolCall?: (toolName: string) => void` parameter; fires for each `tool_use` block in `assistant` stream events
- Thread `onToolCall` through `execClaude` → `run` → `runUserMessage` (all optional, fully backward-compatible)
- Only the primary `runClaudeStream` call receives `onToolCall`; fallback/stale-recovery retries do not (no duplicate notifications)

**`src/config.ts`**
- `TelegramConfig`: add `streamToolCalls?: boolean` (default `false`, strict opt-in via `=== true`)
- `DEFAULT_SETTINGS` and `parseSettings` updated accordingly

**`src/commands/telegram.ts`**
- New helpers: `sendMessageAndGetId`, `editProgressMessage`, `deleteProgressMessage`
- `handleMessage`: when `streamToolCalls` is enabled, build an `onToolCall` callback that chains Telegram API calls via a serial `Promise` to avoid race conditions between rapid tool calls

### Configuration

```json
{
  "telegram": {
    "token": "...",
    "streamToolCalls": true
  }
}
```

### Design notes

- **Opt-in, off by default** — existing deployments are unaffected
- **Serial Promise chain** — `progressOp = progressOp.then(...)` prevents races if two tool calls fire in quick succession
- **Always cleaned up** — `await progressOp` + delete runs on both success and error paths
- **Non-blocking** — the `onToolCall` callback fires synchronously in the stdout reader loop but Telegram API calls are fire-and-forget from Claude's perspective

Closes #13

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.42",
+      "version": "2.2.43",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.42",
+  "version": "2.2.43",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -475,6 +475,52 @@ async function sendTyping(token: string, chatId: number, threadId?: number): Pro
   }).catch(() => {});
 }
 
+async function sendMessageAndGetId(
+  token: string,
+  chatId: number,
+  text: string,
+  threadId?: number,
+): Promise<number | null> {
+  try {
+    const resp = await callApi<{ ok: boolean; result: { message_id: number } }>(
+      token,
+      "sendMessage",
+      {
+        chat_id: chatId,
+        text,
+        ...(threadId ? { message_thread_id: threadId } : {}),
+      },
+    );
+    return resp.result?.message_id ?? null;
+  } catch {
+    return null;
+  }
+}
+
+async function editProgressMessage(
+  token: string,
+  chatId: number,
+  messageId: number,
+  text: string,
+): Promise<void> {
+  await callApi(token, "editMessageText", {
+    chat_id: chatId,
+    message_id: messageId,
+    text,
+  }).catch(() => {});
+}
+
+async function deleteProgressMessage(
+  token: string,
+  chatId: number,
+  messageId: number,
+): Promise<void> {
+  await callApi(token, "deleteMessage", {
+    chat_id: chatId,
+    message_id: messageId,
+  }).catch(() => {});
+}
+
 async function sendDocumentToChat(
   token: string,
   chatId: number,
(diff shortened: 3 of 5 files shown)


Nibbler1250:
Hey @TerrysPOV — quick check before I rebase this. Moaz upstream merged `/verbose` mode (commit `1b3cbe7`, 2026-05-03) which surfaces tool calls in Telegram via append-style lines above the text response. Different UX from this PR's edit-in-place opt-in progress message, but same goal.

Two options:
1. **Keep this PR**: opt-in `telegram.streamToolCalls` + edit-in-place progress (one updating message, deleted on completion). Cleaner UX for users who only want "what's it doing right now".
2. **Skip and inherit on next sync**: Plus pulls Moaz `/verbose` whenever you next sync upstream. Per-chat toggle, runs alongside the text response.

Both end up surfacing tool calls; the experiences are different. Want me to rebase #28, or close it in favor of the upcoming Moaz sync?

TerrysPOV:
Good question — no conflict. The two features use separate callback slots in the call chain:

- `/verbose` (already in Plus via upstream sync) uses `onToolEvent` — appends full tool event lines above the response, stays visible permanently, toggled per-chat
- `streamToolCalls` (this PR) uses `onToolCall` — fires just the tool name, edits a single ephemeral message in-place, deletes it when Claude finishes

The rebased branch passes both independently so they can coexist without interfering. A user with both enabled would see verbose tool lines building up in the response *and* a separate ephemeral indicator — unusual but not broken.

The experiences are genuinely different enough to justify keeping both: `/verbose` is a full audit trail of what tools were called; `streamToolCalls` is a lightweight "currently doing X" indicator that leaves no trace. Good fit for Plus.

One thing worth a follow-up issue: `streamToolCalls` as a global `settings.json` flag is slightly inconsistent with `/verbose` being a per-chat toggle. A `/streamtools` command (same pattern as `/verbose`) would let individual users opt in without requiring a daemon restart. Not a blocker for this merge — just worth tracking.

TerrysPOV:
### Code review

Found 1 issue:

1. Progress message leaks permanently if `runUserMessage` throws (timeout, network error, rate-limit exception). The cleanup sequence — `await progressOp` + `deleteProgressMessage` — sits inline after the call with no `try/finally`. `progressMsgId` is scoped to the `else` block so the outer `catch` in `handleMessage` cannot reach it. The `⚙️ ToolName…` message stays in the chat permanently on any error path.

Fix: wrap in a `try/finally`:
```ts
try {
  result = await runUserMessage(..., onToolCall);
} finally {
  await progressOp;
  if (progressMsgId !== null) {
    await deleteProgressMessage(config.token, chatId, progressMsgId);
    progressMsgId = null;
  }
}
```

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/f62c9983cdaf8bc24eee7b3c265dca27d3a81c54/src/commands/telegram.ts#L1358-L1368

Nibbler1250:
Pushed the try/finally fix in `f93de51`. Sorry for the delay.

Branch is well behind main now (245 commits) — happy to rebase if you want me to before re-review, or leave that to you. Lmk.

Nibbler1250:
Rebased on top of `main` (force-push to `cca5dfc`). Squashed the 5 original commits down to one clean commit since main's runner.ts had refactored substantially since this branch was cut — re-applying the `onToolCall` threading on top of the new shape was cleaner than resolving line-by-line. Try/finally cleanup is included.

Diff is now 5 files, +101/-15:
- `src/config.ts` — `streamToolCalls` field on `TelegramConfig` + default + parse
- `src/runner.ts` — `onToolCall` threaded through `runClaudeStream` → `execClaude` → `run` → `runUserMessage`
- `src/commands/telegram.ts` — `sendMessageAndGetId` / `editProgressMessage` / `deleteProgressMessage` helpers + `handleMessage` opt-in branch with the try/finally cleanup
- `.claude-plugin/{plugin,marketplace}.json` — bumped to 2.2.43

Ready for re-review.

TerrysPOV:
Hey @Nibbler1250 — thanks for the rebase work today, and especially for flagging the upstream overlap back in May. You called the right shape: option 2 turned out to be the right call. The Moaz `/verbose` work landed via sync PR #29 (May 7) so we already have tool-call display, `/kill`, `/fork`, and the streaming machinery in main.

The edit-in-place UX you built here is genuinely nicer for the "what's running right now?" case — but maintaining both is more cost than the difference earns, so going to close this with thanks. The `onToolCall` threading approach is sound and we may revisit it under the Sprint 3/4 Bus adapter when we get to richer streaming UIs.

Really appreciate you doing the second-opinion check on the spec (PR #106) and on this — that's the loop working as designed. Closing now; if I've missed anything from your rebase that's NOT in the upstream path, give me a shout and I'll reopen.

PR #34, opened 2026-05-09
security: 4 fixes (2 Critical / 2 High) — draft for review

Nibbler1250:
## Summary

Internal security review surfaced 4 issues. Each fix is its own atomic commit so you can review/revert independently.

| # | Severity | Fix | Commit |
|---|---|---|---|
| 1 | 🔴 Critical | Sandbox `[send-file:]` directive — was accepting arbitrary paths, allowing exfiltration of any file readable by the bot user (SSH keys, `.env`, credentials) via a forwarded message. Now constrained to `~/.claude/agent/outbox` with `realpath` containment, extension allowlist, and 20MB cap. | b77228e |
| 2 | 🔴 Critical | Redact `api` / `fallback.api` Anthropic tokens in `/api/technical-info` (the raw token was being returned to anyone with Web UI access). | 8589993 |
| 3 | 🟠 High | Require Bearer auth on 6 state-mutating Web UI endpoints (`/api/chat`, `/api/jobs/*`, `/api/settings/heartbeat`, `/api/kanban`). CSRF alone wasn't blocking same-machine or LAN access to endpoints that spawn `claude --dangerously-skip-permissions`. Bearer placed before CSRF, mirroring the existing `/api/inject` pattern. | 64bbeb7 |
| 4 | 🟠 High | Same sandboxing applied to `[voice:]` directive — extracted as helper `validateOutboxPath()` reused between `send-file` and `voice`. | 9f6bc35 |

## Testing

Manually validated locally: forwarded `[send-file:/etc/passwd]` is now refused; `/api/technical-info` no longer leaks the api key. No regressions introduced by these changes.

## Why draft

We exercised these on our own setup, but the threat model surface for community installs is broader than what we can test in-house — different deployment shapes (containerized, public-LAN, multi-user) may reveal corner cases we didn't hit. Looking for your eyes on:

- Whether `OUTBOX_DIR = ~/.claude/agent/outbox` is the right canonical location, or if Plus already has a convention we should align with
- Whether the extension allowlists (`.md .txt .json .log .csv .png .jpg .pdf` for send-file, `.ogg .mp3 .wav` for voice) cover legitimate use cases
- Edge cases on the Bearer middleware order — we mirrored `/api/inject` but you may want different placement for `/api/chat`
- Anything we missed in scope (other directives accepting paths? other endpoints we should have flagged?)

Happy to tighten/loosen scope, add tests, or split into separate PRs based on your read. Marked draft so there's no merge pressure while we discuss.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.10",
+      "version": "2.0.13",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.0.10",
+  "version": "2.0.13",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.github/workflows/bun-smoke-test.yml
+++ b/.github/workflows/bun-smoke-test.yml
@@ -1,0 +1,30 @@
+name: Bun Smoke Test
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+jobs:
+  smoke-test:
+    name: Parse & build check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Bun build (parse + type check)
+        run: |
+          bun build src/index.ts \
+            --outdir /tmp/claw-smoke \
+            --target bun \
+            --no-splitting \
+            2>&1
+          echo "✓ Build clean — no duplicate declarations or parse errors"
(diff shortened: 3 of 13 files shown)


TerrysPOV:
Good work — the four fixes are solid and this addresses real attack surface. I've reviewed, pushed a maintainer patch, and am merging this.

**Maintainer patch (5e11be4)** fixed two issues before merge:

1. `checkBearer` was returning 503 (not 401) when `apiToken` is unset — which would have broken `/api/chat`, `/api/jobs/*`, `/api/settings/heartbeat`, and `/api/kanban` completely on any install that hasn't set an API token. Changed to `return null` when no token is configured, so Bearer auth is opt-in and falls through to CSRF-only protection. Existing behaviour preserved for unconfigured installs.

2. `OUTBOX_DIR` was initialised via a module-level IIFE that called `mkdirSync` at import time — a permission error or bad path would crash the daemon on startup rather than at directive use. Replaced with a lazy `getOutboxDir()` that memoises on first call and degrades gracefully.

**Your "Why draft" questions:**

- *`~/.claude/agent/outbox` as the canonical location* — good choice. It's already the intended agent output staging area in Plus. No change needed there.
- *Extension allowlists* — the lists are reasonable for the current use cases. If you or other users hit legitimate cases that are blocked (e.g. `.zip` for compressed exports, `.svg`), open a follow-up PR to extend them — easy to review in isolation.
- *Bearer before CSRF on `/api/chat`* — correct ordering. Bearer is a stronger gate and should short-circuit before CSRF to avoid unnecessary token validation on unauthenticated requests. The pattern mirrors `/api/inject` intentionally.
- *Anything missed in scope* — the only other directive worth checking is `[image:]` if it exists, but I don't see it in the current codebase. If there are future directives that accept paths, the `validateOutboxPath` helper is already in place to reuse.

Merging now. If you want to extend the extension allowlists or add the TOCTOU hardening (re-verify with `realpathSync` immediately before the actual file read/send), those are welcome as follow-up PRs — they don't need to block this.

PR #41, opened 2026-05-09
feat(skills-tuner): SkillsSubject + companion /tuner skill — directly addresses #14 (3/3)

Nibbler1250:
## Summary

Stacked on PR2. This PR ships **`SkillsSubject`** — the first concrete `TunableSubject`, **directly addressing #14**: detect repeated patterns or unmatched frustrations and propose capturing them as new skills with Telegram inline approval (boutons via PR2's `TelegramAdapter`).

It also ships:
- The **`/tuner` companion skill** (`src/skills-tuner-templates/tuner.md`, 493 LOC, 6 modes) — itself a `TunableSubject` member, demonstrating recursive self-improvement with safeguards (`risk_tier: critical`, dead-man's switch, 30-day cooldown).
- The **CLI** (`commander`-based, 9 commands).

This PR exists primarily as a **canonical example** of how to write a subject and how to integrate one with Claude Code's skill system. Other surfaces (voice lexicon, RAG weights, MCP config, ML hyperparameters) will follow the same pattern.

## Direct mapping to #14

#14 asks for:

> *"When Claude notices it is performing the same multi-step sequence repeatedly, it could propose capturing it as a skill — then ask the user to approve or reject via Telegram inline buttons before writing anything."*

Implementation in this PR:

| #14 ask | Implementation |
|---|---|
| "notices same multi-step sequence repeatedly" | `SkillsSubject.collectObservations()` scans Claude Code session JSONLs over a configurable window |
| "propose capturing it as a skill" | When ≥ `orphan_min_observations` (default 2) frustration signals match no existing skill, the subject emits a `Cluster` with `subjects_touched=[ORPHAN_SUBJECT]`. `propose_change()` returns a `Proposal` with `kind="new_skill"` |
| "Telegram message with inline buttons [Save / Skip / Edit first]" | `TelegramAdapter` from PR2 renders proposal cards with the inline keyboard |
| "On approval: writes the skill file and registers it" | `apply()` dispatches on `kind in CREATE_KINDS` → creates `.md` file in `scan_dirs[0]` (slug from alternative label, collision → timestamp suffix). `validate()` enforces frontmatter has triggers (anti-loop guard) |
| "On rejection: notes the skip to avoid re-proposing" | The framework writes `pattern_signature` to `refused.jsonl` with TTL 30d (PR1 mechanism) |

The auto-capture flow is the literal #14 ask. The framework primitives from PR1 + PR2 made it almost free to implement.

## Bonus beyond #14 — improvement of existing skills

[...]

PR #40, opened 2026-05-09
feat(skills-tuner): Plus adapters + ExternalProcessSubject ML connector (2/3)

Nibbler1250:
## Summary

Stacked on PR1 (`feat: skills-tuner core engine`). This PR adds:

1. **Plus-native adapters** — Telegram inline keyboard (the literal user-facing flow specified in #14), CLI, and a `PlusEventAdapter` stub that becomes a thin wrapper once #31 (plugin → MCP bridge) lands.
2. **`ExternalProcessSubject`** — a language-agnostic stdio JSON-RPC adapter that lets any external process (Python Optuna, Rust solver, etc.) implement a `TunableSubject` without core changes. **This is the ML connector** that keeps the framework native TS while still supporting ML-driven subjects.

## What's included

- `src/skills-tuner/adapters/cli.ts` — stdout pretty-print of proposals.
- `src/skills-tuner/adapters/telegram.ts` — inline keyboard A/B/C + Refuse + Edit buttons. `allowedUserIds` enforced at constructor time (throws if missing). Optional `verifyProposalFn` for callback HMAC re-check.
- `src/skills-tuner/adapters/plus_event.ts` — stub for the Plus event bus (collapses to ~20 LOC of native calls if #31 lands).
- `src/skills-tuner/adapters/base.ts` — common `CallbackPayload` / `FeedbackResponse` types.
- `src/skills-tuner/subjects/base.ts` — `BaseSubject` with frontmatter + scan helpers.
- `src/skills-tuner/subjects/external_process.ts` — the language-agnostic subject adapter.
- `examples/external_subject_python_template.py` — documented Python template implementing the JSON-RPC contract.

## Direct mapping to #14 — Telegram approval flow

#14 specifies the user-facing flow:

> *"Sends a Telegram message: 'I noticed I do X repeatedly. Want me to save this as a skill?' with inline buttons: [✅ Save] [❌ Skip] [✏️ Edit first]"*

PR3 ships the detection + proposal logic. **This PR ships the surface** that lets #14's user interaction happen:

| #14 ask | This PR provides |
|---|---|
| Sends a Telegram message | `TelegramAdapter.renderProposal()` |
| Inline buttons [Save / Skip / Edit] | Inline keyboard from `proposal.alternatives` (A/B/C) + Refuse + Edit |
| On approval: writes the skill file | callback handler verifies HMAC, calls `subject.apply(proposal, alternativeId)`, logs `event=applied` to audit |
[...]

PR #39, opened 2026-05-09
feat(skills-tuner): core engine — TunableSubject framework + security + storage + git_ops (1/3)

Nibbler1250:
## Summary

This PR adds the **core engine** of `skills-tuner` — a modular human-in-loop tuner that addresses #14 (skill self-improvement via Telegram inline approval) by treating skills as just one type of `TunableSubject` in a broader self-improvement framework.

It is the first of three stacked PRs:
- **PR1 (this one)** — core framework: types, engine, security, storage, git ops. No subjects, no integrations.
- PR2 — adapters (Telegram + claude_cli backend + Plus event stub) + `ExternalProcessSubject` (the ML connector for Optuna/sklearn/etc via stdio JSON-RPC).
- PR3 — `SkillsSubject` (concrete answer to #14) + companion `/tuner` skill + CLI.

All written in pure TypeScript (Bun runtime) for native Plus integration. Lives under `src/skills-tuner/`.

## What's included

- `src/skills-tuner/core/` — `TunableSubject` ABC, `Proposal`/`Cluster`/`Observation` types (zod-validated), `Engine` orchestrator, `Registry`, HMAC-signed proposal primitives, append-only audit log, observation sanitization (XML and bracket markers), YAML config loader, claude_cli + anthropic_api LLM backends.
- `src/skills-tuner/storage/` — JSONL stores for proposals + refused signatures (TTL 30d) + schema migration scaffolding.
- `src/skills-tuner/git_ops/branches.ts` — branch isolation per proposal (`tune/proposal-{id}`, branched from `main`), commit + revert helpers.
- `src/skills-tuner/scripts/migrate-from-python.ts` — one-shot migration from the Python reference implementation.
- `src/skills-tuner/README.md` — local readme + architecture diagram.

## Security model — built in from day one

8 defense layers, all in this PR:

1. **HMAC-SHA256 signed proposals** — every proposal signed with a per-user secret (`~/.config/tuner/.secret`, mode 0600). `applyProposal` refuses any proposal whose signature doesn't verify (timing-safe comparison). Defends against tampering and replay.
2. **Append-only audit log** — `~/.config/tuner/audit.jsonl` records every lifecycle event. Never truncated.
3. **Observation sanitization** — strips zero-width chars + neutralizes both XML (`<system>`) and bracket-style (`[system]`) prompt-injection markers.
4. **Path traversal guards** — `realpath()` + containment check before any write. Enforced at multiple layers (subject + BranchManager).
[...]

PR #42, opened 2026-05-09
feat(plugins): MCP bridge — universal plugin+Plus substrate (closes #31)

### LINKED ISSUE #31: feat(plugins): bridge to expose plugin tools via MCP — design discussion
## Context

The current `src/plugins.ts` lets daemon plugins hook **around** Claude Code invocations (lifecycle events, services, commands, channel runtime) — great for cross-cutting concerns like memory, telemetry, or prompt augmentation.

What it does *not* allow: a plugin registering a **tool** that Claude Code can call during a session. Tools today must come from MCP servers, configured separately in `.mcp.json` / `mcpServers`. This means an extension that wants both behaviours has to ship two artifacts:

- a daemon plugin (events, services)
- a separate MCP server (callable tools)

with duplicated config, secrets, and lifecycle. The integration story for "an extension that does both" is rougher than it needs to be.

## Concrete cases where this hurts

- **Archive/RAG plugin** — wants to persist tool results on `tool_result_persist` *and* expose `archive_search` for Claude to query. Currently two processes.
- **Trading dashboard plugin** — wants to read `agent_end` for telemetry *and* expose `quote(symbol)` / `buy(symbol, qty)` tools. Two processes.
- **Voice orchestrator** — wants `message_received` *and* an `inject_prompt` tool. Two processes.

## Why this matters in practice

The real friction isn't the runtime cost of two processes — it's that **adding any custom tool today means patching core**.

The current path for someone running Plus with their own ecosystem (custom RAG, trading rig, voice stack, home-automation, whatever) is:
1. Fork or maintain a long-lived overlay
2. Hand-thread the new tool through `runner.ts` / `telegram.ts` / wherever it integrates
3. Either upstream a PR (slow, requires the maintainer to take on opinionated custom logic that may not generalise), or hold a permanent diff against `master` and re-resolve conflicts on every upstream sync

This makes Plus feel less extensible than it actually is. A user with five custom tools has either five upstream PRs (none of which Terry/Moaz should reasonably want to merge if they're domain-specific) or a maintenance burden every time `master` moves. The lifecycle plugin API solved exactly this problem for *events* — the same shape applied to *tools* would let users ship their own stack as standalone packages without ever touching core.

## Security and compliance angle

Patching core to add tools is not just a maintenance smell — it is a security posture problem, and one that gets worse the closer you get to regulated environments (finance, healthcare, gov). A plugin tool layer would close several gaps that the current event-only API already half-addresses:

- **Centralised audit surface.** Today, a custom tool patched into `runner.ts` or `telegram.ts` is invisible to any auditor scanning for "what callable surface does Claude have in this deployment?". With a registry, every tool has a known origin (plugin id, version, source), surfaced in one place.
- **Provenance and signing.** A manifest-bound tool can carry version + source URL + (eventually) signature. Bespoke patches in core have none of that — they're indistinguishable from any other line of upstream code in a `git blame`.
- **Scoped permissions.** Today there's no way to say "this tool is callable only from channel X" or "only when session is in mode Y" without writing the gate inline at the call site. A registry lets the bridge enforce per-session / per-channel / per-mode policy uniformly, rather than relying on every patch author to remember the gate.
- **Supply-chain isolation.** A domain-specific tool merged upstream becomes a transitive risk surface for every Plus user. As an external plugin, the same code is opt-in and its blast radius is bounded to the consenting deployment.
- **Granular kill switch.** Disabling a misbehaving plugin tool is `enabled: false` and a process restart. Disabling a tool that was patched into core is a revert + rebuild + redeploy — orders of magnitude slower in an incident.
- **Single audit trail.** Plus already centralises lifecycle events in `orchestrator/telemetry.ts`. Routing tool calls through the same surface gives one log stream covering both "what did Claude observe" and "what did Claude do" — required shape for any SOC2 / regulated-finance posture.

Even outside of compliance pressure, these are the controls anyone running Plus against real money or real customer data would want before shipping more than a couple of tools.

### A — Plugin-spawned MCP server
Plugin manifest declares an embedded MCP entrypoint. `PluginManager` spawns it as a stdio child at `loadPlugin()`, registers it with Claude Code's MCP layer.

Pros:
- Reuses existing MCP machinery; Claude Code sees regular MCP tools
- Process isolation — a tool crash doesn't kill the daemon

Cons:
- One node process per plugin tool surface
- IPC overhead; tools can't share in-process state with their plugin's event handlers

### B — Single multiplexing MCP proxy (built into Plus)
ClaudeClaw-Plus runs **one** MCP server that aggregates tools from all loaded plugins via an in-process `api.registerTool({ name, schema, handler })` — symmetric with the existing `registerCommand`.

Pros:
- One process; tools share state with their plugin's event handlers naturally
- New API surface is small and matches the existing shape (`registerTool` next to `registerService` / `registerCommand`)

Cons:
- Slow tools block others unless handlers are explicitly async-safe
- Plus has to track the MCP protocol surface (resource changes, error envelopes, cancellation)

### C — Native binding through claude-code SDK
Pass plugin tool definitions into Claude Code's context at `session_start` via SDK.

Pros:
- Lowest latency

Cons:
- Couples Plus to claude-code SDK internals — fragile across CC versions
- Likely not supported by current CC SDK

## Open questions

1. Is **option B** (multiplexed MCP proxy) acceptable scope for Plus, or should it live in a separate `claudeclaw-mcp-bridge` package?
2. Should the upcoming plugin manifest (separate proposal — skills/hooks shipped with the plugin) also declare tools, or only allow runtime `api.registerTool`?
3. Permission model — should plugin tools be opt-in per session/channel, gated behind plugin-level config, or always-on once the plugin is enabled?
4. Name collisions across plugins — auto-prefix as `<plugin-id>_<tool>`, or require unique top-level names?
5. Cancellation / timeouts — should the bridge inherit Plus's existing per-handler timeout, or expose a separate budget?

## What I'm proposing

Open the design conversation. Happy to prototype option B if there's appetite — it's the smallest delta that gets us symmetric `registerCommand` / `registerTool` ergonomics, and it's the path I'd personally bet on for downstream extensions.

This pairs naturally with a separate "plugin manifest" proposal (skills + hooks + tools shipped together as one artefact); I can open that issue separately so the threads stay independent.

Nibbler1250:
## Summary

Implements the Plugin → MCP Bridge for ClaudeClaw-Plus, resolving #31.

This adds the missing substrate that unifies all Plus plugins (skills-tuner, archiviste, voice, etc.) under a single MCP-compatible tool registration surface with HMAC-signed inter-process calls.

## Architecture

```
Plus daemon
├── PluginManager (existing src/plugins.ts)
│   └── lifecycle events (gateway_start, session_start, agent_end, ...)
└── PluginMcpBridge (NEW src/plugins/mcp-bridge.ts)
    ├── registerPluginTool(name, schema, handler) ← public API
    ├── McpServer (stdio) ← exposes tools to MCP world
    ├── HMAC signed calls ← secure plugin↔Plus connection
    └── Audit log (~/.config/plus/plugin-audit.jsonl)
```

## Security

- **Per-plugin secret**: `~/.config/plus/plugins/<name>/.secret` (0600, 32 random bytes, auto-created)
- **HMAC-SHA256**: every inter-plugin call is signed with the plugin's secret; verification is timing-safe
- **Zod validation**: all tool args validated against plugin-provided schema before invocation
- **Append-only audit log**: every register/invoke/success/error event logged to `~/.config/plus/plugin-audit.jsonl`

## First Consumer: skills-tuner

The `PlusEventAdapter` stub in the skills-tuner TS port becomes ~20 LOC once this lands:

```typescript
import { getMcpBridge } from 'claudeclaw-plus/plugins/mcp-bridge';

const bridge = getMcpBridge();
bridge.registerPluginTool('skills-tuner', {
  name: 'pending',
  description: 'List pending tuner proposals',
  schema: z.object({ subject: z.string().optional() }),
  handler: async (args) => engine.listPending(args.subject),
});
// apply, refuse, stats, feedback...
```

## Future Plugins (same pattern)

| Plugin | Tools |
|--------|-------|
| archiviste | `search_docs`, `index_doc`, `get_doc` |
| voice (Greg) | `play_tts`, `transcribe_audio` |
| trader-ml-hp | ExternalProcessSubject (Python Optuna) + tool registration |

## Try It

```bash
# Start the MCP server standalone
bun run src/plugins/mcp-server.ts

# Add to claude_desktop_config.json:
{
  "mcpServers": {
    "claudeclaw-plus": {
      "command": "bun",
      "args": ["run", "/path/to/ClaudeClaw-Plus/src/plugins/mcp-server.ts"]
    }
  }
}
```

## Testing

```bash
bun test src/__tests__/plugins/mcp-bridge.test.ts
# 19 pass, 0 fail
```

All 19 tests pass. See `docs/plugin-integration.md` for the full integration guide.

Closes #31

## Security hardening (Tier 1+2)

This PR includes **11 adversarial tests** beyond the functional baseline, covering:

**Tier 1 — Edge cases**:
- `applyProposal` race condition (two concurrent calls on same proposal)
- Path traversal via symlinks (target inside scan_dir but link to /etc/passwd)
- JSONL mid-write corruption (kill -9 during append)
- Audit log atomicity (audit_attempted before any side effect)
- Skill cache invalidated after apply/rename

**Tier 2 — Adversarial security**:
- HMAC canonical bypass via JSON key reordering
- Prompt injection via skill content (not just observation verbatim)
- Telegram callback replay attack
- ExternalProcessSubject with encoded target_path (URL-encoded path traversal)
- Audit log tampering (documented limitation — no hash chain at v1)
- Regex DoS on triggers (catastrophic backtracking — N/A, uses `string.includes()`)

**Real bugs caught and fixed via these tests**:
1. `applyProposal` race condition — added `_applying: Set<number>` lock
2. Symlink traversal in `commitPatch` — added `realpathSync()` check
3. Prompt injection via `skillContent` — `sanitizeObservationContent()` now applied to skill bodies, not just verbatims
4. Empty commits in `commitPatch` — refused when nothing staged (otherwise `revert` was silent no-op)

**Total tests**: 163 unit + integration tests passing. 0 fail.

Full test suite in companion repo: https://github.com/Nibbler1250/skills-tuner-ts

## HTTP Gateway for daemon-style plugins

This PR extends the in-process MCP bridge with HTTP endpoints that let daemon-style plugins (Greg/voice, Archiviste, future cross-process tools) register and serve tools without sharing the Plus daemon address space.

### Endpoints (5)

| Method | Path | Auth | Purpose |
|--------|------|------|---------|
| `POST` | `/api/plugin/register` | Bearer bootstrap | Register manifest + tools |
| `POST` | `/api/plugin/:name/tools/:tool/invoke` | HMAC + 15-min replay | Invoke a tool via callback |
| `DELETE` | `/api/plugin/:name` | Bootstrap or plugin token | Unregister + clean bridge tools |
| `GET` | `/api/plugin/list` | None | List registered plugins with health status |
| `GET` | `/api/plugin/:name/health` | None | Proxy to plugin's health_url |

### Security

- **Bootstrap token**: `~/.config/plus/plugin-bootstrap.secret` (0600, 32 random bytes). Retrieved once via `bun run src/plugins/cli.ts print-bootstrap-token`. Used only for register/unregister — not for invoke.
- **Per-plugin HMAC**: 32-byte secret returned at registration. Every invoke is signed `HMAC-SHA256(secret, ts + "\n" + raw_body)`. Timing-safe comparison.
- **15-minute replay window**: timestamp must be within ±15 min. Clock-skew tolerant, prevents replays.
- **Callback host allowlist**: `localhost`, `127.0.0.1`, `::1` by default. Prevents SSRF to internal services.
- **30-second invoke timeout**: `AbortController` fires after 30s. Plugin dead → 502 clean, daemon unaffected.
- **Standardized error format** with `request_id` for cross-process forensic debugging.

### Graceful degradation throughout

- Audit log fail → stderr fallback, invoke continues
- Plugin unreachable → 502 with clear error, daemon stable
- Malformed manifest → 400 with Zod validation details
- Re-register replaces existing (no separate rotate-token endpoint at v0)

### Deferred to follow-up PRs

| Feature | Reason deferred |
|---------|----------------|
| Hash-chained audit log | No concrete compliance requirement yet |
| Strict capability enforcement | Capability list is declared-only (Tier 2) |
| TLS enforcement | Localhost-only via allowlist at v0 |
| Token rotation endpoint | Re-register covers the use case |

### Integration examples

See `docs/plugin-integration.md` for full Python examples for Greg/voice and Archiviste, including callback handler with HMAC verification.

```python
# Greg quick-start
import requests, hmac, hashlib, json

bootstrap_token = open(os.path.expanduser('~/.config/plus/plugin-bootstrap.secret'), 'rb').read().hex()
resp = requests.post(
    "http://localhost:3000/api/plugin/register",
    headers={"Authorization": f"Bearer {bootstrap_token}"},
    json={
        "name": "greg-voice",
        "version": "1.0.0",
        "schema_version": 1,
        "callback_url": "http://localhost:8765/plus-callback",
        "health_url": "http://localhost:8765/health",
        "tools": [
            {"name": "send_tts", "description": "Play TTS in active call",
             "schema": {"type": "object", "properties": {"text": {"type": "string"}}}},
        ],
        "capabilities": ["tools"],
    },
)
plugin_token_hex = resp.json()["plugin_token"]
```

### New tests (17)

All 17 gateway tests pass. Pre-existing 53 failures unchanged (no regressions).

```bash
bun test src/__tests__/plugins/http-gateway.test.ts
# 17 pass, 0 fail
```

**Full test matrix:**
- register valid → plugin_token returned
- register invalid bootstrap → 401
- register malformed manifest → 400 with Zod errors
- register non-localhost callback → 400 (unless allowedHosts)
- re-register replaces existing + cleans bridge tools
- HMAC tamper → 401 invalid_signature
- stale timestamp >15min → 401 stale_or_future_timestamp
- callback headers verified (X-Plus-Ts, X-Plus-Signature, X-Plus-Request-Id)
- invoke timeout → 502 invoke_failed
- list metadata returned
- health check healthy
- health check unhealthy
- unregister by bootstrap token
- unregister by plugin token
- graceful degradation (audit fail doesn't crash)
- non-plugin path returns null (no interference)
- HMAC tamper detection in invoke path

---

*Companion skill (`mcp.md`) + skills-tuner stack PRs (#39, #40, #41) will follow once this lands.*

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.13",
+      "version": "2.0.14",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.18",
+      "version": "2.0.19",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.14",
+      "version": "2.0.18",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.0.13",
+  "version": "2.0.14",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.0.14",
+  "version": "2.0.18",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.0.18",
+  "version": "2.0.19",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/bun.lock
+++ b/bun.lock
@@ -5,6 +5,7 @@
     "": {
       "name": "claude-heartbeat",
       "dependencies": {
+        "@modelcontextprotocol/sdk": "^1.29.0",
         "ogg-opus-decoder": "^1.7.3",
       },
       "devDependencies": {
@@ -15,6 +16,10 @@
   "packages": {
     "@eshaz/web-worker": ["@eshaz/web-worker@1.2.2", "", {}, "sha512-WxXiHFmD9u/owrzempiDlBB1ZYqiLnm9s6aPc8AlFQalq2tKmqdmMr9GXOupDgzXtqnBipj8Un0gkIm7Sjf8mw=="],
 
+    "@hono/node-server": ["@hono/node-server@1.19.14", "", { "peerDependencies": { "hono": "^4" } }, "sha512-GwtvgtXxnWsucXvbQXkRgqksiH2Qed37H9xHZocE5sA3N8O8O8/8FA3uclQXxXVzc9XBZuEOMK7+r02FmSpHtw=="],
+
+    "@modelcontextprotocol/sdk": ["@modelcontextprotocol/sdk@1.29.0", "", { "dependencies": { "@hono/node-server": "^1.19.9", "ajv": "^8.17.1", "ajv-formats": "^3.0.1", "content-type": "^1.0.5", "cors": "^2.8.5", "cross-spawn": "^7.0.5", "eventsource": "^3.0.2", "eventsource-parser": "^3.0.0", "express": "^5.2.1", "express-rate-limit": "^8.2.1", "hono": "^4.11.4", "jose": "^6.1.3", "json-schema-typed": "^8.0.2", "pkce-challenge": "^5.0.0", "raw-body": "^3.0.0", "zod": "^3.25 || ^4.0", "zod-to-json-schema": "^3.25.1" }, "peerDependencies": { "@cfworker/json-schema": "^4.1.1" }, "optionalPeers": ["@cfworker/json-schema"] }, "sha512-zo37mZA9hJWpULgkRpowewez1y6ML5GsXJPY8FI0tBBCd77HEvza4jDqRKOXgHNn867PVGCyTdzqpz0izu5ZjQ=="],
+
     "@types/bun": ["@types/bun@1.3.9", "", { "dependencies": { "bun-types": "1.3.9" } }, "sha512-KQ571yULOdWJiMH+RIWIOZ7B2RXQGpL1YQrBtLIV3FqDcCu6FsbFUBwhdKUlCKUpS3PJDsHlJ1QKlpxoVR+xtw=="],
 
     "@types/node": ["@types/node@25.2.3", "", { "dependencies": { "undici-types": "~7.16.0" } }, "sha512-m0jEgYlYz+mDJZ2+F4v8D1AyQb+QzsNqRuI7xg1VQX/KlKS0qT9r1Mo16yo5F/MtifXFgaofIFsdFMox2SxIbQ=="],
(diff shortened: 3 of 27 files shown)


TerrysPOV:
Two concerns before this merges:

**1. `reasonedInvokeFn` prompt injection + cost model** — `start.ts` builds the invocation prompt directly from the plugin-supplied tool name and args with no sanitization:
```ts
const prompt = `Use tool \`${fqn}\` with these arguments: ${JSON.stringify(args)}.`;
const result = await runUserMessage("inject", prompt);
```
A malicious or misconfigured plugin could craft `fqn` or `args` to inject arbitrary instructions into a Claude session running with `--dangerously-skip-permissions`. Also, this means every MCP tool call spawns a full Claude invocation — the cost model isn't documented anywhere.

**2. HTTP gateway starts unconditionally** — `getMcpProxyPlugin().start()` runs on every daemon startup regardless of whether `mcp-proxy.json` exists or any external plugins are registered. Daemons without any MCP plugins now open an HTTP server they'll never use.

Suggested approach: only start mcp-proxy if `mcp-proxy.json` exists, and have `reasonedInvokeFn` validate/sanitize `fqn` against a registered tool allowlist before constructing the prompt.

Nibbler1250:
Heads-up @TerrysPOV — **recommended merge order**: this PR (#42, MCP substrate) **first**, then the tuner stack #39 → #40 → #41.

The companion `mcp` skill in PR #41 references the runtime contract this PR provides (HMAC headers, signing scope, replay window, ~/.config/plus/plugin-bootstrap.secret). PR #41 still works standalone — substrate-related skill modes (`proxy-status`, `proxy-restart`, etc.) are conditional on plugin presence — but reviewing #42 first will make the skill content read in proper context.

Independent dependencies — PR #42 is self-contained, no need to wait for the tuner stack.

Nibbler1250:
@TerrysPOV — both concerns addressed:

**Concern #1 (prompt injection + cost model)** — fixed in `b1fe099` (`fix(mcp-proxy): allowlist + args cap on reasonedInvokeFn`):
- `reasonedInvokeFn` now validates fqn shape via regex `/^[a-zA-Z0-9_-]+__[a-zA-Z0-9_-]+$/` (2-segment `<server>__<tool>`, allows dashes for real tool names) before interpolation
- New `McpProxyPlugin.hasReasonedTool(fqn)` allowlist check verifies the server is in the live pool AND the tool was advertised by it
- args size cap at 8000 bytes — rejects oversized payloads before incurring the Claude inject cost; cost model documented inline in the comment
- 2 new tests in `mcp_proxy_mode.test.ts` cover the rejection paths (path traversal, newline injection, backtick injection, oversized fqn segments, unregistered server)

**Concern #2 (HTTP gateway unconditional start)** — already addressed by your commit `b1dc6a6` which is now in this branch.

T1+T2 anti-regression: 64/64 pass. Branch HEAD now `b1fe099`. Ready for re-review when you have a moment.

PR #46, opened 2026-05-10
fix: stop crash + suppress memory-search startup noise

mrancier:
Two small bug fixes found during a fresh install on Linux (Fedora).

**Fix 1 — `stop` command crashes with "Settings not loaded"**

`preShutdownMemorySave()` in `commands/stop.ts` calls `getSettings()` which requires the settings cache to be populated first. When invoked via `claudeclaw stop`, `loadSettings()` has never been called, so it always throws.

Changed `getSettings()` → `await loadSettings()` which loads and returns in one call.

- `src/commands/stop.ts:6` — import `loadSettings` instead of `getSettings`
- `src/commands/stop.ts:34` — `const settings = await loadSettings()`

**Fix 2 — memory-search "binary not installed" logs on every boot**

`indexSessionsBackground()` in `memory.ts` uses `console.log` when the optional `memory-search` binary isn't installed. This prints a noisy setup hint on every daemon start for users who haven't installed the optional Python tooling.

Changed to `console.debug` — still accessible with debug logging, invisible in normal output.

- `src/memory.ts:268` — `console.log` → `console.debug`

Both changes are one-liners, no behavior change beyond the fix. Tested locally on Linux (Fedora, kernel 6.19).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.22",
+      "version": "2.0.23",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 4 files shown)


TerrysPOV:
Thank you for your contribution. Please keep them coming and don't forget to star the repo. Please tell your friends.

PR #50, opened 2026-05-11
fix(telegram): handle pending action callbacks

Nibbler1250:
Fixes pending action system where Telegram button clicks were not being recorded.

## Problem
- User clicks a pending action button in Telegram
- Callback arrives but isn't processed by ClaudeClaw
- Database never updates — action stays in 'pending' status
- pending-worker.py never executes the action

## Root Cause
The handler for `pending:<id>:<value>` callback_data pattern was missing from handleCallbackQuery(). Only `sec_` and `btn:` patterns were handled.

## Solution
Added missing handler that:
1. Matches callback_data: `pending:43:skip`, `pending:1:apply_alt-0`, etc.
2. Calls Python resolve_pending(id, decision) to update DB
3. Edits message to show decision emoji and remove buttons
4. Sends acknowledgment to user

## Testing
- Tuner proposals now send Telegram buttons via pending.py
- Button clicks will now update the database
- pending-worker will process approved actions

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.13",
+      "version": "2.0.14",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.0.18",
+      "version": "2.0.19",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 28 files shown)


TerrysPOV:
Closing in favor of #56, which carries your single telegram-callback commit cherry-picked onto current main with original authorship preserved. Full credit for the fix is to you — #56's PR body references this one.

## Why this got closed rather than rebased

Quick maintainer audit found this branch had 10 commits, 9 of which were the MCP-bridge / mcp-proxy substrate work that already merged into main via #48 and follow-ups. Specifically:

- `d276f09`, `fa8bc7e`, `c1a3140` (MCP bridge + HTTP gateway + mcp-proxy reference consumer) — already in main as `ccb8162 feat(plugins): MCP bridge — universal plugin substrate with HMAC gateway (closes #42)` from the #48 cherry-pick.
- `f0533ba`, `9767ca7`, `16de657`, `500c9ab` (mcp-proxy hardening / test additions) — content is in main under the #48 merge and `3bb1678 fix(mcp-proxy): byte-accurate result cap`.
- `73578ac fix(mcp-proxy): guard _doRestart against concurrent stop()` — already present in `src/plugins/mcp-proxy/server-process.ts:201`.
- `9a0d3d3 fix(mcp-proxy): set status=failed before calling restartHook` — already present in `server-process.ts:181–185`.

So `7b09d60 fix(telegram): handle pending action callbacks from pending.py` was the only commit doing new work — but it was hidden behind a +3844 / 28-file diff that looked unreviewable and conflicted heavily.

## Going forward

If you want to land additional MCP-proxy fixes that genuinely aren't in main yet, please open them as separate PRs — they're hard to triage when they ride along with an unrelated telegram fix. Happy to help cherry-pick anything specific that you think didn't make it through #48.

Thanks again for the fix.

Nibbler1250:
it will be include in next one   thx.

Issue #105, opened 2026-05-17
PTY parser: DM-session output collapses to TUI glyphs (`*`, `↓`, `1`) where channel session produces clean text


TerrysPOV:
## Symptom

Same prompt fired via Discord channel and Discord DM produced wildly different parser output through the sentinel-echo PTY parser — exit code 0 in both cases, but the DM session's captured \"model output\" was unusable.

**Channel path** (session \`9a627338-…\` — \`thread:1488536365477138602\`):

\`\`\`
take a screenshot of https://example.com/ using the playwright mcp and provide the path where this is saved.
\`\`\`

Captured output:
\`\`\`
Savedto/home/claw/project/example-com-screenshot.png
\`\`\`

**DM path** (session \`030a8aaa-…\` — \`global\`):

Same prompt. Captured output:
\`\`\`
*
(8 newlines)
↓
(8 newlines)
1
\`\`\`

Raw \`od -c\` dump from the discord log:
\`\`\`
*\n\n\n\n\n\n\n\n\342\206\223\n\n\n\n\n\n\n\n1
\`\`\`

(\`\\342\\206\\223\` = U+2193 \`↓\` — a TUI scroll/list-indicator glyph.)

## What we know

- The playwright tool **did execute successfully**: \`example-com-screenshot.png\` was overwritten at \`09:57\` (during the DM run), not just the earlier channel run at \`09:44\`. File size matches a fresh render.
- Multiplexer routing was correct: both PTYs hit the same shared \`@playwright/mcp\` child via \`/mcp/playwright\` with distinct per-PTY bearers and identity headers (\`pty-id=global\` vs \`pty-id=thread:…\`).
- Both sessions resumed cleanly. Exit code 0. Turn count incremented.
- The parser issue is **only in the captured \"## Output\" text** that gets relayed back to the surface (Discord) — the underlying tool call worked end-to-end.

## Hypothesis

The DM path uses \`threadId: \"global\"\` (per \`commands/discord.ts:1271\` — \"DMs use the global session\"). The resumed history of the global session may have differed enough from the channel session to push Claude into rendering a markdown list (\`* item\` followed by \`1.\` numbered list, separated by vertical whitespace) that the sentinel-echo parser captured between sentinels but failed to extract the textual content — what survived was just the leading glyphs and the scroll-indicator chrome between them.

Adjacent risk areas worth checking:
[...]

Issue #85, opened 2026-05-16
test: seed ANTHROPIC_API_KEY via libc setenv in PR #83 regression so the test exercises the libc-unset path


TerrysPOV:
## Codex P2 from PR #83

Codex flagged that the libc-level regression test in [`src/__tests__/runner.test.ts`](https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/main/src/__tests__/runner.test.ts) seeds `ANTHROPIC_API_KEY` via `process.env.X = SECRET` — but the very docstring on `withCleanProcessEnv` documents that Bun's `process.env` writes do NOT update libc \`environ\`. So the seed never lands in environ, the bun-pty child never sees it via the parent-env merge regardless of whether \`libc.unsetenv\` is called, and the assertion passes vacuously.

## Fix

Seed via libc \`setenv\` (Bun FFI) BEFORE the test runs, so the var actually lives in \`environ\`. Then pre-fix the test reproduces the leak (child inherits via merge); post-fix it passes (libc.unsetenv really clears environ).

## Acceptance

- The "strips keys at the libc level so bun-pty children do not inherit them via fork+exec" test reproduces the leak when `withCleanProcessEnv` is removed.
- Test passes with `withCleanProcessEnv` calling `libc.unsetenv`.

Issue #84, opened 2026-05-16
fix(runner): _waitForReadySettle resolves too early — supervisor writes prompt into half-painted TUI


TerrysPOV:
## Symptom

With `pty.enabled: true` and PR #83 merged (env leak sealed), a Discord test message produces the claude TUI startup banner as the "response" — no model reply, just the splash screen with the suggestion area visible. Daemon logs show the PTY supervisor's sentinel-echo path returning the captured bytes with no actual model output.

## Root cause

`PtyProcess._waitForReadySettle()` at [`src/runner/pty-process.ts:710-735`](https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/main/src/runner/pty-process.ts#L710-L735) resolves on the FIRST BYTE arriving from claude:

```ts
_waitForReadySettle(timeoutMs: number): Promise<void> {
  return new Promise<void>((resolve) => {
    // ...
    this._handleData = (data: string) => {
      originalHandler(data);
      if (this._firstDataAt > 0) {
        this._clearTimeout(timer);
        done();   // resolved on first byte!
      }
    };
  });
}
```

Claude 2.1.89's TUI takes ~1–2 seconds to fully paint. First byte = TUI starts painting, not "ready for input." `spawnPty()` returns immediately, the supervisor's first `runTurn()` writes the user's prompt into a half-painted TUI, and claude either eats those bytes or interprets them as navigation keys. The sentinel-echo path then captures only the splash bytes and returns them when the quiet window fires.

## Fix

Two-phase wait: resolve once we've seen the first byte AND then a quiet window (no new data for `quietWindowMs`) passes. When the TUI finishes painting it sits silent at the input prompt — that silence is the readiness signal.

## Reproduction

1. Daemon with `pty.enabled: true` on claude 2.1.89.
2. Send any prompt via a configured surface (Discord/Telegram).
3. Response = the claude TUI splash banner, no model output.

Verified in claudeclaw-general on 2026-05-16T16:42 UTC.

## Acceptance

- A real PTY smoke against `/usr/local/bin/claude` with a "Ping" prompt returns a model reply, not the splash banner.
- Existing pty-process / pty-supervisor / pty-trust-prompt suites stay green.
- New regression test that fakes a slow-painting TUI and asserts `spawnPty` doesn't resolve until the paint goes quiet.

Issue #81, opened 2026-05-16
pty-parser: claude 2.1.89 dropped OSC 9;4 progress sequences — parser captures empty response


TerrysPOV:
## Summary

PTY mode (`pty.enabled: true`) failed on the production smoke test 2026-05-16 with every Discord message returning `(empty response)`. Diagnosis traced through to **two distinct failure modes** in claude CLI version `2.1.89`. The first is a trivial operator-side workaround; the second is a real parser-breakage that requires source work.

## Failure mode 1 — Trust-prompt gate blocks first invocation

Claude 2.1.89 shows a one-time interactive trust prompt the first time it's invoked in a directory:

```
Quick safety check: Is this a project you created or one you trust?
❯ 1. Yes, I trust this folder
  2. No, exit
```

The supervisor's `bun-pty` spawn doesn't answer this prompt → claude blocks waiting for input → no output, no progress markers, supervisor times out / returns empty stdout.

**Where the state lives:** `~/.claude.json` under `projects["<absolute-cwd>"].hasTrustDialogAccepted: boolean`. Defaults to `false`. Once set to `true`, the prompt is skipped on subsequent invocations in that cwd.

**Diagnostic observation:** during testing on Hetzner I patched `~/.claude.json` with:

```bash
jq '.projects["/home/claw/project"] = (.projects["/home/claw/project"] // {}) + {hasTrustDialogAccepted: true, mcpServers: {}, allowedTools: [], mcpContextUris: [], enabledMcpjsonServers: [], disabledMcpjsonServers: [], projectOnboardingSeenCount: 1, exampleFiles: []}' ~/.claude.json > /tmp/claude.json.new && mv /tmp/claude.json.new ~/.claude.json
```

After the patch, claude responds correctly in normal interactive mode. The trust gate is bypassed cleanly.

**Two fix paths for this part:**

- **Operator-side (cheap):** add the patch above to the deployment runbook / setup script. Document as a prerequisite for `pty.enabled: true` in any new project directory.
- **Supervisor-side (cleaner):** have `pty-supervisor.ts` or a related initialiser pre-acknowledge trust for `process.cwd()` by writing the `hasTrustDialogAccepted: true` field into `~/.claude.json` on daemon start if absent. ~10 LOC.

The supervisor-side fix is preferable because it self-heals — operators don't need to remember the patch when they redeploy or change project paths.

[...]

PR #106, opened 2026-05-17
docs(architecture): Bus runtime spec v2 — structural replacement for PTY parse, retains PTY as fallback

TerrysPOV:
## Summary

Introduces [`docs/ClaudeClaw_Plus_Bus_Architecture_Spec.md`](docs/ClaudeClaw_Plus_Bus_Architecture_Spec.md) — the engineering design document for ClaudeClaw+ v3.0 Bus runtime.

**Core idea.** Replace the PTY stdin-write / stdout-parse runner with a Plus-internal Bus that uses two Claude Code primitives only:

| Direction | Primitive | Replaces |
|---|---|---|
| Plus → Claude (write) | Channels MCP \`notifications/claude/channel\` | \`pty.write(prompt + "\\r")\` |
| Claude → Plus (read) | Live tail of \`~/.claude/projects/.../<session>.jsonl\` | Sentinel-echo TUI parser |

Zero dependence on TUI byte output. PTY runner retained behind \`runtime: pty\` as fallback per §10 Sprint 5 — it stays in the tree, gated by config, until at least one full release cycle after the Bus becomes default.

## Why now — empirical motivation

Three PTY-parser regressions in two months, all of the same structural class:

- **#81** — Turn-boundary detection failed on \`claude\` 2.1.89. Fix: sentinel-echo parser rewrite (#82).
- **#84** — \`_waitForReadySettle\` returned before TUI paint completed. Fix: settle-detection retune (#85).
- **#105** — DM-mode reply captured as \`*\\n↓\\n1\` glyphs. The model's actual JSONL output was \`Saved to \\\`/home/claw/project/example-com-screenshot.png\\\`\` — the parser pulled box-chrome from the TUI rendering of a backtick code-fence and lost the text inside.

Under the Bus runtime, #105 is structurally impossible — the JSONL tailer reads \`message.content[].text\` directly, with no rendering layer between the model and Plus.

## What changed in v2 vs v1

v1 was a draft sitting locally; this PR introduces it to the repo with revisions baked in:

- **Sprint 0 de-risking spikes** (§10.0) — permission flow validation, JSONL schema snapshot, IPC primitive choice. All must complete before Sprint 1.
- **tmux moved from required to opt-in** (§5.3) — process supervision + stdio is the default; tmux is operator convenience on Unix only. Removes Windows friction.
- **IPC mechanism hardened** (§5.4, §14) — UDS path conventions, perms, atomic create, named-pipe Windows path, and a localhost-TCP+token fallback when neither is viable. Spike 0.3 validates empirically.
- **Plugin API translation shim** (§12.5 new) — legacy hook signatures preserved for two minor releases via in-process shim; deprecation timeline documented.
- **Sprint estimate revised** (§10, §13) — ~3 weeks full-time / ~8–10 weeks part-time. v1's 5-week part-time figure under-estimated Sprint 3 (Discord/Telegram port) and assumed zero schema-drift work during the build.
- **Hetzner staging plan** (§11.8) — explicit cutover sequence for the production deployment.

## Reviewers wanted

@Nibbler1250 — would value your read on:
1. Whether the Channels MCP + JSONL tail primitives match what you've seen in aerolalit's reference implementation and the Anthropic channels docs.
2. The IPC transport choices (§5.4) — UDS by default, named pipes on Windows, localhost-TCP fallback. Anything obviously footgun-shaped?
3. The plugin translation shim (§12.5) — sufficient to keep existing plugins working through a one-release migration window?
4. The Sprint 0 spike scope — anything missing that would derail Sprint 1?

Anyone else who's run Channels community plugins or has Bun + UDS production experience: comments welcome.

## Scope of this PR

**Doc-only.** No code changes. The implementation work is gated on community review of this spec.

## Acceptance for merge

This PR is the design conversation. Merge when:
- [ ] At least one external reviewer (Nibbler or other contributor) has signed off or filed substantive review comments addressed in the doc.
- [ ] Sprint 0 spike scope is agreed.
- [ ] Hetzner staging plan (§11.8) is agreed.

After merge, follow-up PRs implement Sprint 0 → Sprint 1 → … per §10.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.37",
+      "version": "2.2.38",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.37",
+  "version": "2.2.38",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/docs/ClaudeClaw_Plus_Bus_Architecture_Spec.md
+++ b/docs/ClaudeClaw_Plus_Bus_Architecture_Spec.md
@@ -1,7 +1,7 @@
 # ClaudeClaw+ Bus Architecture — Engineering Spec
 
-**Status:** Draft v2 — for community review
-**Author:** Claude (POVIEW.AI engineering assist) — v1: 2 May 2026 — v2: 17 May 2026
+**Status:** Draft v2.1 — incorporates community review feedback
+**Author:** Claude (POVIEW.AI engineering assist) — v1: 2 May 2026 — v2: 17 May 2026 — v2.1: 17 May 2026 (Nibbler review)
 **Target audience:** Coding agent / engineer implementing the refactor; ClaudeClaw+ maintainers and contributors
 **Supersedes:** PTY stdin/stdout runner (`src/runner.ts` PTY codepath, `src/runner/pty-*.ts` helpers)
 
@@ -13,6 +13,19 @@
 - **Plugin API migration formalised** (§11.7, §12.5 new): translation shim spec, deprecation timeline, plugin-author migration guide as a Sprint 4 deliverable.
 - **Sprint estimate revised** (§10, §13): ~3 weeks full-time / ~8–10 weeks at part-time cadence (was 5 sprints/5 weeks part-time — under-estimated Sprint 3 reality).
 - **Empirical motivation** added (§2): three PTY-parser regressions in two months (issues #81, #84, #105), with #105 explicitly traceable to TUI markdown-code-fence rendering — the structural class the Bus eliminates.
+
+### What changed in v2.1 (Nibbler review on PR #106)
+
+All changes are empirical corrections from Nibbler running validation traces against `claude 2.1.126` and aerolalit's reference plugin.
+
+- **§5.2 JSONL schema corrected**: `tool_result` is a content block inside a `user` message, not a top-level type (`runner.ts:891-909` already does this). Additional emitted types documented: `attachment`, `permission-mode`, `file-history-snapshot`, `ai-title`, `last-prompt`, `queue-operation` — `attachment` must not silently drop (UX regression).
+- **§5.1 + §11.4 permission flow corrected** to two notifications: `permission_request` (Claude → plugin) and `permission` (plugin → Claude), carrying the same `request_id`. Declaring `claude/channel/permission` is a contract that the plugin authenticates the replier.
+- **§5.1 `ask` tool added** to the Bus MCP tool set for parity with aerolalit's reference plugin (matters for §9 Channels GA migration).
+- **§10.0 Spike 0.4 added**: slash commands via stdin — promoted from §14.2 Open Q because an `isatty(0)` check inside `claude` would force `tmux` back to required-on-Unix. Better to learn this before Sprint 1.
+- **§10.0 Spike 0.5 added**: session lifecycle markers (`/compact`, `/clear`, `/quit`). The v2 spec assumed `session.init` / `session.end` / `session.compact` JSONL events exist; Nibbler couldn't find them in traces or binary strings. If they don't exist, the tailer needs fallback signals (process exit for end; line-delta or hook for compact).
+- **§5.4 IPC tightened**: macOS-without-XDG path budget, multi-daemon collision via instance-id in socket name, `/proc/<pid>/environ` visibility of `CCAW_BUS_TOKEN`, rationale for the same-uid constraint.
+- **§12.5 shim coverage**: all 9 events in `src/plugins.ts` mapped to Bus correspondence (was 4 of 9 in v2).
+- **§4 / §5.4 reuse note**: Bus core extends Gateway (`src/gateway/`) + EventLog (`src/event-processor.ts`, `src/event-log.ts`) — `BusEvent` is the Bus-runtime shape of `NormalizedEvent` / `EventRecord`, not a parallel primitive.
 
 ---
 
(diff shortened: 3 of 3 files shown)


Nibbler1250 (commented):
Read through the spec carefully, ran checks against `claude 2.1.126` JSONL traces and binary strings, walked through aerolalit's reference plugin, and re-read #105 (the JSONL-has-clean-output evidence is a nice empirical confirm of the read-path bet). Structural call (Bus over PTY) feels right and the Sprint 0 framing is the right shape — flags below are mostly stuff worth catching in those spikes before Sprint 2 commits to a parser shape. Tagged by question.

**Q1 — §5.2 `tool_result` shape**
In actual JSONL, `tool_result` isn't a top-level type — it appears as a content block inside a `user` message (`{type: 'user', message: {role: 'user', content: [{type: 'tool_result', tool_use_id, content, is_error}]}}`). FYI `runner.ts:891-909` already extracts it that way in main, so the tailer parser has a working reference to copy from.

**Q1 — §5.2 line types beyond the 4 listed**
A 231-line trace on current `claude 2.1.126` also emits: `attachment`, `permission-mode`, `file-history-snapshot`, `ai-title`, `last-prompt`, `queue-operation`. Worth deciding per-type in Spike 0.2 — `attachment` in particular carries user-uploaded files, silently dropping would regress UX.

**Q1 / Q4 — §5.2 / §5.3 / §11.1 session lifecycle markers**
Couldn't find `session.init`, `session.end`, or `session.compact` either in the JSONL trace or the binary's subtype strings (what I did find: `elicitation_complete`, `error`, `plugin_install`, `status`, `success`, `task_notification`, `terminated`). They may exist in contexts I haven't hit, but the spec leans on them — `stop()` waits for `session.end` (§5.3), tailer emits `session.compact` (§6.3 step 5), probe asserts both (§11.1 steps 7–8). Might be worth a Spike 0.5 to run `/compact`, `/clear`, `/quit` against a real session and capture what JSONL actually emits. If they're not there, the spec needs fallback signals (process exit for end, line-delta or hook for compact).

**Q4 — §5.3 slash commands via stdin**
Already noted as open in §14.2 #3 — just lobbying to promote it to its own Sprint 0 spike. If `claude` does an `isatty(0)` check, the `process` supervision mode loses slash-command relay and tmux quietly slides back to required-on-Unix. Better to know that before Sprint 1 plans around it.

**Q1 / Q4 — §5.1 / §11.4 permission flow has two directions**
The spec mentions `notifications/claude/channel/permission`. Reading aerolalit's plugin (`plugins/telegram/server.ts:599, 1171, 1402`) and confirming in the binary strings, the flow is actually two notifications:
- `notifications/claude/channel/permission_request` — Claude → plugin, with `request_id`, `tool_name`, `description`, `input_preview`
- `notifications/claude/channel/permission` — plugin → Claude, response carrying same `request_id`

Spike 0.1 probably wants both directions, and §5.1's capability discussion could mention both. Also: declaring `claude/channel/permission` is an opt-in that asserts the plugin authenticates the replier (per aerolalit's comment at lines 510–515). Plus's per-adapter allow-list (§7) covers it, but worth calling out as the contract.

**Q2 — §5.4 IPC transport footguns**
A few things worth tightening:

- **macOS without `XDG_RUNTIME_DIR`** — macOS doesn't set it by default, so the fallback `$HOME/.claudeclaw/run/bus-<agentId>.sock` will be hit on every macOS install rather than as an edge case. The 104-byte `sun_path` cap then becomes the default path budget, not a corner case. Probably fine in practice but worth measuring against realistic `$HOME` lengths.
- **Multi-daemon collision** — `bus-<agentId>.sock` is per-agent, but two daemon instances on the same host with overlapping agent_ids (dev + prod side-by-side, or staging) collide. Suggest adding daemon PID or instance-id to the socket name.
- **Localhost-TCP token via env var** — fallback transport #3 passes `CCAW_BUS_TOKEN` through env. On Linux, env vars are visible in `/proc/<pid>/environ` to anyone on the same uid. The on-disk token at `~/.claudeclaw/agents/<id>/bus-token` is mode 0600 (good), but env-passing means any process on the same uid can read it for the lifetime of the `claude` proc. Minor and same-user-only, but worth a note in §5.4.
- **"Daemon and MCP server run as the same user"** — good constraint, would help to say *why* (rejects multi-user-host attack surface, plus permission bits on UDS).

**Q3 — §11.7 / §12.5 shim coverage**
`src/plugins.ts:25-33` declares 9 daemon events today: `gateway_start`, `session_start`, `before_agent_start`, `before_prompt_build`, `tool_result_persist`, `agent_end`, `session_end`, `message_received`, `after_compaction`. §11.7 maps 4. Partial coverage in the shim = surprise breakage for any plugin depending on the unmapped 5 — would be reassuring to see all 9 listed explicitly with their Bus event correspondence (or marked "no-op under Bus" if intentional).

**Q1 — §4 / §5.4 overlap with existing infra**
Reading the Bus core description (prompt router + response broker + audit log), a lot of it overlaps with what `src/gateway/index.ts`, `src/event-processor.ts`, and `src/event-log.ts` already do (gateway admission, dedupe, sequence numbers, EventRecord persistence). Worth a sentence saying "Bus core extends/builds on Gateway + EventLog" vs treating it as new — otherwise Sprint 1 risks growing a parallel primitive. Same question for `BusEvent` shape vs the existing `NormalizedEvent` / `EventRecord` types.

**Q1 — minor, aerolalit's `ask` tool**
The reference plugin exposes an `ask` tool (returns immediately with `ask_id`; answer arrives later as a channel notification carrying the same id — useful for "agent loop unblocked while waiting for user input"). §5.1 lists `reply`, `request_human`, `cancel`. If parity with the official Telegram plugin matters for the §9 GA migration path, `ask` probably belongs in the set.

---

Happy to run validation traces for whichever spikes you want concrete data on before they get scheduled.

TerrysPOV (commented):
@Nibbler1250 — thank you. This is exactly the empirical pass the spec needed. Running validation traces against the actual binary strings and JSONL of `claude 2.1.126` (not just the version I had on hand) caught structural mistakes that would have wasted Sprint 2's first week. Picking up the schema fixture from `runner.ts:891-909` directly is the right move — we already have a working `tool_result` extractor; the tailer parser should mirror it, not reinvent.

I've pushed **spec v2.1** (commit c8e0edd) incorporating all 9 points:

| # | Your point | Spec update |
|---|---|---|
| 1 | `tool_result` is a content block inside `user`, not top-level | §5.2 table corrected with the inside-`user`-message shape; tailer extraction logic references `runner.ts:891-909` as the working source |
| 2 | Six more JSONL line types (`attachment`, `permission-mode`, …) | §5.2 table extended; per-type Bus topic mapping committed; **`attachment` flagged as must-not-drop** |
| 3 | Session lifecycle markers (`session.init/end/compact`) may not exist | §5.2 fallback inference logic (process exit for end, JSONL line-delta for compact, first non-empty line for init); **Spike 0.5 added** to validate empirically |
| 4 | Slash commands via stdin — `isatty(0)` risk | **Spike 0.4 added** to Sprint 0; outcome cascades into §5.3 default flip |
| 5 | Two-direction permission flow + `claude/channel/permission` capability contract | §5.1 corrected with both notifications + `request_id` correlation; §11.4 updated; capability contract called out as "plugin authenticates the replier" |
| 6 | IPC footguns (XDG on macOS, multi-daemon collision, env var visibility, same-uid rationale) | §5.4 hardened: `instanceId` in socket name, macOS `sun_path` budget validation, `/proc/<pid>/environ` caveat, explicit same-uid rationale |
| 7 | Shim coverage: 4 of 9 events mapped, need all 9 | §11.7 full 9-row table; §12.5 scope updated |
| 8 | Overlap with Gateway + EventLog | §5.4 explicit "Bus core extends Gateway + EventLog, doesn't duplicate"; `BusEvent` declared as the Bus-runtime shape of `NormalizedEvent` / `EventRecord` |
| 9 | `ask` tool from aerolalit's plugin | §5.1 ask tool added with async `{ask_id}` return semantics; noted as required for §9 GA migration parity |

Sprint 0 now has 5 spikes (added 0.4 and 0.5 from your feedback). Budget revised: ~5–7 days full-time / ~2 weeks part-time.

Taking you up on the validation-traces offer for Spikes 0.4 and 0.5 specifically — if you have a working `claude 2.1.126` setup with the dev-channels flag wired up, capturing the JSONL diff under `/compact`, `/clear`, `/quit` would settle the biggest remaining ambiguity in the parser shape. No urgency; happy to run them on my side too.

Kicking off Sprint 0 now in parallel with the remaining PR #106 review window.

PR #112, opened 2026-05-17
docs(spikes): cross-validation traces — claude 2.1.126/proot + 2.1.143/Ubuntu native

Nibbler1250:
## Summary

Follow-up to PR #106 review and PR #109 spike findings. Captures the Spike 0.4 and 0.5 behaviors on two platforms different from the ones in `docs/spikes/`, as a cross-version + cross-arch sanity check.

## Platforms

| Platform | claude version | OS | Arch | Reason |
|---|---|---|---|---|
| S26 (proot) | **2.1.126** | Ubuntu in proot (Android) | ARM64 | Different version than PR #109 to test forward compat |
| Native | **2.1.143** | Ubuntu 24.04 native | x86_64 | Same version as PR #109 to validate transfer across hardware/install |

Two-axis sanity check: version skew **and** arch/OS skew.

## TL;DR

All structural findings from `docs/spikes/0.4-slash-stdin.md` and `docs/spikes/0.5-lifecycle-markers.md` reproduce identically on both platforms.

- Plain stdin gates to `--print` and dies with `Error: Input must be provided either through stdin or as a prompt argument when using --print` (exit 1). Timing: ~3.6 s native, ~5.5 s under proot. Behavior identical.
- `compact_boundary` system subtype appears on successful `/compact` with sufficient history, on both 2.1.126 and 2.1.143.
- `compactMetadata` shape on both versions: `{trigger, preTokens, postTokens, durationMs, preservedSegment: {headUuid, anchorUuid, tailUuid}}`. `preservedSegment.headUuid/anchorUuid/tailUuid` looks useful for the tailer to correlate pre/post compact transcript ranges — surfacing as a flag.
- `/clear` rotation: same on both. Old JSONL untouched; new session UUID file materialises; `/clear` + `/exit` envelopes go to the new file.
- `/quit` envelope: `<command-name>/exit</command-name>` + `<local-command-stdout>{farewell}</local-command-stdout>` on both.
- Farewell variants observed on 2.1.126: `Bye!`, `Goodbye!`, `See ya!`, `Catch you later!` (4 distinct; spike doc mentions 2).

## New observation worth flagging

**`stop_hook_summary` system subtype** — not enumerated in §5.2 or in `docs/spikes/0.2-jsonl-schema-snapshot.md`. Fires once per turn when the operator has Stop hooks configured. Shape:

```json
{
  "type": "system",
  "subtype": "stop_hook_summary",
  "hookCount": 1,
  "hookInfos": [{"command": "...", "durationMs": 40}],
  "hookErrors": [],
  "preventedContinuation": false,
  "stopReason": "",
  "hasOutput": false,
  "level": "suggestion",
  "toolUseID": "...",
  "uuid": "...",
  "version": "2.1.143"
}
```

Two ways to handle in the tailer (either is defensible):
- enumerate it explicitly as `bus.event.hook_summary` (carries useful per-turn timing telemetry)
- fall through to `bus.event.unknown` for now; revisit when Hooks-on-Bus is on the table

Also observed: `attachment.type: "plan_mode"` content block when running under `--permission-mode plan`. Same fall-through question.

## Layout

```
docs/spikes/cross-validation/
├── README.md
├── probes/
│   ├── slash-stdin-probe.py       # Spike 0.4, Python pty.fork() port of the bun-pty probe
│   └── lifecycle-deep-probe.py    # Spike 0.5 with multi-turn variants
└── fixtures/
    ├── slash-stdin/               # 3 plain-stdin runs + PTY + --print sanity
    ├── lifecycle/                 # claude 2.1.126 proot, 1-turn history
    ├── lifecycle-multi-turn/      # claude 2.1.126 proot, 4-turn history (compact_boundary fires)
    └── lifecycle-multi-turn-prodesk/  # claude 2.1.143 native, 4-turn history
```

## Probes

Python ports of the `bun-pty` TypeScript probes — same kernel `/dev/ptmx` primitive underneath, runs on any POSIX with Python 3.10+. Useful for cross-validating from a host that doesn't have Bun installed.

## Limitations disclosed

- Bare-slash without a prior prompt was sometimes queued as input text instead of being recognized as a slash command. All multi-turn fixtures got around this by sending a real prompt first. Looks like a probe-timing issue, not a `claude` behavior issue — bun-pty in the Sprint 0 probes presumably gets around it via different REPL-ready detection.
- Test B's farewell enumeration captured 0 farewells in the multi-turn dir because of the same issue. The 4 distinct farewell variants were captured in individual session JSONLs across the broader run set.
- Captures used `--permission-mode plan` to avoid the bypass-permissions confirmation dialog; an `Enter` blast was sent at startup to dismiss the trust dialog.

## Suggested follow-ups (low priority)

1. Add `stop_hook_summary` to §5.2 line types (or document as fall-through)
2. Document `compactMetadata.preservedSegment.{head,anchor,tail}Uuid` in §5.2 (useful for tailer correlation)
3. Add `attachment.type: plan_mode` content block to the per-type table

Happy to extend any of these into a follow-up patch if useful.


TerrysPOV:
Thanks @Nibbler1250 — really valuable. Two-axis sanity check (version skew + arch/OS skew) is exactly the kind of independent confirmation that makes the Bus runtime decisions feel earned rather than guessed. Glad you bothered to do the proot run too — Android-as-host is a real deployment shape we hadn't covered.

The three new findings (`stop_hook_summary`, `compactMetadata.preservedSegment.{head,anchor,tail}Uuid`, `attachment.type: plan_mode`) are exactly the right level of detail to fold into the schema work. Going to merge this now and pick those up in Sprint 4 alongside the spec amendment to §5.2 and the corresponding `src/bus/jsonl-line-types.ts` enumeration — the current `bus.event.unknown` fall-through handles them safely until then, so no rush.

Python ports of the probes are a nice bonus — they unblock anyone who wants to validate from a host without Bun. Going to leave them under `docs/spikes/cross-validation/` as the canonical reference.

Merging.

Issue #139, opened 2026-05-20
telegram adapter: route by origin_id like discord/slack for multi-chat-per-agent


TerrysPOV:
## Context

5-agent code review of #138 surfaced this as Finding A2 (confidence 75). Filed as a follow-up per the option-2 decision on that PR — A1 (lastPromptOrigin lifecycle) was addressed inline in #138; A2 is small enough to land separately and benefit from its own review cycle.

## Problem

PR #138 fixed the foreign-origin fan-out bug by adding a `eventBelongsTo*` guard to every adapter. For Discord and Slack, the fix went further: when `origin === "<this-adapter>"` is set, route ONLY to `origin_id`'s channel. For Telegram it stopped at the drop guard — `handleResponseText`, `handlePermissionRequest`, and `handleRequestHuman` still route via `targetForAgent(agentId)`, which returns `lastChatPerAgent.get(agentId)` (a single target tracking the last inbound chat for that agent).

Today this is invisible because production has one Telegram chat per agent. The moment a deployment routes two Telegram chats to the same agent (multi-channel-per-agent like Discord), Telegram will misroute on interleaved prompts: chat A sends prompt → `lastChatPerAgent[default] = A`. Chat B sends prompt → `lastChatPerAgent[default] = B`. Claude's reply to A's prompt arrives → telegram passes the `eventBelongsToTelegram` guard (origin is "telegram"), then `targetForAgent` returns the latest cache entry = B → reply lands in the wrong chat.

## Files

- `src/adapters/telegram/index.ts` — `handleResponseText` (~line 343), `handlePermissionRequest` (~line 378), `handleRequestHuman` (~line 417), `targetForAgent` (~line 511)

## Proposed fix

Read `event.payload.origin_id` first; build a `PendingPrompt` from it (chat_id only, no message_thread_id since origin_id is a flat string). If origin_id matches `lastChatPerAgent[agent].chat_id`, prefer the cached `PendingPrompt` (it preserves `message_thread_id` for threaded chats). Otherwise build a bare target from origin_id alone.

For events with no origin (cron, heartbeat, scheduler), keep the existing `targetForAgent` fallback.

## Notes

- Discord's fix is at `src/adapters/discord/index.ts:382-398` and Slack's at `src/adapters/slack/index.ts:577-586` (post-#138); use those as the pattern.
- Telegram's `targetForAgent` docstring (~line 511) needs an update to acknowledge the new origin_id path.
[...]

PR #141, opened 2026-05-20
feat(bus): inline progress UX — edit_message tool, typing + braille spinner

Nibbler1250:
Surface-side progress feedback, built on top of the headless prompt-delivery fix (#140).

## What

- **`edit_message` MCP tool** — edits the agent's most-recent outbound message in place (no `message_id`; the adapter tracks the last bot message per agent). New `IpcEditMessage` type; `BusCore` routes it to a `response.edit_text` event the adapter consumes via `editMessageText`. Added to the spawn `--allowedTools`.
- **Auto-spinner** — the instant an inbound message arrives, the Telegram adapter posts a braille placeholder (`⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏` @ 1.1s) and animates it, plus `sendChatAction(typing)`. The first reply edits that placeholder in place; progress replies and `edit_message` calls keep editing the same "live turn" message; `intent:'final'` stops the spinner and finalizes. Guarantees visible activity regardless of claude's pattern — single evolving message.
- **mcp-server instructions** teach DIRECT vs PROGRESSIVE: DIRECT (single `reply:'final'`) for pure chat; PROGRESSIVE (`reply:'progress'` → `edit_message`… → `reply:'final'`) whenever tools are used.

## Note
- **Stacked on the headless-prompt-delivery PR.** This branch's first commit is that fix — review/merge it first; the second commit is this feature.
- Only the Telegram adapter renders the spinner / `edit_message`. Discord/Slack adapters receive the `edit_message` event but ignore it for now (easy follow-up).

## Tests
- `edit_message` MCP tool test added; telegram subscription-count test updated 6→8 (new `response.edit_text` topic).
- 60/60 telegram + core + mcp-server pass.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.65",
+      "version": "2.2.82",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.82",
+      "version": "2.2.84",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.84",
+      "version": "2.2.88",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.82",
+  "version": "2.2.84",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.65",
+  "version": "2.2.82",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.84",
+  "version": "2.2.88",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -84,3 +84,67 @@ bun run bump:marketplace-version
 ```
 
 Both `plugin-version-guard` and `marketplace-version-guard` are required CI checks. They fail if `.claude-plugin/plugin.json` or `.claude-plugin/marketplace.json` still carry the same version as the merge base. Run the bumps, commit alongside your code changes, and push before creating the PR.
+
+---
+
+## Software Development Cycle (SDC)
+
+This project follows a **9-stage Claude-Code-managed SDC**. The Claude Code agent loop is the orchestrator; specialised sub-agents do the focused work, parallelised where independent and sequential where dependent. Worktrees isolate parallel work streams.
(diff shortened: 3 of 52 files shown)


TerrysPOV (changes requested):
### Code review

Found 3 issues:

1. Regression: the `eventBelongsToTelegram` foreign-origin guard added in #138 is removed from all three outbound handlers in `src/adapters/telegram/index.ts`. The shipped production fix for the post-#137 cross-adapter mirror bug is undone — webui / discord / slack replies will fan out into Telegram again, and the Codex P1 follow-up on #138 (allowing cron/heartbeat to fan-out through the same path) is also lost. Re-add the guard at the top of `handleResponseText`, `handlePermissionRequest`, and `handleRequestHuman`, plus the `CHANNEL_DRIVEN_ORIGINS` / `BusOrigin` imports.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/355b83fb4e7a5706aa7fd03af8f56976d4075c49/src/adapters/telegram/index.ts#L373-L378

2. Spinner infinite loop after Telegram 400. `startSpinner`'s 1100ms interval calls `editMessageText` with `.catch(() => {})`, which silently swallows `message to edit not found` and `message is not modified`. If the user deletes the bot placeholder message, the interval fires forever generating a continuous 400 stream. Suggest stopping the spinner on the first 400 (or any non-rate-limit error from `editMessageText`).

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/355b83fb4e7a5706aa7fd03af8f56976d4075c49/src/adapters/telegram/index.ts#L519-L555

3. Missing plugin/marketplace version bumps. Required by the project Development Workflow note in `CLAUDE.md`: both `plugin-version-guard` and `marketplace-version-guard` CI checks fail when `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` aren't bumped past the merge base. Run `bun run bump:plugin-version` + `bun run bump:marketplace-version` and commit alongside.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/35944aeed072ae4f409da7ceb51759b1c3922546/CLAUDE.md#L79-L86

TerrysPOV (changes requested):
### Code review

Re-review after follow-up commits. The original P0 + spinner inf-loop are resolved — thanks for restoring the `eventBelongsToTelegram` guard and wiring `isTelegramRateLimit` into the spinner. Three remaining concerns introduced or surfaced by the new PTY-prompt-delivery design.

**1. New P1 — `send_prompt_stream` is concurrent-write racy.**

`session-agent-process.ts` writes the prompt text, sleeps 200ms, then sends `\r`. If two prompts dispatch against the same PTY within that 200ms window (a burst from a busy chat, or two adapters firing close together), the second `write(line)` lands before the first call's `write("\r")` — interleaving text in the PTY input buffer and corrupting both prompts. The bus has no per-agent serializer on the dispatch path; the webui-bridge has a mutex for exactly this reason (see `src/bus/webui-bridge.ts` Codex P1 on PR #136). Suggest a per-agent `Promise` chain in `send_prompt_stream` so writes serialise.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5b9db3a2ea0a/src/bus/session-agent-process.ts#L100-L135

**2. New P2 — Dialog-dismiss `\r` can fire mid-turn under PTY prompt delivery.**

The widened dialog-dismiss spread (1s / 2.5s / 5s / 8s) was harmless when prompts arrived via `notifications/claude/channel`. Now that the same PTY carries actual prompts, a `\r` firing at 5s or 8s lands on whatever's in the REPL state at that moment — including a mid-turn prompt the supervisor just wrote. The comment says "harmless empty REPL submit" but each `\r` does start a claude turn and emits usage telemetry. Suggest cancelling the timers once the first prompt is observed (or replacing the spread with a single up-front write that the supervisor knows it owns).

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5b9db3a2ea0a/src/bus/session-manager.ts#L515-L560

**3. P1 carried over — pending maps keyed by `agentId` only.**

This was flagged on the first review pass and remains. `lastBotMessage`, `spinnerState`, and `turnActive` are all `agentId`-keyed. Same #117 P1 / #138 Discord-fix pattern: when two telegram chats route to the same agent (the exact scenario issue #139 already tracks), the spinner placeholder + `lastBotMessage` entry for chat A get overwritten when chat B arrives, and chat A's reply edits chat B's placeholder. Plus `lastBotMessage` is never evicted per-turn, so a delayed cron/heartbeat reply will edit a stale message_id from hours ago. Composite-key the maps as `${agentId}:${chatId}` and `delete()` after the final reply.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5b9db3a2ea0a/src/adapters/telegram/index.ts#L83-L105

Requesting changes — happy to land once these three are addressed (or items 1+2 if you'd prefer to fold the agentId-keying work into the issue #139 PR).

TerrysPOV (approved):
### Code review

Re-review of commit 4330117. All three concerns from the prior pass are cleanly addressed:

1. `send_prompt_stream` write race — fixed via per-process `writeChain` that serialises the full text → 200ms → CR sequence; test exercises two overlapping prompts and asserts non-interleaved byte order.

   https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/96095ef1e46d7ede48f5328d79a2b74189230bf6/src/bus/session-agent-process.ts#L145-L160

2. Dialog-dismiss CR mid-turn — the four boot timers are now owned by `PtyAgentProcess` and cancelled on the first real prompt and on `_kill()`; test verifies the first-prompt cancel path.

   https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/96095ef1e46d7ede48f5328d79a2b74189230bf6/src/bus/session-agent-process.ts#L130-L165

3. Composite-key maps — `lastBotMessage`, `spinnerState`, `turnActive` keyed by `${agentId}:${chatId}` via `convKey()`, with per-turn `delete()` on final reply; test confirms a post-final `edit_message` falls back to a fresh send rather than editing a stale id.

   https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/96095ef1e46d7ede48f5328d79a2b74189230bf6/src/adapters/telegram/index.ts#L83-L110

No new findings. Approving.

PR #152, opened 2026-05-22
fix(telegram): bound spinner edits + per-chat flood cooldown

Nibbler1250:
Follow-up to #141. Found in production on a live deployment.

## The incident

The inline-progress spinner from #141 edits the live Telegram message every
1.1s for the **entire duration** of a claude turn. A multi-minute turn issues
100+ `editMessageText` calls to a single chat. Telegram answered with an
escalated flood-ban — captured live:

```
FAIL sendMessage 429 {"ok":false,"error_code":429,
  "description":"Too Many Requests: retry after 1730",
  "parameters":{"retry_after":1730}}
```

Once banned, **every** send to that chat 429'd (placeholder, edits, the final
reply), so the bot silently stopped responding. The 1.1s edit loop kept firing
into the ban, and each attempt risked extending it — the chat never recovered
on its own.

## The fix

**1. Bound the spinner (`src/adapters/telegram/index.ts`)**
- Interval `1100ms` → `2500ms`.
- Hard cap `SPINNER_MAX_TICKS = 12` (~30s of animation), after which the message
  goes static. Real text still flows via `response.text` / `edit_message`
  events, which are bounded by the agent's output rather than a wall-clock timer.
- The spinner now stops on **any** API error (429 included) instead of looping
  through rate-limits.

A long turn now issues ≤12 edits instead of 100+.

**2. Per-chat flood cooldown (`src/adapters/telegram/api.ts`)**
- On a `429` carrying `parameters.retry_after`, record the deadline per chat.
- Subsequent sends to that chat are short-circuited **locally** (no API call)
  until it clears, so a misbehaving caller can't sustain the ban.
- `retry_after` is surfaced in the thrown error for observability.

## Validation

Reproduced and verified live on a real deployment: after the fix, a ~21s turn
produced **10** spinner edits (under the cap), ~2.5s apart, with **zero** 429s,
and replies delivered normally. Before the fix the same usage triggered the
~1730s ban.

## Tests

`src/adapters/telegram/__tests__/api.test.ts` (new) covers the cooldown:
same-chat short-circuit (no second fetch), other chats unaffected, `retry_after`
surfaced. `bun test src/adapters/telegram` — green; `bun run lint` clean.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.88",
+      "version": "2.2.89",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.88",
+  "version": "2.2.89",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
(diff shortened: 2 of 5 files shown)


TerrysPOV (approved):
Approving. The fix correctly bounds the spinner (≤12 edits over ~30s) and adds per-chat short-circuit on 429 — the production flood-ban scenario is unreproducible under these constraints. Tests cover the cooldown short-circuit, other-chat isolation, and \`retry_after\` surfacing.

Three non-blocking residual observations (memory leak in \`floodUntil\`, concurrent in-flight race, dead \`isTelegramRateLimit\`) tracked in #154 for follow-up — none block merge.

PR #155, opened 2026-05-22
telegram(api): #154 follow-ups — flood-map eviction, per-chat serialization, dead-code removal

### LINKED ISSUE #154: telegram(api,spinner): three follow-ups from PR #152 review
Non-blocking observations carried over from the #152 review. The PR ships the production hot-fix (bounded spinner + per-chat 429 cooldown); these are residual hardening items.

## 1. `floodUntil` map has no eviction

`createTelegramApi` in `src/adapters/telegram/api.ts` stores `chatKey -> deadlineMs` on every 429 but never deletes entries. Stale entries are harmless for correctness (the `now < deadline` guard fails after expiry), but the map never shrinks.

Fix options:
- `floodUntil.delete(chatKey)` in the `remaining <= 0` else-branch on the next observation of that chat.
- Periodic GC tick.

Low urgency — entries are tiny and rare in practice.

## 2. Concurrent in-flight race on the cooldown write

`floodUntil.set(chatKey, ...)` happens *after* `await fetch(...)` resolves with the 429. Two concurrent sends to the same chat can both fire to Telegram before either's 429 lands in the map.

Post-PR-#152 this is at most one extra in-flight request, not the 100+ edit storm of the original incident — the spinner is now serial through `setInterval`. But the cooldown blocks attempt N+1, not simultaneous in-flights. Worth surfacing if any future caller dispatches parallel sends to the same chat.

Possible fix: optimistically record `floodUntil` at request-start time if a recent 429 exists, or guard with a per-chat in-flight semaphore.

## 3. `isTelegramRateLimit` is now dead code

After #152 the spinner stops on any API error (429 included), so the `isTelegramRateLimit` helper in `src/adapters/telegram/index.ts` has no remaining caller. Either delete it or keep it for future use — operator's call.

## Related

- PR #141 — original spinner introduction
- PR #152 — bounded spinner + per-chat 429 cooldown (the hot-fix)
- Review thread on #152 surfaced these via the 5-agent pass

Nibbler1250:
Closes #154. The three non-blocking hardening items carried over from the #152 review.

## 1. `floodUntil` eviction
Entries are deleted once their cooldown elapses (checked on the next send to that chat), so the map no longer grows unbounded.

## 2. Per-chat send serialization
Sends to a given `chat_id` are chained so at most one is in flight at a time. This closes the concurrent-in-flight race on the cooldown write that #154 flagged — a 429 from one send now lands in `floodUntil` before the next send starts — and is rate-limit-friendly (no same-chat bursts, which is what caused the original #152 incident). `getUpdates` (no `chat_id`) is never chained, so the long-poll is unaffected. The per-chat tail entry is dropped once it settles and nothing newer chained.

## 3. Remove dead `isTelegramRateLimit`
After #152 the spinner stops on any API error (429 included), so the helper had no remaining caller. Deleted it and its tests.

## Tests
`api.test.ts` gains two cases: same-chat sends never overlap (max 1 in-flight), and different chats still send concurrently. `bun test src/adapters/telegram` green; `bun run lint` clean. Version 2.2.89 → 2.2.90.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.89",
+      "version": "2.2.90",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.89",
+  "version": "2.2.90",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/adapters/telegram/__tests__/api.test.ts
+++ b/src/adapters/telegram/__tests__/api.test.ts
@@ -67,3 +67,38 @@ describe("createTelegramApi — per-chat flood cooldown", () => {
     );
   });
 });
+
+describe("createTelegramApi — per-chat send serialization (#154)", () => {
+  function stubConcurrencyTracker(): { maxConcurrent: number } {
+    const state = { inFlight: 0, maxConcurrent: 0 };
+    globalThis.fetch = (async () => {
+      state.inFlight += 1;
+      state.maxConcurrent = Math.max(state.maxConcurrent, state.inFlight);
+      await new Promise((r) => setTimeout(r, 25));
+      state.inFlight -= 1;
+      return new Response(JSON.stringify({ ok: true, result: {} }), { status: 200 });
+    }) as typeof fetch;
+    return state;
+  }
+
+  it("never has two in-flight sends to the same chat at once", async () => {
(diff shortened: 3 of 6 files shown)


TerrysPOV (approved):
Approving. All three issue #154 follow-ups addressed cleanly:

  - \`floodUntil\` evicts expired entries via \`delete(chatKey)\` in the \`remaining <= 0\` branch — lazy eviction, bounded.
  - Per-chat serialisation via \`chatTail\` Promise chain — symmetric success/failure handlers so a thrown send doesn't jam the queue.
  - \`isTelegramRateLimit\` removed and zero callers remain at HEAD.

Two non-blocking observations for a future tidy:
 
1. \`chatTail\` cleanup is identity-guarded and lazily evicts on settle. A chat with a slow-but-perpetual send rate (one msg per minute) keeps one Promise per chat in the map — not a leak in practice but bounded by active-chat count, not total observed chats.
2. \`stubConcurrencyTracker\` in \`api.test.ts\` mutates \`globalThis.fetch\` without restore. Could pollute subsequent test files.

TerrysPOV (approved):
Approving. All three issue #154 follow-ups addressed cleanly: floodUntil evicts expired entries, chatTail per-chat serialisation with symmetric
  error handlers, isTelegramRateLimit cleanly removed. Two non-blocking observations: (1) chatTail map keeps one Promise per active chat in slow-perpetual streams — bounded but worth knowing; (2)
  stubConcurrencyTracker in api.test.ts mutates globalThis.fetch without restore.

PR #157, opened 2026-05-24
fix(bus): Windows support for schema-probe (cwd encoding + claude spawn)

Nibbler1250:
## Summary
Two **Windows runtime bugs** in the bus schema-probe / JSONL tailer, found via a real Windows 11 host validation:

1. **cwd encoding (core read path).** `encodeCwdForProjectsDir` replaced only `/`. On Windows the cwd (`C:\Users\...`) was left intact, so `predictJsonlPath` produced a malformed path that never matched the directory claude.exe actually writes (`C--Users-...`) — the tailer could not locate the session on Windows. Now replaces `\`, `/` and the drive `:` on win32 (verified claude.exe maps `C:\Users\foo\bar` -> `C--Users-foo-bar`); POSIX still keeps `:` (a legal path char). `platform` is injectable so both branches are covered on Linux CI.

2. **claude spawn.** `captureClaudeVersion` spawned `claude` directly; on Windows that's `claude.cmd`, and Node refuses to spawn a `.cmd` without a shell since CVE-2024-27980 (EFTYPE) — crashing the probe at startup. Routed through the shell on win32 only (static `--version` args, no injection).

Also makes the `predictJsonlPath` test OS-aware (it hardcoded a POSIX separator).

## Validation
- **Linux** (ARM64 proot + x86_64 Ubuntu): probe + tailer suites **40 pass / 0 fail**.
- **Windows 11** (build 26200, claude 2.1.150): reproduced both failures; after the fixes the probe resolves and reads the session JSONL (the cwd-encoding ENOENT crash is gone) and `captureClaudeVersion` succeeds.

## Known follow-up (out of scope)
Full green of the `SchemaProbe.run()` *unit* suite on Windows needs test-harness portability (the `.sh` claude stub + the non-zero-exit test's bad-bin stub aren't runnable on Windows). Test-only — CI (Linux) is green.

## Test plan
- [ ] CI green (biome / build / plugin + marketplace version-guards)
- [ ] Reviewer sanity-check the two `win32` branches

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.89",
+      "version": "2.2.90",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.90",
+      "version": "2.2.91",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.89",
+  "version": "2.2.90",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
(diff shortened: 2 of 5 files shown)


TerrysPOV (approved):
Approving. The win32 cwd encoding fix is correct (replaces \`/\`, \`\\\`, and \`:\` only on win32, leaves POSIX behaviour untouched), the \`predictJsonlPath\` test is properly cross-platform via \`path.join\`, and the \`shell: true\` spawn on win32 is correctly scoped to a hardcoded \`--version\` invocation with \`claudeBin\` as operator config — no injection surface, CVE-2024-27980 acknowledged in the comment.

Pushed a small maintainer follow-up (4854711) to land the same injectable-platform pattern on \`captureClaudeVersion\` that you used on \`encodeCwdForProjectsDir\` — its \`shell: process.platform === "win32"\`branch was untestable off-Windows. Adds an injectable \`platform\` + \`spawn\` and three regression tests (shell:true on win32, shell:false on linux, default matches host). No production behaviour change. Feel free to tweak / squash / drop.

Win11 host validation appreciated. SchemaProbe full-suite Windows follow-up acknowledged.

TerrysPOV (approved):
Approving. Win32 cwd encoding fix is correct, POSIX behaviour preserved, shell:true spawn properly scoped to hardcoded --version with claudeBin
  as operator config. Pushed maintainer follow-up 4854711 to apply the injectable-platform pattern to captureClaudeVersion as well — was inconsistent with encodeCwdForProjectsDir in the same PR. Three new tests
  verify shell flag per platform. Win11 host validation appreciated.

PR #160, opened 2026-05-25
fix(telegram): client-side timeout on getUpdates — recover from hung long-poll (bot goes silently deaf)

Nibbler1250:
## Problem

`getUpdates(offset, timeoutSeconds, signal)` in `src/adapters/telegram/api.ts` passed only the Telegram **server-side** long-poll `timeout` param plus the `stop()` abort signal. There was **no client-side request deadline**.

When a long-poll connection goes half-open — gateway 502/504 flakiness or a network blip — the underlying `fetch` neither resolves nor rejects. `await getUpdates(...)` then hangs forever, and the poll loop in `index.ts` (`pollLoop`, the `while (running && generation === gen)` / `await this.api.getUpdates(...)`) **freezes silently on that await**:

- no error thrown (a hang doesn't throw), so the loop's `catch` never runs;
- no `"getUpdates failed"` / `"poll loop crashed"` log line;
- process otherwise healthy — web dashboard and cron keep running.

The bot just **stops receiving messages until a manual restart**. This bit us in production (Telegram adapter went deaf with zero log signal).

## Fix

Layer a **client-side hard timeout** onto the getUpdates request:

```ts
const deadline = AbortSignal.timeout(timeoutSeconds * 1000 + GETUPDATES_TIMEOUT_BUFFER_MS);
const combined = signal ? AbortSignal.any([signal, deadline]) : deadline;
```

- **Buffer (5000ms over the long-poll window)** keeps normal ~30s long-polls untouched — only a genuinely stuck request aborts.
- **`AbortSignal.timeout` rejects with a `TimeoutError`**, which is *distinct* from the `AbortError` that `stop()`'s manual abort produces. `AbortSignal.any` forwards whichever source fires first, so the distinction is preserved end-to-end.
- The poll loop's **existing** catch already does the right thing with each case:
  - `stop()` → `AbortError` → loop breaks (clean exit, unchanged);
  - hung request → `TimeoutError` → falls through to `logger.error("getUpdates failed", err)` → sleep → retry (**self-heals**).

Scoped to `getUpdates` (the only call with a long timeout). Other methods are unchanged.

## Tests

`src/adapters/telegram/__tests__/api.test.ts`
- a getUpdates whose `fetch` never settles **aborts within the client timeout** (rejects with `TimeoutError`) instead of hanging forever;
- `stop()`'s abort still propagates as `AbortError` (clean-stop path preserved).

`src/adapters/telegram/__tests__/telegram.test.ts`
- the poll loop **logs the failure and keeps polling** after a `TimeoutError` rejection, then dispatches the next inbound update (proves self-heal — before the fix the loop simply froze).

Telegram adapter suite green (34/34). Full-suite delta: **+3 passing tests, no new failures** (the pre-existing failures on `main` are environmental — mcp-multiplexer/escalation/session-persistence — and unrelated to this change).

## Versioning

`bump:plugin-version` + `bump:marketplace-version` run → `2.2.93` (`plugin-version-guard` / `marketplace-version-guard` satisfied). Rebased on top of current `main`.

## Why now — the Bus cutover (#159) just landed

#159 flipped the default runtime to the Bus, so the Telegram adapter is now on the **default** path for most installs. That makes this fix more urgent, not less: a poll loop that can freeze silently with zero log signal is exactly the failure mode you don't want load-bearing right after the cutover. Small, self-contained, rebased on top of #159.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.92",
+      "version": "2.2.93",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.92",
+  "version": "2.2.93",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/adapters/telegram/__tests__/api.test.ts
+++ b/src/adapters/telegram/__tests__/api.test.ts
@@ -102,3 +102,67 @@ describe("createTelegramApi — per-chat send serialization (#154)", () => {
     expect(tracker.maxConcurrent).toBe(2);
   });
 });
+
+/**
+ * Hung long-poll recovery. A half-open connection (gateway 502/504 / network
+ * blip) leaves `fetch` neither resolving nor rejecting; without a client-side
+ * deadline `getUpdates` hangs forever and the poll loop freezes silently — the
+ * bot goes deaf with no error logged. `getUpdates` layers a client hard-timeout
+ * (`timeoutSeconds + buffer`) so a stuck request aborts and surfaces as a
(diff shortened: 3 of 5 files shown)


TerrysPOV (approved):
Approving. The client-side hard-timeout cleanly recovers the poll loop from a half-open getUpdates. AbortSignal.any composition correctly propagates caller signal abort (verified by the AbortError test on api.test.ts), so stop() latency is preserved. Tests cover both the TimeoutError path (loop self-heals + retries) and the clean stop() path. Three CLAUDE.md / cross-PR checks clean; no overlap with the chatTail serialiser from #155 or the per-chat 429 cooldown from #152.

PR #184, opened 2026-05-27
fix(bus): wire MCP multiplexer issuer before spawning bus agents (#165)

### LINKED ISSUE #165: Bus agents spawned before MCP multiplexer issuer is wired -> no --mcp-config (all mcp.shared servers unreachable)
## Summary
Bus-runtime agents spawn with **no `--mcp-config`**, so they cannot see ANY multiplexed MCP server declared in `settings.mcp.shared`. The agent silently falls back to default tool discovery (Bash / built-ins). This is a **startup-ordering bug**, not a config error.

## Impact (systemic — not one plugin)
Affects **every bus agent without a static `mcp_config`** (the normal case) and **every server in `mcp.shared`** — e.g. `brave-search`, `claudeclaw-tuner`, `gws-gmail`, `mistral-brain`. The entire `mcp.shared` mechanism is effectively dead for bus agents.

Symptoms: the agent "ignores" its MCP tools and shells out instead; any MCP observability/telemetry keyed on multiplexed tool calls shows **zero volume** (no multiplexed call ever happens). Often **latent**, because deployments commonly instruct the agent (via `CLAUDE.md`) to use CLI scripts, so the broken MCP path is never exercised.

## Root cause (`src/commands/start.ts`)
Startup order:
1. **~L506** — `mountBusRuntime({ agents: [...] })`. `src/bus/runtime-mount.ts:221-223` then loops `for (const agent of opts.agents) await sessionManager.spawnAgent(agent, origin)` — i.e. **eagerly spawns the agents at mount time.**
2. **~L790-791** — `pluginManager.startServices()` + `emit("gateway_start")` (gateway comes up).
3. **~L801-816** — `getMcpMultiplexerPlugin().start()`, then (only if `isActive()`) `injectMcpIdentityIssuer({...})` wires `_mcpIssueIdentity` into the PTY supervisor.

Agents are spawned in step 1, but the multiplexer + identity issuer only exist after step 3. At spawn time, `synthesizeMcpConfigIfActive` (`src/runner/pty-supervisor.ts`) hits its `if (!_mcpIssueIdentity) return undefined` guard → no config is written → `buildClaudeArgs` emits no `--mcp-config`. The agent PTY is long-lived, so it stays tool-less for its whole lifetime; later messages reuse the same PTY.

(Separately: the bus path has its own `buildClaudeArgs` at `src/bus/session-manager.ts:315` that only emits `--mcp-config` from a static `agent.mcp_config` and never synthesizes from `mcp.shared`. Wiring that synthesis in is inert on its own because of the ordering above.)

## Repro
1. `settings.mcp.shared` non-empty, `web.enabled: true`, an agent (e.g. `default`) with no `mcp_config`, `runtime: "bus"`.
2. Start the daemon; message the agent; ask it to use a multiplexed tool.
3. Observe: `ps` shows the agent's `claude` has no `--mcp-config`; the agent uses Bash instead; the multiplexer logs no tool call.

Instrumenting `synthesizeMcpConfigIfActive` confirms `issuer NOT wired` at spawn time.

## Suggested fix
The multiplexer (start + `injectMcpIdentityIssuer`) must be ready **before** the bus eagerly spawns agents:
- **Preferred — defer the eager spawn:** mount the bus without eager agents (`agents: []`), then spawn the configured agents *after* the issuer is wired (after ~L816). Smallest blast radius; verify nothing depends on agents existing at mount (heartbeats, adapter wiring, the `mountedAdapterNames` handle).
- **Or reorder the boot:** move gateway-start + multiplexer-start + issuer-wiring ahead of `mountBusRuntime`. Conceptually cleaner but cascades through the gateway / `startServices` dependency.

Either way, an explicit ordering guarantee (so it can't silently regress) would help.

## Test plan
- **Unit:** assert the issuer is wired (or multiplexer `isActive()`) before any `spawnAgent` call.
- **Integration:** with `mcp.shared` non-empty + an agent with no `mcp_config`, assert the spawned `claude` args include `--mcp-config` and the synthesized config lists the shared servers.
- **Regression:** with `mcp.shared: []` (dormant), assert NO `--mcp-config` is emitted (byte-identical to today).

Happy to follow up with a PR once you've picked the reorder approach.

Nibbler1250:
Fixes #165.

## Problem
Bus-runtime agents spawned at mount time (`start.ts` → `mountBusRuntime` → `runtime-mount.ts` synchronous spawn loop), **before** the MCP multiplexer identity issuer was wired (the issuer depends on `pluginManager.startServices()`, which runs after the mount). On top of that, the bus's own `buildClaudeArgs` (`session-manager.ts`) only emitted `--mcp-config` for a statically-configured `agent.mcp_config` — it never synthesized from `settings.mcp.shared` at all.

Net effect: every bus agent without a static `mcp_config` was launched with **no `--mcp-config`**, so all `mcp.shared` servers (`brave-search`, `gws-gmail`, `claudeclaw-tuner`, `mistral-brain`, …) were silently unreachable for the agent's entire lifetime. Latent because deployments often route work to CLI scripts instead of those MCP tools.

## Fix (Option A — defer the spawn, per your review on #165)
- **`runtime-mount.ts`** — add `deferSpawn` + `handle.spawnAgents()`. The daemon mounts the bus (slash relay, orphan scan, adapters) without spawning, then spawns agents after the issuer is wired. The spawn loop is refactored into a reusable batch helper with per-batch rollback. `mountBusRuntime` already accepted `agents: []`, so no API break.
- **`session-manager.ts`** — `setMcpConfigSynthesizer()` + `synthesizeBusMcpConfig()`. When the multiplexer is active and the agent has no static `mcp_config`, the bus spawn mints a per-agent HMAC identity, writes a per-PTY `--mcp-config` via the shared `pty-mcp-config-writer`, and revokes the identity + deletes the `0600` file on `stop()`. Static `mcp_config` still wins; a dormant multiplexer stays byte-identical (no flag).
- **`start.ts`** — mount with `deferSpawn`; wire the synthesizer + spawn agents + `BusScheduler` (deferred too — it targets the first spawned agent id) after the multiplexer block. On deferred-spawn failure, tear the bus down and fall back to legacy command surfaces, matching a mount failure.

## Tests
+13 tests, all green:
- `deferSpawn` / `spawnAgents()` ordering + rollback (issuer-before-spawn guarantee).
- `synthesizeBusMcpConfig` precedence (static wins), dormancy, and synthesized JSON shape (shared servers, bridge URLs, identity headers).
- Spawn integration asserting `--mcp-config <path>` lands in the child argv; **graceful-degradation** branch (multiplexer inactive → no flag, byte-identical to dormant) and zero-shared regression.
- `stop()` revokes the identity + deletes the synthesized config file.

Version guards bumped. Build + lint clean. Full suite adds 13 passing tests with zero new failures vs `origin/main`.

## Validated live
Deployed a downstream variant to a real `runtime: bus` daemon (`mcp.shared: [brave-search, claudeclaw-tuner, gws-gmail]`, `default` agent, no static config). After the change the agent spawns with `--mcp-config` listing all three servers (HTTP routes through the bridge, `0600`), and the multiplexer bridge accepts the synthesized per-PTY identity (401 without auth → request reaches the MCP server with it).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.108",
+      "version": "2.2.109",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.107",
+      "version": "2.2.108",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.106",
+      "version": "2.2.107",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.108",
+  "version": "2.2.109",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.107",
+  "version": "2.2.108",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.106",
+  "version": "2.2.107",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/mcp-config-synth.test.ts
+++ b/src/bus/__tests__/mcp-config-synth.test.ts
@@ -139,7 +139,7 @@ function writeArgvRecorder(dir: string): string {
   return path;
 }
 
-async function readArgvWhenReady(dir: string, timeoutMs = 2000): Promise<string[]> {
+async function readArgvWhenReady(dir: string, timeoutMs = 8000): Promise<string[]> {
   const path = join(dir, "argv.txt");
   const deadline = Date.now() + timeoutMs;
   while (Date.now() < deadline) {
--- a/src/bus/__tests__/mcp-config-synth.test.ts
+++ b/src/bus/__tests__/mcp-config-synth.test.ts
@@ -13,7 +13,15 @@
  *      `stop()` revokes the identity + deletes the 0600 config file.
  */
 import { afterEach, beforeEach, describe, expect, it } from "bun:test";
-import { chmodSync, existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import {
+  chmodSync,
+  existsSync,
+  mkdtempSync,
+  readFileSync,
+  realpathSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { PtyIdentity } from "../../runner/pty-mcp-config-writer";
@@ -100,6 +108,29 @@ describe("synthesizeBusMcpConfig (issue #165)", () => {
     expect(existsSync(configPathFor(cwd, "a"))).toBe(false);
   });
 
+  it("revokes the just-minted identity when writeConfigForPty throws", async () => {
+    // Maintainer follow-up on the PR #184 third re-review: synth.issue
+    // happens BEFORE writeConfigForPty, and the outer try/catch in
+    // spawnAgentInternal can't see the cwd because mcpConfigCwd is
+    // only assigned AFTER synthesizeBusMcpConfig returns. Without the
+    // inner rollback, an EACCES/ENOSPC during the file write leaks the
+    // multiplexer identity in the issuer's registry forever.
+    const { synth, issued, revoked } = makeSynth();
+    // Make the cwd read-only so writeConfigForPty hits EACCES on its
+    // internal writeFileSync.
+    chmodSync(cwd, 0o500);
+    try {
+      expect(() => synthesizeBusMcpConfig(agentCfg("triage", cwd), synth, cwd)).toThrow();
+    } finally {
+      // Restore for afterEach cleanup.
+      chmodSync(cwd, 0o700);
+    }
+    expect(issued).toEqual(["triage"]);
+    // `revoke` is fire-and-forget — give the microtask queue a tick.
+    await new Promise((resolve) => setTimeout(resolve, 10));
+    expect(revoked).toEqual(["triage"]);
+  });
+
   it("writes a per-agent config listing the shared servers with bridge URLs + identity headers", () => {
     const { synth, issued } = makeSynth();
     const out = synthesizeBusMcpConfig(agentCfg("triage", cwd), synth, cwd);
@@ -179,7 +210,10 @@ describe("SessionManager spawn → synthesized --mcp-config in argv (issue #165)
 
     const idx = argv.indexOf("--mcp-config");
     expect(idx).toBeGreaterThanOrEqual(0);
-    expect(argv[idx + 1]).toBe(configPathFor(cwd, "a"));
+    // SessionManager resolves `cwd` through realpathSync, so on macOS the
+    // synthesized argv path is under `/private/var/...` while `cwd` itself
+    // is `/var/...`. Compare against the resolved-path variant.
+    expect(argv[idx + 1]).toBe(configPathFor(realpathSync(cwd), "a"));
     expect(issued).toEqual(["a"]);
     expect(existsSync(configPathFor(cwd, "a"))).toBe(true);
   });
@@ -224,6 +258,12 @@ describe("SessionManager spawn → synthesized --mcp-config in argv (issue #165)
     expect(existsSync(configPathFor(cwd, "a"))).toBe(false);
     // revoke is fire-and-forget on a microtask; let it settle.
     await new Promise((r) => setTimeout(r, 25));
-    expect(revoked).toEqual(["a"]);
+    // PR #184 re-review (commit df52157) added cleanup-in-onExit alongside
+    // the pre-existing stop()-path cleanup, so a clean stop() observes
+    // both paths firing. The double-call is documented as safe per the
+    // idempotency contract on `cleanupAgentMcpConfig`; assert the agent
+    // appears in the revoked list rather than that it appears exactly
+    // once.
+    expect(revoked).toContain("a");
   });
 });
--- a/src/bus/__tests__/mcp-config-synth.test.ts
+++ b/src/bus/__tests__/mcp-config-synth.test.ts
@@ -1,0 +1,229 @@
+/**
+ * Issue #165 — the bus spawn path must synthesize a per-agent
+ * `--mcp-config` for `mcp.shared` servers (the legacy PTY supervisor
+ * already does; the bus's own `buildClaudeArgs` never did).
+ *
+ * Two layers of coverage:
+ *   1. `synthesizeBusMcpConfig` (pure): precedence + dormancy + the
+ *      synthesized JSON shape (shared servers, bridge URLs, identity
+ *      headers).
+ *   2. `SessionManager.spawnAgent` integration: the synthesized
+ *      `--mcp-config <path>` actually lands in the spawned process's argv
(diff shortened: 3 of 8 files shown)


TerrysPOV:
### Code review

5-agent review on `a07c7e9f`. Architecture is right (Option A, deferred eager spawn + synthesizer wiring) but found 5 blockers/important issues before merge.

**1. Version bump regressed — both version-guard CI checks will fail.** Bumped to `2.2.101` but `main` is already at `2.2.102`. Both `plugin-version-guard` and `marketplace-version-guard` compare against the merge base and fail on equal-or-lower. Rebase onto current main and re-run `bun run bump:plugin-version` + `bun run bump:marketplace-version` to get to `2.2.103`.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/a07c7e9f0b877772e7db808aefc252c19d9770be/.claude-plugin/plugin.json#L1-L5

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/a07c7e9f0b877772e7db808aefc252c19d9770be/.claude-plugin/marketplace.json#L8-L13

**2. Re-introduces the PR #123 Codex P2 prompt-loss race in the opposite direction.** The new boot order is mount(`deferSpawn:true`) → `wireBusAdapters` + `attachAdapters` (adapters begin polling) → ~3-400 lines of code~ → `plugin.start()` (HTTP round-trip, can be hundreds of ms) → `setMcpConfigSynthesizer` → `handle.spawnAgents()`. Adapters are live and dispatching while **zero agents are registered**. Inbound prompts during that window hit a bus with no targets and are silently dropped. The original Codex P2 invariant (adapters AFTER agents) is violated. Fix: move `wireBusAdapters` + `attachAdapters` down to after `spawnAgents()`, or buffer dispatch in `BusCore` until at least one agent registers.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/a07c7e9f0b877772e7db808aefc252c19d9770be/src/commands/start.ts#L523-L530

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/a07c7e9f0b877772e7db808aefc252c19d9770be/src/commands/start.ts#L886-L890

**3. The numbered boot-order comment that documents the fix is now stale.** Step 2 still says "mountBusRuntime — starts the IPC server + spawns agents" — with `deferSpawn: true` the mount no longer spawns. The comment exists precisely to be the authoritative ordering reference and it's wrong about its own PR's new order. Same with the `runtime-mount.ts` Sprint 5.2a bullet (`Auto-spawn of named agents from settings.agents` with no `deferSpawn` qualification).

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/a07c7e9f0b877772e7db808aefc252c19d9770be/src/commands/start.ts#L489-L503

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/a07c7e9f0b877772e7db808aefc252c19d9770be/src/bus/runtime-mount.ts#L13-L17

**4. Collision-rotation retry leaks the synthesized `--mcp-config` file and the multiplexer identity.** The first `spawnAgentInternal` attempt has already called `synthesizeBusMcpConfig` for `agent.id` — which mints a multiplexer identity and writes a `0o600` config file. On session-id collision, the retry path clears the registry slot but never calls `this.mcpSynth?.revoke(agent.id)` or `deleteConfigForPty(realCwd, agent.id)` before the recursive retry. The retry's `synthesizeBusMcpConfig` mints a fresh identity (replacing the in-memory one) but the stale on-disk config file persists until daemon shutdown.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/a07c7e9f0b877772e7db808aefc252c19d9770be/src/bus/session-manager.ts#L558-L579

**5. `spawnAgents()` partial-failure rollback only stops agents, not adapters.** `spawnAgentBatch` correctly rolls back the agents in the failing batch. But by the time `handle.spawnAgents()` runs, `handle.attachAdapters` has already attached adapters to the handle (start.ts:528). On a `spawnAgents()` throw, those adapters stay live until the daemon's outer catch calls `handle.stop()` — adapters dispatch to a partially-populated agent set during that window. Either tear adapters down inside the `spawnAgents` failure path, or document this in the `spawnAgents` JSDoc contract.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/a07c7e9f0b877772e7db808aefc252c19d9770be/src/bus/runtime-mount.ts#L260-L292

**6. (minor) Dormant-multiplexer with non-empty `mcp.shared` silently boots agents without `--mcp-config`.** When `mcp.shared` is non-empty but `web.enabled=false`, the entire multiplexer block is skipped and `setMcpConfigSynthesizer` is never wired — agents launch without `--mcp-config` and the operator has no log indication. The same silent-failure mode the PR set out to fix, resurfaced under a different condition. One `console.warn` would close it.

Verified clean: dual `--mcp-config` flag is correctly gated by `if (agent.mcp_config) return undefined` in `synthesizeBusMcpConfig`; `writeConfigForPty` already enforces `0o600` + explicit `chmodSync` (inherited from PR #131); `wireBusScheduler` is now correctly post-spawn; the `@inline` plugin tag and the multiplexer-stop placement are untouched.

Nibbler1250:
Thanks for the thorough review — all 6 addressed. Force-pushed (rebased onto current `main`, now at `aae0a99`).

1. **Version guards** — rebased onto `main` (2.2.103) and re-bumped both to **2.2.104**.
2. **Prompt-loss race (opposite direction)** — moved `wireBusAdapters` + `attachAdapters` out of the mount block and into the deferred block, **after** `spawnAgents()`. Adapters no longer start polling while zero agents are registered, restoring the PR #123 Codex P2 invariant (adapters after agents). Added a pure `configuredBusAdapterNames(settings)` in `adapter-wiring.ts` (same token/busRouting predicates as the mount fns) so the startup banner + legacy-skip logs — which run before the now-deferred wiring — still report the right platforms.
3. **Stale boot-order comments** — updated the numbered order in `start.ts` (step 2 no longer claims it spawns) and the `runtime-mount.ts` Sprint 5.2a bullet (now documents `deferSpawn`).
4. **Collision-rotation leak** — extracted `cleanupAgentMcpConfig()` and call it on both the rotation-retry and persisted-failure paths, so the first attempt's multiplexer identity is revoked and its 0600 config file deleted before the retry re-synthesizes.
5. **`spawnAgents()` partial-failure rollback** — resolved by #2: adapters are now attached inside the same deferred `try`, after `spawnAgents()`, so a spawn throw happens before any adapter is attached (and the catch's `handle.stop()` tears down anything that was).
6. **Dormant-but-configured silent boot** — added a `console.warn` when `mcp.shared` is non-empty but `web.enabled` is false (multiplexer can't mount → no `--mcp-config`).

Also re-verified live on a downstream `runtime: bus` daemon after the changes.

TerrysPOV:
### Code review

Re-review on `aae0a992fd688bd82c88e93783c78cb49e55375e`. All 6 issues from the prior review verified addressed. Three new issues found.

**1. Identity + 0600 bearer file leaked when `spawnPty`/`spawnChild`/`spawnTmux` throws.**

`synthesizeBusMcpConfig` mints the multiplexer identity and writes the 0600 bearer file *before* the spawn dispatch. If the spawn primitive throws (bun-pty native fault, ENOENT on `commandOverride`, tmux not on PATH), control unwinds through `spawnAgentBatch` which only tears down *previously-successful* peers via `sessionManager.stop(id)` — the just-failed spawn's synthesized identity is never revoked and the 0600 file is never deleted. Fix: wrap the spawn dispatch in `try { ... } catch (err) { this.cleanupAgentMcpConfig(mcpConfigCwd, agent.id); throw err; }`.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/aae0a992fd688bd82c88e93783c78cb49e55375e/src/bus/session-manager.ts#L505-L525

**2. Natural PTY exit (crash / OOM / operator-kill) does not call `cleanupAgentMcpConfig`.**

The `proc.onExit(...)` handler at spawn time clears the registry entry but does *not* invoke `cleanupAgentMcpConfig`. Cleanup only fires on `stop()`. When `claude` exits on its own — crash, claude-side auth failure, operator `kill`, OOM — the 0600 bearer file persists and the multiplexer identity is never released (handler-side `releasePty` cleanup for that ptyId never runs). The `cleanupAgentMcpConfig` JSDoc already promises idempotency, so adding the same call inside `onExit` is safe alongside the existing `stop()`-path cleanup.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/aae0a992fd688bd82c88e93783c78cb49e55375e/src/bus/session-manager.ts#L537-L545

**3. Orphaned JSDoc on `stop()` — the new `cleanupAgentMcpConfig` JSDoc landed *between* the pre-existing `stop()` Spike-0.5 docblock and the `stop()` method, leaving `stop()` undocumented and `cleanupAgentMcpConfig` with two stacked docblocks.

Fix: move the Spike-0.5 docblock down to immediately precede `stop()` (line 768), or merge it into the new `cleanupAgentMcpConfig` block.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/aae0a992fd688bd82c88e93783c78cb49e55375e/src/bus/session-manager.ts#L730-L770

**Verified resolved from prior review:**

- Version bump 2.2.103→2.2.104, both manifests
- Adapters now mount *after* `spawnAgents()` (PR #123 Codex P2 prompt-loss invariant preserved)
- Boot-order comments + Sprint 5.2a reference updated to match the new sequence
- `cleanupAgentMcpConfig` extracted and invoked on rotation-retry + persisted-collision + stop
- Rollback covers deferred-spawn failures via `handle.stop()` then legacy fallback
- `console.warn` for the dormant-but-`mcp.shared`-configured silent-boot case

**Verified false positive (Agent #2 raised, I disagreed):**

- Revoke/re-issue race on session-id collision rotation — `releaseIdentity` does the identities-map `delete` synchronously at function entry, before any await. Recursive respawn's `issueIdentity` runs after; no window for the late revoke to clobber the fresh identity.

Nibbler1250:
Thanks — all 3 fixed. Force-pushed (rebased onto current `main`, now `df52157`, version 2.2.107). All three are in `session-manager.ts`:

1. **Spawn-primitive throw leak** — wrapped the `spawnPty`/`spawnChild`/`spawnTmux` dispatch in `try { … } catch (err) { this.cleanupAgentMcpConfig(mcpConfigCwd, agent.id); throw err; }`, so a throw before registry insertion still revokes the identity + deletes the 0600 file.
2. **Natural PTY exit** — the `proc.onExit(...)` handler now calls `cleanupAgentMcpConfig(current.mcpConfigCwd, agent.id)` before dropping the registry entry, so crash/OOM/operator-kill releases the identity + file (idempotent, safe alongside the `stop()` path).
3. **Orphaned JSDoc** — moved the Spike-0.5 docblock back onto `stop()`; `cleanupAgentMcpConfig` keeps its own single docblock.

Re: tests — `cleanupAgentMcpConfig` is exercised by the existing `stop() revokes + deletes` test; the two new call sites reuse that same idempotent helper. I left the spawn-failure/natural-exit paths without dedicated tests since reliably forcing a bun-pty spawn fault is flaky, but happy to add them if you'd prefer.

Thanks for the careful passes on this one.

TerrysPOV:
@codex review

Issue #193, opened 2026-05-28
Fresh install: bus agent never responds — missing agents[]/busRouting + new Bypass-Permissions dialog kills the agent


Nibbler1250:
## Summary

On a fresh ClaudeClaw-Plus install (bus runtime, the default), the agent never responds on Telegram/Discord/Slack. Tracing this with @GordonWu in moazbuilds/claudeclaw#216 surfaced **three layered causes** — two are setup/wizard gaps, one is a new Claude CLI dialog the spawn supervisor mishandles. The surface looks healthy (process up, channel polling) which is why it's hard to spot.

## Root causes (confirmed in the field)

**1. Bus mounts with no agent.** The `/claudeclaw-plus:start` wizard doesn't write `settings.agents[]`, so the bus runtime mounts with **zero agents** (startup banner: `no agents`).

**2. Telegram adapter won't mount without routing config.** Even with `telegram.token` set, the bus Telegram adapter refuses to mount unless `telegram.busRouting` is present (per @GordonWu's trace, `adapter-wiring.ts:149`) → startup banner: `no adapters`.

**3. New "Bypass Permissions mode" dialog kills the agent.** Recent Claude CLI (2.1.x) shows a `WARNING: Bypass Permissions mode` confirmation whose default selection is **"No, exit"**. The spawn supervisor auto-dismisses startup dialogs with a blind `\r` — that handler was written for the older `--dangerously-load-development-channels` dialog (where the default is "I am using this for local development", so Enter = accept). On the **new** dialog, `\r` selects **exit** → every spawned agent dies immediately. Tell-tale: `No MCP connection for agent_id=<name>` in the log, and `pgrep -f 'claude --plugin-dir'` shows nothing.

## Workaround (per @GordonWu)

- `~/.claude/settings.json`: `"skipDangerousModePermissionPrompt": true` (post-migration name of `bypassPermissionsModeAccepted`)
- `<cwd>/.claude/claudeclaw/settings.json`:
  ```jsonc
  {
    "agents": [{ "id": "main" }],
    "telegram": { "token": "…", "busRouting": { "chats": {}, "defaultAgentId": "main" } }
  }
  ```

## Proposed fix (PR incoming)

1. **Wizard** — when `telegram` is configured, auto-write a default `agents: [{ id: "main" }]` and a `telegram.busRouting` block (`defaultAgentId` → the declared agent), so a fresh install boots a working agent + adapter without manual edits.
[...]

PR #195, opened 2026-05-28
fix(bus): answer the Bypass-Permissions startup dialog so fresh-install agents survive

Nibbler1250:
## Problem

On a fresh install (bus runtime, the default), the agent spawns but **dies immediately** and never responds on any channel — the log shows `No MCP connection for agent_id=<name>` and `pgrep -f 'claude --plugin-dir'` finds nothing.

Root cause: recent Claude CLI shows a `WARNING: Bypass Permissions mode` confirmation whose **default selection is "No, exit"**:

```
❯ 1. No, exit
  2. Yes, I accept
```

The bus spawn dismissed startup dialogs with a **blind Enter** (`pty.write("\r")` on fixed timers) — a handler written for the older `--dangerously-load-development-channels` dialog, whose default IS the accept option. On this new dialog the blind Enter selects **"No, exit"**, killing every spawned agent at boot.

Confirmed in the field via moazbuilds/claudeclaw#216 (thanks @GordonWu for the breakdown). Tracked as #193.

## Fix

Replace the blind dialog-dismiss timers with an **output-driven watcher** in `PtyAgentProcess`. It inspects early PTY output and sends the **correct key per dialog**:

- dev-channels confirmation (default = accept) → `Enter`
- "Bypass Permissions mode" (default = "No, exit") → `Down` + `Enter` to land on "Yes, I accept"

Scoped to the daemon's own spawn — **no global `~/.claude/settings.json` mutation**, so a user's interactive `claude` sessions keep their safety prompts. Matches on text unique to each dialog (`"Yes, I accept"` / `"development channels"`) so the REPL footer — which also contains "bypass permissions on" — can't false-trigger. Disengages once a real prompt is dispatched.

- `src/bus/session-manager.ts` — drop the blind `[1000,2500,5000,8000]ms` `\r` timers.
- `src/bus/session-agent-process.ts` — `handleBootDialog()` watcher + `endBootDialogPhase()`.
- tests — new `pty-boot-dialog.test.ts` (per-dialog keys, footer no-false-trigger, once-only); existing `session-agent-process.test.ts` updated to assert the watcher disengages on the first prompt.

## Verification

- The accept keystroke (`Down`+`Enter`) was verified against the **real** Bypass-Permissions dialog (reproduced on a fresh authed install): it selects "Yes, I accept" and reaches the REPL.
- `bun test` → boot-dialog (4) + session-agent-process (2) + watchdog smoke (17) all green.
- `bun run lint` (biome) clean.

## Scope

Surgical fix for the agent-killer. The companion setup-wizard gaps from #193 (auto-writing `agents[]` + `telegram.busRouting` when Telegram is configured) are a follow-up.

Refs #193

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.121",
+      "version": "2.2.122",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.120",
+      "version": "2.2.121",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.119",
+      "version": "2.2.120",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.121",
+  "version": "2.2.122",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.120",
+  "version": "2.2.121",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.119",
+  "version": "2.2.120",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/__tests__/pty-boot-dialog.test.ts
+++ b/src/__tests__/pty-boot-dialog.test.ts
@@ -1,0 +1,71 @@
+/**
+ * Tests for the boot-dialog watcher (issue #193).
+ *
+ * PtyAgentProcess answers claude's interactive startup confirmation dialogs by
+ * inspecting early PTY output and sending the correct key per dialog: Enter for
+ * the dev-channels confirmation, and Down+Enter for the new "Bypass Permissions
+ * mode" dialog (whose default is "No, exit" — a blind Enter would kill the
+ * agent).
+ */
+import { describe, test, expect } from "bun:test";
+import { PtyAgentProcess, type PtyHandle } from "../bus/session-agent-process";
+
+function makeFakePty() {
+  const writes: string[] = [];
+  let dataCb: ((d: string) => void) | null = null;
+  const pty: PtyHandle = {
+    pid: 4242,
+    onData(cb) {
+      dataCb = cb;
+      return { dispose() {} };
+    },
+    onExit() {
+      return { dispose() {} };
+    },
+    write(d) {
+      writes.push(d);
+    },
+    kill() {},
+  };
+  return { pty, writes, emit: (s: string) => dataCb?.(s) };
+}
+
+describe("PtyAgentProcess boot-dialog watcher (issue #193)", () => {
+  test("answers the Bypass Permissions dialog with Down then Enter", async () => {
+    const { pty, writes, emit } = makeFakePty();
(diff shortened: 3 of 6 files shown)


TerrysPOV:
Thanks for this @Nibbler1250 — the fix is solid. `ensureDangerousModeAccepted()` mirrors `ensureTrustAccepted` exactly (atomic write, malformed-JSON safe, idempotent, returns instead of throws), writes both the current and pre-migration flag names, and the 5 tests cover create / preserve / no-op / malformed / homedir. Clean.

**One thing to make clear — scope.** Issue #193 has **three** separate root causes:

1. **Wizard writes no `settings.agents[]`** → bus mounts with `no agents`. (Now tracked in #196.)
2. **Telegram adapter won't mount without `telegram.busRouting`** → `no adapters`, bot silent even with a valid token. (Now tracked in #197.)
3. **New Bypass-Permissions dialog kills the agent at spawn** → this PR. ✅

So **this PR fully fixes cause #3 and only #3** — and that is the right scope. It is a clean, standalone fix. Causes #1 and #2 are separate wizard/config gaps and should land in their own PRs (#196, #197) so each gets its own review. You do **not** need to expand this PR to cover them.

**On "accept / not accept":** I think the open question was whether this PR is "enough" to close #193 — it is enough for cause #3, but #193 itself stays open until #196 + #197 also land. This PR should reference cause #3, not fully close #193.

**Maintainer commit:** the only red CI was the version guards (the branch version still matched the merge base). I pushed a version bump (`6d199fa`) — no code change — so CI can go green. I'm also taking it out of draft so the full CI + Codex review run.

**Over to you:** can you let Terrence know on Discord whether **you** want to tackle the follow-ups (#196 wizard `agents[]`, #197 Telegram `busRouting`), or whether you'd like **him** to pick them up? Either is fine — just so we don't both grab the same one.

TerrysPOV:
**Correction / re-review after the force-push** — my earlier comment described approach **A** (the global `skipDangerousModePermissionPrompt` flag), which is no longer on the branch. Reviewed the current HEAD (**B**, the dialog-aware watcher). No apology needed @Nibbler1250 — the switch was the right call.

**Decision: go with B.** Terrence's call, and we agree.

**Why B over A:** A's concern is real — writing `skipDangerousModePermissionPrompt: true` into the user's *global* `~/.claude/settings.json` disables the bypass-permissions confirmation for the operator's **own interactive `claude` sessions** too, not just the daemon. A daemon-spawn fix shouldn't silently weaken a safety prompt for the human's whole CLI. B is correctly scoped: the dialog handling lives in `PtyAgentProcess` (the daemon's own spawned PTY), no global mutation, and it directly replaces the blind-`\r` that was the actual bug. Bonus: it simplified `session-manager.ts` by ~56 lines. Nice.

**One caveat to keep in mind (not a blocker):** the `\x1b[B` (Down) + `Enter` is positional — it assumes "Yes, I accept" sits below the default. If a future Claude CLI reorders the menu it'll break. That's acceptable because (a) it's the same CLI-coupling risk any approach has, and (b) B fails *visibly* (agent dies + logs) rather than A's *silent* global security downgrade. A possible future hardening: write the skip flag to a **daemon-scoped** settings file (project-local `.claude/settings.json` in the daemon cwd), which gets A's no-parsing robustness without the global side-effect — backlog, not now.

**Maintainer commit:** the force-push clobbered the earlier version bump, so I re-applied it (`4c41636`) — no code change — to clear the version-guard CI.

**Scope unchanged — both follow-ups still stand.** B fixes cause #3 only, same as A did; the switch doesn't touch causes #1/#2. So #196 (wizard writes `settings.agents[]`) and #197 (Telegram `busRouting`) remain required. Still keen to know on Discord whether you'll take those or want Terrence to.

Terrence will mark this ready + merge after Codex review.

TerrysPOV:
@Nibbler1250 — small correction to my last line: **you can take it out of draft yourself whenever you're ready**, and Terrence will review it for merge once it's marked ready (after Codex runs). No need to wait on him to flip it.

TerrysPOV:
@codex review

TerrysPOV:
**Maintainer patch — addressed the Codex P2** (@Nibbler1250 you're offline, Terrence asked me to push this so the PR isn't blocked; squash/adjust however you like).

Codex was right: `send_prompt_stream` disengaged the dialog watcher on the **first prompt**, but a prompt can be dispatched as soon as the agent is registered (`spawnAgentInternal` only waits out the collision-detect window, not REPL readiness). On a slow fresh-install boot where the dialog renders *after* an early heartbeat/scheduler prompt, the watcher was already off → dialog defaulted to "No, exit" → agent died. That's the exact regression the old timed dismissals avoided.

**Fix (`3ee92ce`):** disengage the watcher on a positive **REPL-ready signal** — the footer marker `"bypass permissions on"` (distinct from the dialog's `"Yes, I accept"`, and already in your false-trigger guard) — or a bounded **15s timeout**, instead of on first prompt. Since the watcher only ever writes in response to specific dialog text, leaving it engaged past an early prompt can't inject keys mid-turn once the REPL is up. Prompt dispatch is unchanged — no blocking, no added latency.

Updated the boot-dialog unit test to the new contract: a dialog rendering *after* an early prompt is still answered, then the watcher disengages on the REPL footer marker (a later stray chunk is ignored). 6/6 pass, lint clean, version bumped to 2.2.121.

One residual (not blocking, backlog-worthy): if a prompt's keystrokes are typed while a dialog is *still on screen*, those bytes land on the dialog rather than the REPL — same edge as before. Fully closing that would mean holding early prompts until the boot phase ends; left out here to keep the change minimal and avoid a startup-latency hit. Flagging for awareness.

Terrence will re-check Codex on this revision before merging.

TerrysPOV:
@codex review

TerrysPOV:
**Maintainer patch #2 — addressed the new Codex P2** (the "recognize non-bypass REPL footers" one on `3ee92ce`).

Valid finding. My REPL-ready marker from the last patch was `"bypass permissions on"`, which is **permission_mode-specific**. An agent configured with a non-bypass `permission_mode` (plan / acceptEdits / default / dontAsk / auto) shows a different footer (`"plan mode on"`, `"accept edits on"`, …), so the bypass-only marker never matched — the watcher then idled until the 15s timeout, and within that window a stray `"Yes, I accept"` / `"development channels"` model echo could re-trigger the dialog handler and inject Down/Enter into an already-live REPL.

**Fix (`5ff104f`):** match the **mode-independent** `"shift+tab to cycle"` hint that every mode footer carries, instead of the bypass-specific text. The watcher now disengages promptly across all six supported permission modes, and no dialog contains that hint so it can't false-trigger. The bounded 15s timeout stays as the backstop.

Test now drives a non-bypass (`"plan mode on"`) footer to lock in mode-independence. 6/6 pass, lint clean, version bumped to 2.2.122.

Both follow-ups (#196 wizard `settings.agents[]`, #197 Telegram `busRouting`) still stand — unchanged by this. Terrence will re-check Codex on `5ff104f` before merging.

TerrysPOV:
@codex review

Issue #207, opened 2026-05-31
feat(bus): per-message receipt chain to enable root-cause analysis of intermittent agent wedges


Nibbler1250:
## Pattern observed

Over the past few weeks the bus agent on Plus has wedged intermittently — agent process alive, daemon healthy, Telegram poll OK (409 conflict on manual `getUpdates`), but inbound messages hang and never return a reply. We've now seen this on Claude CLI **2.1.153, 2.1.156, and 2.1.158** — different CLI versions, same symptom. Restart sometimes clears it, sometimes doesn't.

The earlier framings in #195 (dialog) and #206 (CLI version) turned out to be too narrow — neither was the actual root cause of the recurring wedge.

## Why it's hard to debug from outside

The only operator-visible signal is *"inject hangs"* or *"Telegram message never gets a reply"*. There's no way to distinguish between:

- the prompt never reached the agent (PTY write lost / consumed by a transient overlay)
- the agent received it but never generated a turn
- the agent generated a turn but the bus's tailer never observed it
- the bus is serialising behind an earlier still-pending turn (lost first turn)

So we end up guessing causes (CLI version, dialog, session collision, mcp-config) and shipping workarounds for symptoms rather than fixing the root.

## Proposal — per-message receipt chain

Following @Snailflyer's idea on moazbuilds/claudeclaw#216, scoped concretely for Plus:

**Emit a structured receipt per inbound message**, recording each transition through the bus path:

```jsonc
{
  "message_id": "tg-12345",
  "received_at": "...",
  "selected_route": "agent=default",
  "agent_id": "default",
  "session_id": "...",
  "agent_cwd": "/home/.../agent",
  "process_pid": 84238,
  "process_generation": 3,            // bumps on respawn
  "prompt_hash": "sha256:...",        // proves what went in
  "claude_jsonl_path": ".../<sid>.jsonl",
  "turn_event_offset": null,          // populated when observed
  "final_state": "wedged_prompt"      // see enum
}
```

**`final_state` enum:**

- `message_polled` — adapter received it (Telegram 409 only confirms up to here)
- `route_resolved` — routing rule matched, agent selected
- `stdin_written` — bus wrote prompt + CR to the PTY
- `turn_observed` — JSONL tailer saw the matching assistant turn (**real success**)
- `wedged_prompt` — `stdin_written` ✓ but `turn_observed` never fired within timeout ← this issue's surface
- `stale_session` — session-id rejected at spawn / rotated
[...]

PR #209, opened 2026-05-31
feat(bus): receipt chain for bus → PTY observability (refs #207)

Nibbler1250:
## Problem

Issue #207: when an inbound message produces no agent turn (intermittent
wedge), we can't tell *where* the chain broke — bus, PTY write, the
agent's REPL, or the JSONL tailer — without instrumenting at debug time.
Today's logs let us see *that* something hung; they don't let us count
*how often* each surface fails, or correlate a specific message to the
exact pid that should have started a turn.

## Approach

A per-message **receipt** opens at the bridge entry and accumulates
stamps as the message moves through the bus → PTY pipeline:

```
  message_polled → route_resolved → stdin_written → turn_observed
                                                    (or wedged_prompt /
                                                     stale_session /
                                                     timeout)
```

Each receipt closes on a terminal `final_state` and is appended as one
JSON line to `~/.claude/claudeclaw/receipts.jsonl` (mode `0600`) for
offline analysis. Failure surfaces are **distinct terminal states** so
they're countable: `stale_session` = "PTY rejected the write or the
agent wasn't registered", `timeout` = "no `final` event in the window",
`wedged_prompt` = "the bus refused the prompt up-front".

## Scope of this PR — wired end-to-end

Three sites are instrumented:

1. **`src/bus/receipt.ts`** — `ReceiptStore` + `createReceiptStore` +
   `getDefaultReceiptStore()` (process-wide singleton) +
   `findByPromptHash(hash)` (secondary index used by the bus → PTY
   seam).
2. **`src/bus/webui-bridge.ts` (`streamBusPrompt` / `runPrompt`)** —
   opens a receipt at entry with `agent_id`, `selected_route`, and a
   short `prompt_hash`; closes on `intent: "final"` as `turn_observed`,
   on the timeout as `timeout`, or on a `bus.sendPrompt` rejection as
   `wedged_prompt`.
3. **`src/bus/receipt-wiring.ts` (`createPromptStreamHandler`) wired
   into `runtime-mount.ts`** — at the bus → PTY seam, looks the receipt
   up by `prompt_hash`, back-fills `process_pid` + `process_generation`
   + `route_resolved_at`, stamps `stdin_written_at` on successful PTY
   write, or closes as `stale_session` on PTY write failure / no agent
   registered.

`runtime-mount.ts` itself stays small (one helper call); the wiring
helper is extracted so its pid/generation/error paths can be unit-tested
without booting the full bus.

## Security / compliance posture

- Prompt text is **not** stored — only a 16-char SHA-256 prefix
  (`hashPrompt`), enough to correlate a receipt to a specific message
  without leaking content into observability logs.
- The receipts log is **chmod 0600** after the first append; default
  umask would otherwise leave it group-readable, which is loose for a
  log that carries operational metadata (session id, pid, cwd-derived
  fields, prompt hash).
- The store **never throws**: filesystem errors are forwarded to a
  caller-supplied `onError`, defaulting to a single stderr line.
  Instrumentation must not be able to break the bus.
- Identity fields (`message_id`, `received_at`, `final_state`,
  `duration_ms`) cannot be overridden via `patch()` — a typo in caller
  fields can't change which receipt is which or close it prematurely.
- `close()` is idempotent — first call wins, second is a no-op. Prevents
  a buggy caller from double-appending or transitioning a receipt
  through two terminal states.

## Behaviour change

- **None for the success path.** Receipts are write-only — no caller's
  control flow depends on whether a receipt opened or closed
  successfully.
- A new `~/.claude/claudeclaw/receipts.jsonl` file appears on first use.
  No retention/rotation in this PR — operators can `truncate` /
  `logrotate` it externally if needed (followup candidate).

## Tests

`bun test src/bus/__tests__/{receipt,receipt-wiring,webui-bridge}.test.ts` →
**35 → 43 pass / 0 fail** (18 new tests added by this PR):

- **`receipt.test.ts`** — store basics, hash index transitions, singleton
  swap, identity-field guards, 0600 mode, write-failure forwarding,
  duration accuracy, drain.
- **`receipt-wiring.test.ts`** — pid back-fill, generation bump on pid
  change / stable on same pid, `stale_session` on PTY write error,
  `stale_session` when agent unregistered or no `send_prompt_stream`,
  no-receipt back-compat path.
- **`webui-bridge.test.ts`** — three new `streamBusPrompt — receipt
  chain` cases asserting `turn_observed` on final, `timeout` on no
  reply, and that the open receipt is discoverable by `findByPromptHash`
  during the prompt's in-flight window. A `beforeAll/afterAll` swap
  redirects the singleton to a tmp path so the existing 11 bridge tests
  don't append to the real receipts log.

`bun test src/bus` → **308 pass / 0 fail / 1 skip** (env-gated
`SCHEMA_PROBE_INTEGRATION`). Zero regressions vs `upstream/main`
(2054 → 2072 tests, same pre-existing fail count outside `src/bus`).

A standalone functional smoke at `scripts/smoke_receipt.ts` (21/21
checks: default path, single open/close, 50-way concurrency, idempotent
close, re-open merging, drain, write-failure surfacing, `hashPrompt`
stability, 0600 verification) is kept local-only — not part of the
committed test surface, but useful when iterating on the module.

`bun x biome check` clean on all touched files. `bun x tsc --noEmit`
introduces **zero** new errors (337 pre-existing on `upstream/main`,
337 on this branch — none in the touched files).

## What this PR doesn't do

- **Adapter-side `message_polled` stamps.** Adapters (telegram, discord,
  slack) and `/api/inject` currently call `bus.sendPrompt` directly,
  bypassing `streamBusPrompt`. They'd open their own receipt with the
  adapter's `message_id` (e.g. `tg-<update_id>`) before delegating.
  Tracked as a followup to keep this PR scoped to the bus interior.
- **JSONL-tailer-side `turn_event_offset` back-fill.** The
  `turn_observed` state is currently inferred from `response.text` with
  `intent: "final"`; recording the byte offset where the tailer
  *actually saw* the `turn_duration` event would prove the tailer side
  too. Cheap to add once the schema settles.
- **Retention / rotation** of `receipts.jsonl`.

Refs #207.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.132",
+      "version": "2.2.133",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.132",
+  "version": "2.2.133",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/receipt-wiring.test.ts
+++ b/src/bus/__tests__/receipt-wiring.test.ts
@@ -1,0 +1,138 @@
+/**
+ * Tests for the bus → PTY receipt wiring helper (issue #207). Verifies the
+ * pid/generation back-fill and the stale_session rollback paths without
+ * booting the full bus runtime.
+ */
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { existsSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { createReceiptStore, hashPrompt, type ReceiptRecord, type ReceiptStore } from "../receipt";
+import { type AgentProcessLike, createPromptStreamHandler } from "../receipt-wiring";
+
+let tmpDir: string;
+let logPath: string;
+let store: ReceiptStore;
+
+beforeEach(() => {
+  tmpDir = mkdtempSync(join(tmpdir(), "receipt-wiring-"));
+  logPath = join(tmpDir, "receipts.jsonl");
+  store = createReceiptStore({ path: logPath });
+});
+
+afterEach(() => {
+  if (existsSync(tmpDir)) rmSync(tmpDir, { recursive: true, force: true });
+});
+
+function readReceipts(): ReceiptRecord[] {
+  if (!existsSync(logPath)) return [];
+  return readFileSync(logPath, "utf-8")
+    .split("\n")
+    .filter((l) => l.trim().length > 0)
+    .map((l) => JSON.parse(l) as ReceiptRecord);
+}
+
+function fakeAgent(pid: number, sendImpl?: (line: string) => Promise<void>): AgentProcessLike {
+  return {
+    pid,
+    send_prompt_stream: sendImpl ?? (() => Promise.resolve()),
+  };
+}
+
+describe("createPromptStreamHandler", () => {
+  test("writes to the PTY when no receipt is open (back-compat)", async () => {
+    let captured: string | null = null;
+    const handler = createPromptStreamHandler(
+      () =>
+        fakeAgent(111, async (t) => {
+          captured = t;
+        }),
+      { store },
+    );
+    await handler("alpha", "hi");
+    expect(captured).toBe("hi");
+    expect(readReceipts()).toHaveLength(0);
+  });
+
+  test("back-fills pid + generation + route_resolved_at when a receipt is open", async () => {
+    const r = store.open("m1", { agent_id: "alpha", prompt_hash: hashPrompt("hi") });
+    const handler = createPromptStreamHandler(() => fakeAgent(2222), { store });
+    await handler("alpha", "hi");
+    expect(r.record.process_pid).toBe(2222);
+    expect(r.record.process_generation).toBe(1);
+    expect(r.record.notes?.route_resolved_at).toBeDefined();
+    expect(r.record.notes?.stdin_written_at).toBeDefined();
+  });
+
+  test("generation bumps when the same agent_id is seen with a different pid", async () => {
+    const handler = createPromptStreamHandler(() => fakeAgent(currentPid), { store });
+    let currentPid = 100;
+    const r1 = store.open("m-a", { agent_id: "alpha", prompt_hash: hashPrompt("a") });
+    await handler("alpha", "a");
+    expect(r1.record.process_generation).toBe(1);
+    expect(r1.record.process_pid).toBe(100);
+
+    currentPid = 200;
+    const r2 = store.open("m-b", { agent_id: "alpha", prompt_hash: hashPrompt("b") });
+    await handler("alpha", "b");
+    expect(r2.record.process_generation).toBe(2);
+    expect(r2.record.process_pid).toBe(200);
+  });
+
+  test("generation stays stable when the same pid is observed twice", async () => {
+    const handler = createPromptStreamHandler(() => fakeAgent(777), { store });
+    const r1 = store.open("m-1", { agent_id: "alpha", prompt_hash: hashPrompt("one") });
+    await handler("alpha", "one");
+    const r2 = store.open("m-2", { agent_id: "alpha", prompt_hash: hashPrompt("two") });
+    await handler("alpha", "two");
+    expect(r1.record.process_generation).toBe(1);
+    expect(r2.record.process_generation).toBe(1);
+  });
+
+  test("closes receipt as stale_session on PTY write error and re-throws", async () => {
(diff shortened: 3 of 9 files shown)


Nibbler1250:
### Live-deployed and validated end-to-end

Wired the patched files into a real bus daemon and hit `/api/inject`
three times in a row. Result in `~/.claude/claudeclaw/receipts.jsonl`
(redacted operational fields below):

```jsonl
{"message_id":"webui:inject:…","final_state":"timeout",       "process_pid":…,"process_generation":1,"notes":{"route_resolved_at":"…","bus_send_ok":true,"stdin_written_at":"…","timeout_ms":300000,"accumulated_chars":0},"duration_ms":300000,"prompt_hash":"sha256:…"}
{"message_id":"webui:inject:…","final_state":"turn_observed", "process_pid":…,"process_generation":1,"notes":{"route_resolved_at":"…","bus_send_ok":true,"stdin_written_at":"…","output_chars":8}, "duration_ms":6072,"prompt_hash":"sha256:…"}
{"message_id":"webui:inject:…","final_state":"turn_observed", "process_pid":…,"process_generation":1,"notes":{"route_resolved_at":"…","bus_send_ok":true,"stdin_written_at":"…","output_chars":6}, "duration_ms":4487,"prompt_hash":"sha256:…"}
```

The first request demonstrates exactly the diagnostic the chain is
designed to expose: **`stdin_written_at` is stamped** (the bus seam
successfully wrote the prompt to the agent's PTY) but **`final_state` is
`timeout`** with `accumulated_chars: 0` (the agent never produced a
turn). The next two requests on the same `process_pid` /
`process_generation` complete in seconds — so the wedge is *agent-side*
in that window, not a bus-routing or PTY-write failure.

Without the receipt chain, the operator would see "request timed out"
on the adapter and could not tell whether the bus, the PTY write, or
the agent itself was at fault.

### Followup commit

`fix(bus/receipt): unwrap <channel> block before hashing in handler` —
`BusCoreImpl.sendPrompt` wraps the prompt text in a `<channel ...>...
</channel>` block *before* invoking the stream prompt handler. The
receipt is opened with the hash of the **raw** prompt text, so the
handler must unwrap the channel block before computing `hashPrompt(text)`
or `findByPromptHash` returns `undefined` and the PTY-side fields
(`process_pid`, `process_generation`, `route_resolved_at`,
`stdin_written_at`) never get back-filled. Caught on first live inject;
fixed + unit-tested (13/13 pass on `receipt-wiring.test.ts`).

TerrysPOV:
Reviewed conceptually + architecturally (not the 5-agent technical pass — happy to run that separately if useful).

Solid PR. The state-machine + secondary hash index + never-throws + idempotent-close shape is the right one, the test discipline is dense (18 new, real coverage not box-ticking), and the dogfood loop has already paid off once (channel-unwrap bug caught + fixed + unit-tested in this same PR).

Two small pre-merge asks, then two follow-ups filed alongside this comment.

### Pre-merge (cheap)

1. **Document the enum rename vs #207's vocabulary.** A two-line comment near `ReceiptFinalState` noting that this PR's `wedged_prompt` ≠ #207's `wedged_prompt`: here it means "`bus.sendPrompt` rejected at entry"; in the issue, `wedged_prompt` was defined as `stdin_written` ✓ but `turn_observed` never fires within timeout — which is what this PR calls `timeout`. Not a bug, just stops every future reader of #207 + `receipts.jsonl` from getting confused.
2. **One-line caveat on `turn_observed`** that it's bridge-inferred from `intent: "final"` on `response.text`, not tailer-confirmed via `claude_jsonl_path` + `turn_event_offset`. The follow-up wires the JSONL offset; until then operators should know `turn_observed` ≈ "the bridge thinks the turn happened".

### Architectural concerns — deliberate MVP scope cuts, not blockers

1. **Adapter coverage.** Until Telegram/Discord/Slack/`/api/inject` open their own receipts, #207's headline symptom (Telegram message hangs) is *not* observed by this PR. The pattern works; the surface it covers (the webui bridge) isn't where #207 was filed from. Filing a follow-up to wire adapters immediately so this doesn't drift.
2. **`turn_observed` is bridge-inferred, not tailer-confirmed.** Filing a follow-up for the `turn_event_offset` JSONL-tailer back-fill so #207's "tailer never saw it" ambiguity gets closed.

### Smaller notes (not asks)

- `stale_session` is overloaded across three causes (PTY threw / agent unregistered / no `send_prompt_stream`). Cause is in `notes.reason` / `notes.stage`, so the data is there — just `final_state` alone won't slice them. Live with it for now; split later if querying by terminal state alone becomes the friction.
- `receipts.jsonl` grows unbounded. Acknowledged in the PR. Cheap follow-up when it matters.
- The factor-out of `createPromptStreamHandler` from `runtime-mount.ts` so the pid/generation/stale-session paths can unit-test without booting the bus is the right shape — and pays off in the new `receipt-wiring.test.ts`.

Approving in principle pending the two doc-only tweaks above. The live-deploy smoke transcript in the PR description (`stdin_written_at` stamped + `final_state: timeout` + `accumulated_chars: 0`) is exactly the diagnostic this whole chain exists to make legible — that's the validation that matters.

Nibbler1250:
Both doc tweaks applied in ecff7b9:

1. Vocabulary note near `ReceiptFinalState` mapping this PR's `wedged_prompt` ↔ #207's usage, plus that `#207::wedged_prompt` corresponds to this PR's `timeout`.
2. One-paragraph caveat that `turn_observed` is bridge-inferred from `intent: "final"`, not tailer-confirmed via `claude_jsonl_path` + `turn_event_offset` (pending the follow-up).

Both in the same comment block above the type definition to keep them co-located with what they qualify. No code change. Combined into one commit for cleanliness.

Re the two follow-ups you mentioned (adapter coverage + tailer-confirmed `turn_observed`): happy to take either if useful — I have a related issue I filed earlier today (#213) on bus-mode session bookkeeping that touches some of the same surface, so let me know if you want me to scope a follow-up to either of them while I'm in there.

Issue #213, opened 2026-06-02
streamBusPrompt skips incrementMessageCount — session rotation never fires on bus path


Nibbler1250:
## Summary

Routes that go through `src/bus/webui-bridge.ts::streamBusPrompt` (`/api/inject`, `/api/chat`, `/api/jobs/fire`) never call `incrementMessageCount()`. On daemons running with the bus runtime (default since the receipt-chain migration), the `messageCount` field in `.claude/claudeclaw/session.json` stays at its bootstrap value forever, and `needsRotation()` returns `false` regardless of volume — so session rotation never triggers.

## Reproduction

1. Start the daemon with bus runtime enabled.
2. Send N prompts via `/api/inject` (any N).
3. `cat .claude/claudeclaw/session.json` — `messageCount` stays at 1.
4. Any code path that depends on `needsRotation()` is dead.

## Root cause

`src/runner.ts:2157` and `:2604` call `await incrementMessageCount()` after each successful Claude invocation on the legacy path:

```ts
if (!agentName && !threadId) await incrementMessageCount();
```

`src/bus/webui-bridge.ts::streamBusPrompt` resolves on `payload.intent === "final"` (line ~152) but performs no equivalent bookkeeping. Since the bus path is now the default for the three webui routes named above, the counter is effectively dead on bus-mode deployments.

## Observable impact

1. **`session.jsonl` grows unbounded** — every Claude turn replays the full session context, so per-prompt latency creeps up over days/weeks of usage.
2. **Bootstrap loads stale summaries** — `loadLatestSummary()` picks the newest file in `summaries/`. With rotation dead, the newest file can be weeks old. The next bootstrap reads back an obsolete `Wakeup, my friend!` context, which then becomes the persistent baseline for the new session.
3. **Health probes false-positive as "wedged"** — once per-prompt latency exceeds typical probe timeouts (e.g. the 25 s default in some monitoring scripts), the daemon is reported as wedged when it's actually just slow.

Confirmed signature: on a Linux daemon running CLI 2.1.159 with a multi-week-old session, a `respond: pong` probe times out at 25 s while the daemon is still actively processing the request in the background (longer-timeout calls succeed). Last rotation in `summaries/` is dated weeks before, with no entries in between.

## Secondary issue (related)

[...]

PR #214, opened 2026-06-02
fix(bus): track messageCount and trigger rotation on bus path

Nibbler1250:
## Problem

Refs #213.

The webui bridge routes (`/api/inject`, `/api/chat`, `/api/jobs/fire`) never call `incrementMessageCount()` or `needsRotation()`/`rotateSession()`. On bus-mode deployments — which is the default for any daemon driving `streamBusPrompt` — the `messageCount` field in `.claude/claudeclaw/session.json` stays frozen at its bootstrap value and `needsRotation()` returns `false` regardless of volume. Sessions grow unbounded; subsequent bootstraps reload a `summaries/*.md` snapshot that may be weeks stale; per-prompt latency creeps up; monitoring scripts with short timeouts (e.g. 25 s) start reporting the daemon as "wedged" when it is actually just slow.

## Root cause

`src/runner.ts:2157` and `:2604` perform the bookkeeping on the legacy path:

```ts
if (!agentName && !threadId) await incrementMessageCount();
```

`src/bus/webui-bridge.ts streamBusPrompt()` resolves on `payload.intent === "final"` but performs no equivalent. The bus path replaces the legacy path for the three webui routes named above (see PR #136), so the counter has been dead on bus-mode deployments since that migration.

## Fix

Mirror the legacy bookkeeping at the bus seam: after `runPrompt` returns a successful result, increment the global session counter and check whether rotation should fire. The block is wrapped in `try/catch` so a bookkeeping failure cannot break the prompt flow the caller is awaiting.

`getSettings()` throws when called before `loadSettings()` (early boot, unit tests), so we skip rotation in that specific case but keep the increment. The next call after settings are loaded will pick it up.

This matches the legacy `runner.ts:1583-1589` pattern and uses the same `session` config (not a new key).

## Why option B (and not A/C from #213)

In the issue I proposed three options. This PR is **option B (agent-aware bookkeeping at the bus seam)** without the agent-vs-global session unification piece — that is a separate concern that affects what `rotateSession()` actually summarizes, and deserves its own discussion. **Option A** (single-line `incrementMessageCount`) would not have addressed the rotation trigger. **Option C** (moving the rotation check entirely out of `runner.ts`) is a bigger architectural change that I'd rather see decided in #213 discussion before opening another PR.

## Test plan

- [x] **T1 — Build / parse**: `bun build src/index.ts --outdir /tmp/claw-smoke --target bun --no-splitting` → clean
- [x] **T1 — Lint**: `bun x biome check src/bus/webui-bridge.ts` → clean
- [x] **T1 — Type check**: `bun x tsc --noEmit` → no new errors in `src/bus/webui-bridge.ts`
- [x] **T2 — Unit tests**: `bun test src/bus/__tests__/webui-bridge.test.ts` → 10 pass, 0 fail, no new noise (the `getSettings` early-skip prevents the test-env log spam)
- [x] **Live test on a real bus-mode daemon**:
  - Pre-restart: `session.json.messageCount = 1`, probe latency 25 s+ timeout, daemon flagged "wedged" 3× in 24 h.
  - Post-patch + restart: probe latency **3 s** OK, `messageCount` increments correctly on each successful `/api/inject` (1 → 2 → 3 verified).

## Scope and follow-ups (not in this PR)

- **Global vs agent-default session divergence** — `createSession()` at bootstrap creates a global session with a different `sessionId` than the agent the bus actually drives. `rotateSession()` will summarize the global (small bootstrap session) rather than the agent (the actually large one). Flagged in #213 "Secondary issue (related)"; happy to file a separate issue or take direction in this PR's review.
- **`createdAt` reset on every bootstrap** — makes `maxAgeHours`-based rotation ineffective for daemons that restart more than once per `maxAgeHours` window. Also flagged in #213.

## Version bump

Plugin + marketplace bumped to `2.2.133` per repo policy (`src/` touched).

## Migration notes

None — pure additive fix. Legacy `runner.ts` path is untouched; bus path now mirrors its bookkeeping.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.132",
+      "version": "2.2.133",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.132",
+  "version": "2.2.133",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/webui-bridge.ts
+++ b/src/bus/webui-bridge.ts
@@ -18,6 +18,9 @@
 
 import type { BusCore } from "./core";
 import type { BusOrigin } from "./types";
+import { incrementMessageCount, peekSession } from "../sessions";
+import { needsRotation, rotateSession } from "../rotation";
+import { getSettings } from "../config";
 
 /**
  * Per-agent mutex tail used to serialize `streamBusPrompt` calls. Codex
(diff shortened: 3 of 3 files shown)


Nibbler1250:
Closing in favour of a new PR per maintainer review on #213.

The agent-aware bus rotation design (issue-body B + comment-A) supersedes the global-only bookkeeping this PR implements. Quoting #213:

> A 1-line `await incrementMessageCount()` in `streamBusPrompt` bumps the *global* counter. The global session is the brief bootstrap "wakeup" turn — not the long-lived agent PTY where every Telegram / `/api/inject` / `/api/chat` turn accumulates. Flipping `needsRotation()` to true would then rotate the bootstrap session, write a summary of the wakeup turn, and leave the bloated agent session.jsonl growing untouched. That's a worse failure mode than the current silence.

Going to send the agent-scoped replacement: pass `agentId` through `incrementMessageCount`, `peekSession`, `needsRotation`, plus a new `rotateSession(config, agentName)` overload that writes summaries to the agent's directory. Legacy `runner.ts` no-arg call shape stays unchanged.

Ordering per #213: lands between #216 (preserve `createdAt`) and #217 (silent-drop safety net). Will post the new PR shortly.

Issue #215, opened 2026-06-02
Bus mode: final assistant text without explicit `reply` tool call is silently dropped


Nibbler1250:
## Summary

In bus-mode deployments, when an agent finishes a turn with only a final-assistant text block — **without** invoking the `reply` MCP tool with `intent: "final"` — the response is silently dropped. The text is persisted to the session `.jsonl` but never reaches any user-facing channel (Telegram, Discord, Slack, dashboard SSE).

There's no log, no fallback delivery, no operator warning. The user sees the agent as wedged when it actually completed the request.

## Relationship to existing issues

- **#211** (wire receipt chain into adapters) will make this *visible*: a turn that ends without `reply` would leave a receipt with `stdin_written` ✓ but no `turn_observed`. That's the diagnostic. **This issue is the defensive remedy** to actually deliver the response when the diagnostic fires.
- **#213** (`incrementMessageCount` bookkeeping gap on bus path) is adjacent — both describe state-tracking that depends on the agent calling tools explicitly, with no safety net when it forgets.

## Reproduction

1. Send a prompt to a bus-mode agent via a chat surface.
2. Have the agent process via Bash/Read tools, then emit a final assistant text block — but **without** invoking `reply`.
3. Check the session `.jsonl`: text present, turn closed.
4. Check the chat surface: nothing arrived.

Observed in the wild today: in the same multi-turn conversation, the agent correctly called `reply` on turn N (delivered) and forgot on turn N+1 (silent drop). Same agent, same session, same model. No `reply` ≠ no text — the text was there.

## Root cause

The bus-side delivery pipeline is **only** triggered by an explicit `reply` MCP tool call (`src/bus/mcp-server.ts handleReply` → bus `response.text` event → channel adapter dispatch). The `jsonl-tailer` also publishes `response.text` for every text block in the assistant stream (`src/bus/jsonl-tailer.ts:373`), but those events lack the channel `origin` that adapters use to decide whether to deliver — so they're filtered out. Only `reply`-originated `response.text` events carry the right tag.

The legacy `runner.ts` path doesn't have this issue: it captures the final stdout directly and delivers it. The bus path swapped that capture for a tool-call-driven flow but didn't carry over the "*something* always reaches the user" guarantee.

[...]

Issue #211, opened 2026-06-01
feat(bus): wire receipt chain into adapters (telegram/discord/slack) + /api/inject


TerrysPOV:
## Context

#209 landed the per-message receipt chain (refs #207) — receipts open at the webui bridge (`streamBusPrompt` / `runPrompt`), the bus → PTY seam back-fills pid / generation / `stdin_written_at` via `findByPromptHash`, and terminal states (`turn_observed` / `timeout` / `wedged_prompt` / `stale_session`) are appended to `~/.claude/claudeclaw/receipts.jsonl`.

#209 deliberately scopes the instrumentation to the bus interior — but the failure surface that motivated #207 (intermittent Telegram message hangs on Plus across CLI 2.1.153 / .156 / .158) lives behind the **adapters**, not the webui bridge. Adapters and `/api/inject` call `bus.sendPrompt` directly, bypassing `streamBusPrompt`, so they currently don't open receipts. Until they do, #207's headline symptom remains uninstrumented.

## Goal

Wire every inbound surface to open a receipt at `message_polled` so the chain becomes universal.

## Scope

- **Telegram adapter** — open receipt on inbound update with `message_id: "tg-<update_id>"`, `selected_route`, `agent_id`, `prompt_hash`, before delegating to the bus.
- **Discord adapter** — same shape, `message_id: "discord-<message_id>"`.
- **Slack adapter** — same shape, `message_id: "slack-<event_ts>"`.
- **`/api/inject` HTTP route** — same shape, `message_id: "webui:inject:<short-nonce>"` (mirroring what `runPrompt` does for the chat path).

Each site should:
1. Hash the prompt with `hashPrompt(text)` *before* the bus's `<channel ...>` wrapper is applied (so the bus → PTY seam's `findByPromptHash` still hits — same invariant #209's `unwrapChannelText` follow-up fix established).
2. Open the receipt with the adapter-side `message_id` and let the singleton store track it.
3. Defer the terminal close to whichever layer first observes a `response.text` + `intent: "final"` for the agent, or to the timeout path. (Or open + close at the adapter boundary if the adapter has its own request/response window — Telegram's `sendMessage` reply, for example.)

## Why this matters

The receipt chain's whole value is making intermittent wedges *legible*. #207's wedge is on Telegram. Until Telegram receipts exist, an operator hitting the wedge sees the same opaque "no reply" they had before #209.

## Design questions to resolve in implementation

[...]

PR #217, opened 2026-06-02
fix(bus): synthesize ingestReply on turn_end if reply tool was not called

Nibbler1250:
## Problem

Refs #215 (the issue this PR fixes), refs #211 (adapter receipt coverage that makes this measurable), refs #213.

In bus-mode deployments, when an agent ends a turn with `stop_reason: "end_turn"` and a final assistant text block **without** invoking the `reply` MCP tool, the user-facing surface (Telegram, Discord, Slack, webui) receives **nothing**. The text is persisted to the session `.jsonl` but never delivered.

**Confirmed live 2x in 12 hours** on the same daemon — the agent that wrote #215 the previous evening forgot to call `reply` the next morning, twice consecutively. Prompt-level enforcement alone is not sufficient. From #215:

> Compounding factor: no prompt explicitly tells the agent that `reply` is mandatory in bus mode. The tool description ("Use `intent: 'final'` for the turn-final message") reads like advice, not a hard requirement.

This PR adds the **code-level safety net** (option B of #215). Option A (prompt enforcement) is a complementary follow-up.

## Fix

A two-piece change at the bus seam:

### 1. JSONL tailer emits `response.turn_end`

`src/bus/jsonl-tailer.ts dispatchAssistant`: on an assistant line with `stop_reason: "end_turn"`, emit a new `response.turn_end` event carrying the concatenated text blocks (text-only, thinking excluded, joined + trimmed). `stop_reason: "tool_use"` keeps existing behavior — the turn continues after the tool result, no synthesis needed.

```ts
if (line.message?.stop_reason === "end_turn") {
  const turnText = blocks
    .filter((b) => b.type === "text")
    .map((b) => (b as { text?: string }).text ?? "")
    .join("\n")
    .trim();
  this.publish("response.turn_end", { stop_reason: "end_turn", text: turnText }, line);
}
```

### 2. Bus core tracks per-agent reply state + synthesizes on omission

`src/bus/core.ts`: add `currentTurnReplied: Map<agentId, boolean>`. Reset on every `sendPrompt`, set to `true` when `ingestReply` with `intent: "final"` lands. On `response.turn_end`, if still `false` AND turn produced non-empty text AND `lastPromptOrigin` is cached, synthesize an `ingestReply` so the user receives the response.

Cleared in cancel/error/disconnect paths alongside `lastPromptOrigin` to keep the map bounded.

```ts
private handleTurnEnd(agentId: string, text: string): void {
  if (this.currentTurnReplied.get(agentId) === true) return;
  if (!text || text.trim().length === 0) return;
  const origin = this.lastPromptOrigin.get(agentId);
  if (!origin) {
    console.warn(`[bus] silent-drop detected for agent=${agentId} ...`);
    return;
  }
  console.warn(`[bus] silent-drop recovered for agent=${agentId} ...`);
  this.currentTurnReplied.set(agentId, true); // single-flight guard
  this.ingestReply({ agent_id: agentId, text, intent: "final" });
}
```

### Why a warn-level log on synthesis

Operators can correlate with #211's receipts to measure the rate of agent-side reply-tool omission over time. The log line carries the agent_id, origin, and char count — enough to triage without exposing payload. If the synthesis path fires more than expected, that's a signal to revisit option A (prompt enforcement).

## Edge cases (all tested)

| Scenario | Behavior |
|---|---|
| Reply was called with `intent: "final"` | No synthesis (no duplicate delivery) |
| Turn ended with empty text | No synthesis |
| No `lastPromptOrigin` (cron / ambient turn) | Log warning, no synthesis (preserves current behavior — nowhere to route to) |
| Multi-prompt session | Flag resets per `sendPrompt`, single-flight per turn |
| Duplicate `response.turn_end` events (e.g. tailer replay) | Idempotent, single synthesis |
| Cancel / error / socket disconnect | Tracking cleared alongside `lastPromptOrigin` |

## Test plan

- [x] **T1 build**: `bun build src/index.ts --outdir /tmp/claw-smoke --target bun --no-splitting` → clean
- [x] **T1 lint**: `bun x biome check src/bus/core.ts src/bus/jsonl-tailer.ts src/bus/types.ts` → clean
- [x] **T1 type check**: `bun x tsc --noEmit` → no new errors in touched files (one pre-existing error in `src/skills-tuner/core/types.ts:19` is unchanged and unrelated)
- [x] **T2 unit tests new**:
  - `bus/__tests__/core.test.ts` → +5 tests covering synthesis/skip/dedup/reset
  - `bus/__tests__/jsonl-tailer.test.ts` → +2 tests covering turn_end emission and tool_use skip
- [x] **T2 unit tests regression**: full bus suite **283 pass, 0 fail, 1 skip** (no regression)
- [x] **Live-validated** on a real bus-mode daemon:
  - Pre-deploy: 2 confirmed silent drops in 12h (text in `.jsonl`, nothing on Telegram).
  - Post-deploy + restart: normal probe inject `pong` answered in 6s with `ok: true exitCode: 0`; no spurious `silent-drop` warnings (reply path used normally). The synthesis path only fires on omission — transparent in the common case.
  - Backups kept for fast rollback if needed.

## Scope and non-goals

- **Telegram-specific behavior** is not changed in this PR; the fix is at the bus core layer, so Discord/Slack/webui all benefit identically once the synthetic `ingestReply` flows through their normal `response.text` subscribers.
- **Receipt-chain integration** (`#211`) is intentionally out of scope here. Once `#211` lands, this PR's `handleTurnEnd` is the natural place to stamp `auto_recovered: true` on the receipt — happy to do that as a small follow-up after `#211`.
- **Option A (prompt enforcement)** from #215 is also out of scope. The two layers are complementary; B is the more robust of the two and unblocks the immediate UX-fatal pattern.

## Version bump

Plugin + marketplace bumped to `2.2.139` per repo policy (`src/` touched).

## Migration notes

None — additive. Existing behavior is preserved when the agent calls `reply` correctly. The new safety net only fires when the agent ends a turn with non-empty text without delivering — the exact pattern that was previously silent.

### DIFF
--- a/src/bus/core.ts
+++ b/src/bus/core.ts
@@ -195,9 +195,27 @@ export class BusCoreImpl implements BusCore {
    * in-flight prompts on one agent would let the second `sendPrompt` reset
    * the flag mid-turn for the first; the webui-bridge mutex serialising
    * prompt→reply per agent is the workaround for the path that would
-   * otherwise violate the one-turn-at-a-time assumption.
+   * otherwise violate the one-turn-at-a-time assumption. The residual
+   * cross-channel stale/misroute race this leaves open (a lagged turn_end
+   * synthesizing the wrong prompt's text to the wrong surface) is tracked
+   * as deferred follow-up #239 — see the note in `handleTurnEnd`.
    */
   private readonly currentTurnReplied = new Map<string, boolean>();
+
+  /**
+   * Cross-transport dedup for the silent-drop net (#217, finding 2).
+   * `currentTurnReplied` is reset to `false` on every `sendPrompt` and set
+   * to `true` once a final lands — but a real `reply` (IPC) and the
+   * synthesized recovery (`response.turn_end` via the fs.watch tailer)
+   * travel on two independent, unordered async channels. If the tailer
+   * wins the race, `handleTurnEnd` synthesizes+delivers a final, then the
+   * real `ingestReply(final)` IPC arrives and would deliver the SAME text
+   * a second time. This per-turn flag records that a final `response.text`
+   * was already PUBLISHED for the in-flight turn so the loser of the race
+   * is suppressed. Set in both `ingestReply(final)` and the synthesizer;
+   * cleared alongside `currentTurnReplied` on prompt / disconnect / cancel.
+   */
+  private readonly currentTurnFinalPublished = new Map<string, boolean>();
 
   constructor(opts: BusCoreOptions = {}) {
     this.socketPath = opts.socketPath ?? null;
@@ -228,6 +246,7 @@ export class BusCoreImpl implements BusCore {
           // origin and misroute (5-agent review on PR #138, A1 finding).
           this.lastPromptOrigin.delete(agentId);
           this.currentTurnReplied.delete(agentId);
+          this.currentTurnFinalPublished.delete(agentId);
         }
       },
       onError: (err, agentId) => this.onError(err, { ctx: "ipc", agentId }),
@@ -289,6 +308,9 @@ export class BusCoreImpl implements BusCore {
     // turn call reply?" flag. If the agent ends the turn (response.turn_end)
     // without ever setting this to true, we'll synthesize delivery.
     this.currentTurnReplied.set(req.agent_id, false);
+    // #217 finding 2: a new turn starts undelivered — clear the
+    // cross-transport "final already published" dedup flag.
+    this.currentTurnFinalPublished.set(req.agent_id, false);
 
     const ipcMsg: IpcPrompt = {
       type: "prompt",
@@ -382,7 +404,22 @@ export class BusCoreImpl implements BusCore {
 
   /* ─────────────────────────────── ingest ─────────────────────────────── */
 
-  ingestReply(req: IngestReplyRequest): void {
+  ingestReply(req: IngestReplyRequest, opts?: { synthetic?: boolean }): void {
+    // Cross-transport dedup (#217 finding 2): a final reply can arrive both
+    // as the agent's real `reply` IPC AND as the synthesized recovery from
+    // `response.turn_end` (the JSONL tailer). They race on two unordered
+    // channels; whichever lands first publishes and sets
+    // `currentTurnFinalPublished`. The loser is suppressed here so the same
+    // turn never delivers two finals. `synthetic` calls (from
+    // `handleTurnEnd`) bypass the check — they only run AFTER confirming no
+    // final has published yet, and must be allowed to publish the first.
+    if (
+      req.intent === "final" &&
+      !opts?.synthetic &&
+      this.currentTurnFinalPublished.get(req.agent_id) === true
+    ) {
+      return;
+    }
     const topic: BusEventTopic =
       req.intent === "tool_status" ? "response.tool_use" : "response.text";
     // Attach the originating surface so adapters can route the reply
@@ -431,6 +468,10 @@ export class BusCoreImpl implements BusCore {
       // intent → mark this turn as delivered, so the `response.turn_end`
       // handler skips the synthetic-delivery fallback for this turn.
       this.currentTurnReplied.set(req.agent_id, true);
+      // #217 finding 2: record that a final was published for this turn so
+      // the cross-transport race loser (real reply vs synthesized) is
+      // suppressed above.
+      this.currentTurnFinalPublished.set(req.agent_id, true);
     }
   }
 
@@ -446,7 +487,23 @@ export class BusCoreImpl implements BusCore {
    */
   private handleTurnEnd(agentId: string, text: string): void {
     if (this.currentTurnReplied.get(agentId) === true) return;
+    // #217 finding 2: if a final already published for this turn (e.g. the
+    // real reply IPC landed first), don't synthesize a duplicate.
+    if (this.currentTurnFinalPublished.get(agentId) === true) return;
     if (!text || text.trim().length === 0) return;
+    // RESIDUAL RACE (#217 finding 3 → deferred #239): `lastPromptOrigin` and
+    // the per-turn flags are single-slot, keyed by agent_id only, with no
+    // prompt/turn identity. The JSONL tailer's `response.turn_end` is
+    // delivered asynchronously (fs.watch + drain), so on a cross-channel
+    // interleave — P1(telegram) ends → reply → P2(webui) arrives (resets the
+    // flag, rewrites origin=webui) → P1's lagged turn_end lands here — this
+    // can route P1's text to P2's origin (webui). The flag-reset path is
+    // largely covered by the webui-bridge mutex serialising prompt→reply per
+    // agent, but cross-surface interleave is NOT fully closed. A complete fix
+    // threads the originating prompt_id onto the turn_end event and matches
+    // it to the in-flight prompt before synthesizing; that requires the
+    // tailer to carry prompt identity and is tracked as deferred follow-up
+    // #239 to avoid destabilizing this safety-net PR.
     const origin = this.lastPromptOrigin.get(agentId);
     if (!origin) {
       // No origin to route to (cron tick / unprompted ambient turn).
@@ -477,11 +534,17 @@ export class BusCoreImpl implements BusCore {
     // because we set the flag here, but the explicit ordering makes the
     // single-fire intent clearer).
     this.currentTurnReplied.set(agentId, true);
-    this.ingestReply({
-      agent_id: agentId,
-      text,
-      intent: "final",
-    });
+    // `synthetic: true` lets this first delivery through the cross-transport
+    // dedup (#217 finding 2); it sets `currentTurnFinalPublished`, so a
+    // later real `reply` IPC for the same turn is suppressed.
+    this.ingestReply(
+      {
+        agent_id: agentId,
+        text,
+        intent: "final",
+      },
+      { synthetic: true },
+    );
   }
 
   ingestSessionEvent(e: BusEvent): void {
@@ -641,6 +704,7 @@ export class BusCoreImpl implements BusCore {
         // a real reply OR a `response.turn_end` event — drop the
         // tracking flag to keep the map bounded.
         this.currentTurnReplied.delete(agentId);
+        this.currentTurnFinalPublished.delete(agentId);
         break;
       case "request_human": {
         // Forward the correlation id along with the question. Without
@@ -698,6 +762,7 @@ export class BusCoreImpl implements BusCore {
         // Silent-drop safety net (#215): error path won't produce a
         // turn_end either — clear tracking too.
         this.currentTurnReplied.delete(agentId);
+        this.currentTurnFinalPublished.delete(agentId);
         break;
       // hello already handled in the IPC layer; outbound types (prompt,
       // permission_response, ask_answer) shouldn't arrive from MCP.
--- a/src/bus/core.ts
+++ b/src/bus/core.ts
@@ -178,6 +178,18 @@ export class BusCoreImpl implements BusCore {
    */
   private readonly lastPromptOrigin = new Map<string, { origin: BusOrigin; origin_id: string }>();
 
+  /**
+   * Silent-drop safety net (issue #215): tracks per-agent whether the
+   * current turn has called the `reply` MCP tool with `intent: "final"`.
+   * Reset on every new inbound prompt (`sendPrompt`), set to `true` when
+   * an `intent: "final"` `ingestReply` lands. On `response.turn_end`
+   * (emitted by the JSONL tailer when `stop_reason === "end_turn"`), if
+   * this is still `false` and the turn produced non-empty text, the
+   * agent ended the turn without delivering — we synthesize an
+   * `ingestReply` so the user actually receives the response.
+   */
+  private readonly currentTurnReplied = new Map<string, boolean>();
+
   constructor(opts: BusCoreOptions = {}) {
     this.socketPath = opts.socketPath ?? null;
     this.ringbufferCapacity = opts.ringbufferCapacity ?? DEFAULT_RINGBUFFER_CAPACITY;
@@ -206,6 +218,7 @@ export class BusCoreImpl implements BusCore {
           // agent (after a reconnect) would inherit the dead session's
           // origin and misroute (5-agent review on PR #138, A1 finding).
           this.lastPromptOrigin.delete(agentId);
+          this.currentTurnReplied.delete(agentId);
         }
       },
       onError: (err, agentId) => this.onError(err, { ctx: "ipc", agentId }),
@@ -263,6 +276,10 @@ export class BusCoreImpl implements BusCore {
       origin: req.origin,
       origin_id: req.origin_id,
     });
+    // Silent-drop safety net (#215): new prompt → reset the "did this
+    // turn call reply?" flag. If the agent ends the turn (response.turn_end)
+    // without ever setting this to true, we'll synthesize delivery.
+    this.currentTurnReplied.set(req.agent_id, false);
 
     const ipcMsg: IpcPrompt = {
       type: "prompt",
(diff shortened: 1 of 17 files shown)


Nibbler1250:
Heads-up before this merges — I traced why the safety net wasn't recovering dropped replies in a live deployment, and the root cause sits upstream of this PR's logic.

**The synthesize handler can't fire in production today.** `handleTurnEnd` is driven by `response.turn_end` events arriving through `ingestSessionEvent`, and the only thing that emits those is the `JsonlTailer`. But the `JsonlTailer` is never instantiated in the runtime — only in tests. So no `response.turn_end` events are ever produced outside the test suite, and this safety net is inert on every deployment. The existing tests pass because they inject `turn_end` directly, bypassing the tailer.

I have a fix that completes this PR by wiring the tailer in. Implementing it surfaced two adjacent gaps that have to be handled for the wiring to actually work:

1. **Replay-on-resume would re-synthesize stale replies.** `start()` replays from byte 0. On a resumed session (same `--session-id`) that re-emits every historical `turn_end`; with a prompt in flight it would synthesize a stale reply. I added a `startAt: "end"` mode so the runtime tailer live-tails new turns only (default stays `"begin"` for existing consumers and tests).

2. **The tailer gives up if the session file doesn't exist yet at `start()`.** That's the common fresh-session case — `claude` writes the JSONL lazily on the first turn. Today `watch()` throws ENOENT and the code only logs it (the "schedule a one-shot retry" comment was never implemented), so the tailer never attaches and no events ever flow. I added a poll-for-appearance + attach.

Wiring summary: `SessionManager` spawns a per-agent `JsonlTailer` (when a bus is configured) bound to the agent's live session, created after any session-id rotation so it binds the final id, torn down on `stop()`/exit; `runtime-mount` passes the bus. A `projectsDir` option keeps it hermetically testable.

E2E test added through `SessionManager`: a real `end_turn`-with-text-but-no-`reply` JSONL line now produces a delivered synthesized reply — exercising the whole chain instead of a direct event inject. Full suite 87/0, typecheck delta 0.

I'll push this onto the branch shortly (with the manifest version bumps) unless you'd prefer it as a separate stacked PR — your call on the shape.

TerrysPOV (dismissed):
### Code review

Re-review of `5caba8b` (maintainer-rebased onto current main, version → 2.2.136). All prior concerns resolved — A1 lifecycle cleanup paired at all four sites, tailer binds the FINAL post-rotation session_id ([`session-manager.ts:660-682`](https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5caba8b77d58e2f06fc02f4eb0612a5b7a89e16c/src/bus/session-manager.ts#L656-L683)), wiring fix from `1532b100` correctly addresses your own inert-handler flag. 330 pass / 0 fail on the rebase.

Found 1 issue:

1. **`startAt: "end"` startup TOCTOU — missing follow-up drain on the existing-file path.** The `awaitFileCreation` (not-yet-created) path correctly closes the file-appearance race by calling `scheduleDrain()` after `attachFileWatcher()` ([`jsonl-tailer.ts:230-232`](https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5caba8b77d58e2f06fc02f4eb0612a5b7a89e16c/src/bus/jsonl-tailer.ts#L228-L234)). The existing-file branch with `startAt: "end"` doesn't — the sequence is `statSync(filePath).size → emitReplayDone → attachFileWatcher → return`, with no follow-up drain. Bytes written between the `statSync` and `attachFileWatcher` aren't picked up by the watcher (it only fires on changes after attach) and stay unread until the **next** write triggers `scheduleDrain`. If the missed bytes contain an `end_turn` line and the agent then goes idle (the exact pattern this safety net exists to recover), synthesis is delayed indefinitely instead of firing. Window is narrow (microseconds), but the failure mode is precisely the silent-drop case #215 was filed for.

   https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5caba8b77d58e2f06fc02f4eb0612a5b7a89e16c/src/bus/jsonl-tailer.ts#L179-L196

   One-line fix: add `this.scheduleDrain()` after `this.attachFileWatcher()` in the existing-file branch — mirrors what `awaitFileCreation` already does, and the existing test in `startAt: "end"` describe block would catch any regression in the symmetry.

Two low-priority nits worth a look while you're in there (not blockers):

- `currentTurnReplied` JSDoc at [`core.ts:181-191`](https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5caba8b77d58e2f06fc02f4eb0612a5b7a89e16c/src/bus/core.ts#L178-L192) doesn't mirror the explicit single-slot last-write-wins caveat that `lastPromptOrigin` carries at lines 170-179. It inherits exactly the same limitation (and depends on `lastPromptOrigin` for routing), so worth a sentence pointing to the webui-bridge mutex as the workaround for the path that violates it.
- No direct unit test asserting `currentTurnReplied.delete` runs on cancel/error/disconnect. Currently relies on observable downstream behaviour via `lastPromptOrigin`. One test per site would close the gap.

Once the TOCTOU drain is wired, ready to merge.

Nibbler1250:
Thanks for the re-review — TOCTOU drain wired in `6361330`.

**Blocker — startup TOCTOU (fixed).** Added `this.scheduleDrain()` right after `attachFileWatcher()` on the existing-file `startAt: "end"` branch, mirroring exactly what `awaitFileCreation` already does. Bytes written in the seek-to-EOF → attach window are now swept by the follow-up drain instead of waiting on the next write. The `startAt: "end"` describe block stays green (no symmetry regression), full bus suite 330 pass / 0 fail, biome clean.

**Nit 1 — JSDoc (done).** Documented `currentTurnReplied` as single-slot last-write-wins, noting it depends on `lastPromptOrigin` for routing and inherits the same caveat, with a pointer to the webui-bridge mutex as the serialisation workaround.

**Nit 2 — `currentTurnReplied.delete` test (held, on purpose).** I looked at wiring a per-site test and it can't be made into an honest guard with the current design: `currentTurnReplied` is deleted in the *same* `onClose` handler that clears `lastPromptOrigin`, and is reset by every `sendPrompt`. There's no path that clears it in isolation, so a behavioural test would pass even with the `.delete` removed (the co-deleted `lastPromptOrigin`, already covered by the A1 disconnect test, would suppress synthesis on its own) — a green test that doesn't actually protect the line. I'd rather not ship that. If you want explicit coverage I'm happy to add a tiny test-only state accessor and assert the map entry is gone post-disconnect — say the word.

**Versioning.** Branch was 1 behind `main` with the only conflict in the two version manifests (`main` moved to 2.2.137 via #229). Bumped 2.2.136 → **2.2.138** to stay above `main`. The manifest conflict is trivial — rebase at your convenience or I can take it, whichever you prefer.

Nibbler1250 (commented):
Self-review of the safety net. **Two confirmed correctness bugs are already fixed in 5f9e956** (pushed to this branch):

1. The production daemon (`start.ts`) built its own `SessionManager` without `bus`, bypassing `mountBusRuntime`'s bus-wired default — the whole net was inert in prod. Now passes `bus`.
2. `handleTurnEnd` fired for `cron`/`heartbeat` origins (which set `lastPromptOrigin` but aren't channel-driven), spamming a bogus "recovered" warning + an undeliverable `response.text` on every scheduled tick. Now gated on `CHANNEL_DRIVEN_ORIGINS`.

Remaining notes inline below — none blocking, listed worst-first.

src/bus/jsonl-tailer.ts
```diff
@@ -164,31 +177,85 @@ export class JsonlTailer {
...
     // path picks up the first byte when it appears.
     if (existsSync(this.filePath)) {
-      await this.drainFromOffset();
+      if (this.startAt === "end") {
+        // Issue #215 runtime wiring: live-tail NEW turns only. Replaying
+        // from byte 0 on a resumed session would re-emit historical
+        // `response.turn_end` events and, with a prompt in flight, could
+        // synthesize a stale reply. Seek to EOF and watch forward.
+        try {
+          this.offset = statSync(this.filePath).size;
```
Nibbler1250, line 195: statSync throw here is swallowed and `this.offset` stays 0. Because `existsSync` already passed, the only way in is a transient failure (perm flip, NFS hiccup, file rotated under us) — but then `scheduleDrain()` below runs from offset 0 and replays the **entire** history, re-emitting historical `response.turn_end` and (with a prompt in flight) synthesizing a stale reply. That is exactly what `startAt: "end"` exists to prevent. Suggest: on stat failure, skip the drain / bail to the await-creation path rather than continuing with offset 0.

src/bus/core.ts
```diff
@@ -178,6 +179,26 @@ export class BusCoreImpl implements BusCore {
...
+   *
+   * Single-slot per agent, last-write-wins — the same limitation
+   * `lastPromptOrigin` carries above (and this map depends on
+   * `lastPromptOrigin` for routing the synthesized reply). Interleaving two
+   * in-flight prompts on one agent would let the second `sendPrompt` reset
+   * the flag mid-turn for the first; the webui-bridge mutex serialising
+   * prompt→reply per agent is the workaround for the path that would
+   * otherwise violate the one-turn-at-a-time assumption.
+   */
+  private readonly currentTurnReplied = new Map<string, boolean>();
```
Nibbler1250, line 203: Single-slot `currentTurnReplied` + `lastPromptOrigin`. The doc rightly flags interleaving, but the mitigation (webui-bridge mutex) only covers the webui path. In the real deployment one default agent serves telegram + webui-inject + cron on this one slot: a telegram prompt arriving mid-cron-turn resets the slot, then the cron turn_end synthesizes a reply routed to telegram — the user gets unsolicited ambient text. The `CHANNEL_DRIVEN_ORIGINS` gate in 5f9e956 narrows but does not close this (telegram-during-webui still races). Worth a per-agent prompt→turn_end serialization, or keying turn state by origin_id.

src/bus/core.ts
```diff
@@ -402,9 +428,74 @@ export class BusCoreImpl implements BusCore {
...
+   * Silent-drop safety net handler (issue #215). Wired by
+   * `ingestSessionEvent`/JSONL tailer when it observes a `response.turn_end`
+   * with `stop_reason: "end_turn"`. If the agent ended the turn with
+   * non-empty text but never called the `reply` tool for this prompt,
+   * synthesize an `ingestReply` so the user actually receives the
+   * response. Without this, the text sits in the session `.jsonl` and
+   * the user-facing surface gets nothing — confirmed live 2x in 12h
+   * on a real bus-mode deployment after issue #215 was filed.
+   */
+  private handleTurnEnd(agentId: string, text: string): void {
```
Nibbler1250, line 488: Altitude: this delivers the raw concatenated `text` blocks of the whole turn as the user-facing reply. `thinking` is excluded, but any working/planning prose the agent wrote as `text` (intending it as scratch, expecting to summarize via `reply`) now ships verbatim. The `reply` tool exists precisely so the agent curates what the user sees. Consider treating a missing-reply as a bounded signal (operator alert / truncated notice) rather than a full text dump, or at least tagging the synthesized event so surfaces can choose to suppress it.

src/bus/core.ts
```diff
@@ -402,9 +428,74 @@ export class BusCoreImpl implements BusCore {
     if (req.intent === "final") {
       this.lastPromptOrigin.delete(req.agent_id);
     }
+    // Silent-drop safety net (#215): the agent called `reply` with a
+    // final intent → mark this turn as delivered. The
+    // `response.turn_end` handler will skip the synthetic-delivery
+    // fallback for this turn.
+    if (req.intent === "final") {
```
Nibbler1250: Minor: these two consecutive `if (req.intent === "final")` blocks can merge into one — both branch on the same condition (delete `lastPromptOrigin`, set `currentTurnReplied`).

Nibbler1250:
5-agent review run (per Terry's delegated process while away). It caught two real correctness bugs that this PR's substance review and my self-review both missed — both fixed in `5f9e956`:

1. The production daemon (`start.ts`) built its `SessionManager` without `bus`, bypassing `mountBusRuntime`'s bus-wired default — the silent-drop net was **inert in prod** (no tailer, no `turn_end`). Now passes `bus`.
2. `handleTurnEnd` fired for `cron`/`heartbeat` origins → bogus "recovered" warnings + an undeliverable `response.text` every scheduled tick. Gated on `CHANNEL_DRIVEN_ORIGINS` + regression test.

Remaining review notes addressed in `53ef2f3`:

- **Note 1** (jsonl-tailer): a transient `statSync` failure on the EOF seek left `offset` at 0, so the follow-up drain replayed the whole history and could synthesize a stale reply — the exact thing `startAt:"end"` prevents. Now deferred via a `seekToEndPending` flag so the next drain re-seeks to current EOF and tails forward.
- **Note 4** (core): merged the two consecutive `if (req.intent === "final")` blocks.

Non-blocking, deeper notes filed as follow-ups: per-origin turn-state serialization → #239, synthesized-reply altitude → #240.

CI green (4/4). T1 (full bus suite) / T2 (tsc + biome) / self-review / security-review all clean, no regressions.

Nibbler1250:

A multi-agent review (per-dimension finders → adversarial verifiers) ran over this PR: **3 confirmed**, 3 dismissed as false-positive/intended.
All confirmed findings are **addressed in 11b8713**.

### 🔴 HIGH — Wiring the tailer into the live bus double-delivers every assistant text block (raw response.text echo, in addition to the agent's reply / the synthesized final)
*src/bus/jsonl-tailer.ts:462 (dispatchAssistant) + src/bus/session-manager.ts:653-685 + src/commands/start.ts:506 + src/bus/runtime-mount.ts:238*

This PR is the first to wire JsonlTailer into the PRODUCTION bus that adapters subscribe to. On `main`, `new JsonlTailer` appears only in tests (verified: `git grep 'new JsonlTailer' origin/main` → 3 test files only).

**Fix applied:** The fan-out path for raw per-block `response.text` from the tailer must not reach delivery adapters. Options, in rough order of cleanliness:

(a) Tag tailer-origin events so adapters can distinguish "observability echo" from "deliverable reply".

### ⚪ LOW — Duplicate final reply when the JSONL tailer wins the race against the real reply IPC
*src/bus/core.ts:447-483 (handleTurnEnd) + ingestReply IPC path src/bus/core.ts:598-599*

handleTurnEnd guards only on currentTurnReplied. The real `reply` tool and the synthetic recovery arrive over two independent, unordered async channels: the agent's correct `reply` lands via IPC (`case "reply"` -> ingestReply, sets currentTurnReplied=true) while `response.turn_end` lands via the fs.watch+drain JSONL tailer. Nothing guarantees the IPC ingestReply lands before the tailer processes the end_turn line.

**Fix applied:** Add a content/recency dedup so a real and synthetic final reply can't both deliver regardless of ordering. Cheapest: record per-turn whether a final response.text was already published (set it in BOTH ingestReply(final) and handleTurnEnd's synthesis), and in ingestReply suppress (or coalesce) a final-intent publish if synthesis already fired for the same turn; symmetrically handleTurnEnd already skips when the flag is set.

### 🟠 MEDIUM — Cross-channel interleave can synthesize a STALE prompt's text and misroute it to a different channel
*src/bus/core.ts:447-483 (handleTurnEnd), currentTurnReplied/lastPromptOrigin single-slot maps (core.ts:200, 291)*

currentTurnReplied and lastPromptOrigin are single-slot, keyed by agent_id only, with no turn/prompt identity. sendPrompt has no cross-channel per-agent serialization in the bus core (the only mutex is webui-bridge-local, so it does not serialize a telegram turn against a webui turn on the same agent). Sequence: P1(telegram) ends, agent calls reply (flag=true, telegram origin deleted); P2(webui) arrives (flag reset to false, origin=webui); the tailer — which is fire-and-forget/async — only now publishes P1's response.turn_end. handleTurnEnd sees flag=false (reset by P2) and origin=webui (channel-driven) and synthesizes P1's text routed to the webui user.

**Fix applied:** Give the safety net per-turn/per-prompt identity instead of agent-only single-slot state. Concretely: stamp each sendPrompt with the promise_id (already generated) and propagate it onto the tailer's response.turn_end event (via the session/turn boundary), then have handleTurnEnd only synthesize when the turn_end's prompt-id matches the CURRENT in-flight prompt id (and the flag for THAT id is unreplied).

PR #71, opened 2026-05-15
feat: MCP multiplexer — shared MCP processes across PTYs (#64)

TerrysPOV:
## Summary

Implements the MCP multiplexer (issue #64). Replaces per-PTY MCP-server spawning (216 processes at 36 PTYs × 5 MCPs — OOM-bound on small-to-medium daemon hosts) with a daemon-hosted multiplexer that runs each shared MCP server **once** and serves it over loopback HTTP to all PTY-resident `claude` instances. Target footprint: `1 (bridge) + 5 (shared MCPs) + 36 (claudes) = 42 processes`, ~5× reduction.

## What this unblocks

Per the `> [!WARNING]` callout in README (PR #63), production `pty.enabled: true` was blocked on this milestone landing. After this merges + operator validates RAM under realistic load + OAuth refresh sanity-check passes, the flag flips and the warning is removed.

## Architecture

Frozen by SPEC in `.planning/mcp-multiplexer/SPEC.md` (518 LOC). High-level:

- **Multiplexer plugin** (`src/plugins/mcp-multiplexer/`) hosts shared MCP server children once at daemon startup. Exposes each on a per-server HTTP route via `PluginHttpGateway.registerMcpHandler`. Streamable HTTP transport (`@modelcontextprotocol/sdk` SDK Server + `WebStandardStreamableHTTPServerTransport`) per `(serverName, ptyId)` bucket — stateless servers collapse to one bucket, stateful keep per-PTY isolation.
- **PTY supervisor** synthesises `${cwd}/.claudeclaw/mcp-pty-${ptyId}.json` (mode `0600`) at spawn with `--mcp-config` pointing at `http://127.0.0.1:<gateway-port>/mcp/<server-name>` and per-PTY HMAC bearer headers. The bridge URL is read from the multiplexer plugin (single source of truth — no recomputation in the supervisor).
- **Auth:** `Authorization: Bearer <hex-secret>` + `X-Claudeclaw-Pty-Id` + `X-Claudeclaw-Ts`. Per-PTY secrets are 32-byte random, minted on every PTY spawn/respawn (rotation cadence bounded by `pty.idleReapMinutes`, default 30). Bearer-as-secret is the pragmatic equivalent of SPEC §4.3's per-request HMAC, since Claude Code's stock MCP HTTP client can't sign request bodies — documented in `.planning/mcp-multiplexer/W1-COORD.md` §1 and validated in `SECURITY-AUDIT.md`.
[...]

PR #218, opened 2026-06-02
fix(bus): agent-scoped session rotation on bus path (closes #213)

### LINKED ISSUE #213: streamBusPrompt skips incrementMessageCount — session rotation never fires on bus path

Nibbler1250:
## Problem

Closes #213. Implements the agent-scoped design Terry outlined in the [10:40 UTC #213 review](https://github.com/TerrysPOV/ClaudeClaw-Plus/issues/213): issue-body **B** + comment-**A** rolled into one focused PR.

The bus path never called `incrementMessageCount()` or `needsRotation()` after migration, leaving session rotation silently dead. The naive fix — `await incrementMessageCount()` with no agent argument — would have made things worse by rotating the **global** session (the brief bootstrap "wakeup" turn) instead of the long-lived agent PTY where every Telegram / `/api/inject` / `/api/chat` turn actually accumulates. Quoting #213:

> Flipping `needsRotation()` to true would then rotate the bootstrap session, write a summary of the wakeup turn, and leave the bloated agent session.jsonl growing untouched. That's a worse failure mode than the current silence.

This PR threads `agentId` through the bookkeeping so rotation operates on the agent's own `session.json` — the one that actually grows — without touching the global tracker the legacy `runner.ts` code path still owns.

### `src/bus/webui-bridge.ts streamBusPrompt`

After `runPrompt` resolves `ok: true`:

```ts
await incrementMessageCount(agentId);
const sessionConfig = getSettings().session;
if (sessionConfig.autoRotate) {
  const peeked = await peekSession(agentId);
  if (peeked && needsRotation(peeked, sessionConfig)) {
    await rotateSession(sessionConfig, agentId);
  }
}
```

Wrapped in try/catch — a bookkeeping failure must never break the prompt flow the caller is awaiting. `getSettings()` is also wrapped because it throws when called before `loadSettings()` (early boot, unit tests); the increment still fires, rotation skipped until settings are loadable.

### `src/rotation.ts rotateSession`

New optional second parameter `agentName`. When provided:

- `peekSession(agentName)` — reads the agent's `session.json`.
- Summary written to `agents/<agentName>/summaries/` (NOT the global `summaries/` dir — per-agent histories stay isolated).
- `backupSession(agentName)` + `resetSession(agentName)` — touch only the agent's files.

Omitted = legacy global behaviour, unchanged. The legacy `runner.ts` callsites at `:1583` and `:2386` still call `rotateSession(config)` with no agent → no regression.

### `src/runner.ts:2152` comment

The "Count this invocation for rotation tracking (global session only; **agent sessions don't rotate**)" comment was an artifact of the pre-bus world where the user *was* talking to the global session. Updated to point to the bus-side rotation hook in `streamBusPrompt` and clarify that legacy and bus paths now track separate sessions by design.

## Why this and not the alternatives from #213

Per maintainer guidance:

| Option | Verdict |
|---|---|
| Issue-body A (1-line global `incrementMessageCount`) | **Rejected** — rotates the wrong session, worse than the bug |
| Issue-body B (agent-aware bookkeeping + needsRotation at bus seam) | **Adopted** — this PR |
| Comment-A (new `rotateSession(config, agentName)` overload) | **Adopted** — this PR |
| Comment-B (sync `global.sessionId := agents/default/...`) | **Rejected** — encodes "default is special", foot-gun for #71 multiplexer |
| Comment-C (eliminate global tracker entirely) | **Deferred** — right end state, too much for this PR (touches legacy `runner.ts` + several no-arg `peekSession` callsites). File as follow-up milestone. |

## Test plan

- [x] **T1 build**: `bun build src/index.ts --outdir /tmp/claw-smoke --target bun --no-splitting` → clean
- [x] **T1 lint**: `bun run lint` → exit 0 (no new biome errors introduced; one pre-existing tsc error in `src/skills-tuner/core/types.ts:19` unchanged and unrelated)
- [x] **T2 bus + sessions tests**: 19 pass, 0 fail (no regression)
- [x] **Live-validated on a real bus-mode daemon**:
  - **Pre-deploy**: global `messageCount = 6` (orphan, doesn't reflect actual usage); agent-default `session.json` had no `messageCount` field.
  - **Post-deploy + restart + `/api/inject "pong"`**:
    - Global: `messageCount = 6` (unchanged — bus path no longer touches it ✓)
    - Agent default: `messageCount = 1` (new field, incremented correctly ✓)
  - Daemon answered the probe in 10s with `ok: true`, `exitCode: 0`.
  - Backups kept for fast rollback if needed.

## Ordering constraint (from #213)

Per maintainer:

1. **#216** — review + merge first (clean orthogonal precondition; preserves `createdAt`).
2. **THIS PR** — agent-scoped rotation.
3. **#217** — silent-drop safety net; lands cleanly on top because the synthesized `ingestReply` will now route through `lastPromptOrigin` correctly and the receipt's `session_id` will match what the tailer actually reads.
4. **Follow-up milestone**: comment-C (eliminate the global tracker entirely).

## Supersedes #214

PR #214 was closed (implemented the rejected issue-body A approach). This PR replaces it with the maintainer-approved design.

## Version bump

Plugin + marketplace bumped to `2.2.133` per repo policy (`src/` touched).

## Migration notes

None — additive. Legacy `runner.ts` behaviour unchanged. Existing daemons gain a per-agent `messageCount` field the first time an `/api/inject`, `/api/chat`, or `/api/jobs/fire` turn completes successfully; rotation activates as soon as `maxMessages` or `maxAgeHours` thresholds are reached on the agent's session.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.134",
+      "version": "2.2.135",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.134",
+  "version": "2.2.135",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/webui-bridge.test.ts
+++ b/src/bus/__tests__/webui-bridge.test.ts
@@ -9,8 +9,8 @@
  * reply path by calling `bus.ingestReply` directly — that's how the
  * plus-bus MCP server normally publishes claude's responses.
  */
-import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from "bun:test";
-import { mkdtempSync, readFileSync, rmSync, existsSync } from "node:fs";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, spyOn } from "bun:test";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync, existsSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { createBusCore, type BusCore } from "../core";
@@ -23,6 +23,8 @@ import {
   type ReceiptStore,
 } from "../receipt";
 import { streamBusPrompt } from "../webui-bridge";
+import { createSession, peekSession, incrementMessageCount } from "../../sessions";
+import { reloadSettings } from "../../config";
 
 // Redirect the process-wide default receipt store to a throwaway path for the
 // duration of this file so the existing 11 streamBusPrompt tests (which
@@ -348,5 +350,73 @@ describe("streamBusPrompt — receipt chain", () => {
     expect(recs[0].process_generation).toBe(1);
     // After close, the hash index is cleared.
     expect(store.findByPromptHash(hashPrompt("lookup me"))).toBeUndefined();
+  });
+});
+
+describe("streamBusPrompt — per-agent bookkeeping (#213, split per #218)", () => {
+  // getAgentsDir() resolves from the live process.cwd(), so chdir'ing into a
+  // temp dir isolates each test's agent session files (removed with the temp
+  // dir). config's SETTINGS_FILE is frozen at import-time cwd, so the
+  // rotation-gate test writes there explicitly and cleans up.
+  let frozenCwd: string;
+  let tmp: string;
+
+  async function driveOneTurn(agentId: string): Promise<void> {
+    const bus = makeBus();
+    const pending = streamBusPrompt(bus, agentId, "hi", { timeoutMs: 2000 });
+    await Promise.resolve();
+    await Promise.resolve();
+    bus.ingestReply({ agent_id: agentId, text: "ok", intent: "final" });
+    // Bookkeeping is awaited inside streamBusPrompt before it resolves.
+    expect((await pending).ok).toBe(true);
+  }
+
+  beforeEach(() => {
+    frozenCwd = process.cwd();
+    tmp = mkdtempSync(join(tmpdir(), "ccaw-218-"));
+    process.chdir(tmp);
+  });
+
+  afterEach(() => {
+    process.chdir(frozenCwd);
+    rmSync(tmp, { recursive: true, force: true });
+  });
+
+  it("increments the named agent's message count on a successful prompt", async () => {
+    await createSession("11111111-1111-1111-1111-111111111111", "alpha");
+    expect((await peekSession("alpha"))?.messageCount ?? 0).toBe(0);
+    await driveOneTurn("alpha");
+    expect((await peekSession("alpha"))?.messageCount).toBe(1);
+  });
+
+  it("WARNs and defers to #227 when the agent crosses the threshold — without rotating", async () => {
+    const settingsFile = join(frozenCwd, ".claude", "claudeclaw", "settings.json");
+    mkdirSync(join(frozenCwd, ".claude", "claudeclaw"), { recursive: true });
+    writeFileSync(settingsFile, JSON.stringify({ session: { autoRotate: true, maxMessages: 2 } }));
+    try {
+      await reloadSettings();
+      await createSession("22222222-2222-2222-2222-222222222222", "beta");
+      await incrementMessageCount("beta"); // 0 -> 1; the next turn trips the gate
+      const warnSpy = spyOn(console, "warn").mockImplementation(() => {});
+
+      await driveOneTurn("beta"); // 1 -> 2, needsRotation(2 >= 2) === true
+
+      const warned = warnSpy.mock.calls.some((c) => String(c[0]).includes("deferred to #227"));
+      warnSpy.mockRestore();
+      expect(warned).toBe(true);
+
+      // The rotation MECHANISM is not wired: the session is counted, not reset.
+      const after = await peekSession("beta");
+      expect(after?.messageCount).toBe(2);
+      expect(after?.sessionId).toBe("22222222-2222-2222-2222-222222222222");
+    } finally {
+      rmSync(settingsFile, { force: true });
+    }
+  });
+
+  it("does not break the prompt when there is no session to count (best-effort)", async () => {
+    // No createSession — incrementMessageCount no-ops; the prompt still resolves.
+    await driveOneTurn("gamma");
+    expect(await peekSession("gamma")).toBeNull();
   });
 });
--- a/src/bus/__tests__/webui-bridge.test.ts
+++ b/src/bus/__tests__/webui-bridge.test.ts
@@ -24,7 +24,7 @@ import {
 } from "../receipt";
 import { streamBusPrompt } from "../webui-bridge";
 import { createSession, peekSession, incrementMessageCount } from "../../sessions";
-import { reloadSettings } from "../../config";
+import { reloadSettings, _setSettingsFileForTests } from "../../config";
 
 // Redirect the process-wide default receipt store to a throwaway path for the
 // duration of this file so the existing 11 streamBusPrompt tests (which
@@ -357,7 +357,8 @@ describe("streamBusPrompt — per-agent bookkeeping (#213, split per #218)", () 
   // getAgentsDir() resolves from the live process.cwd(), so chdir'ing into a
   // temp dir isolates each test's agent session files (removed with the temp
   // dir). config's SETTINGS_FILE is frozen at import-time cwd, so the
-  // rotation-gate test writes there explicitly and cleans up.
+  // rotation-gate test redirects it into the temp dir via
+  // _setSettingsFileForTests() instead of touching the real settings.json.
   let frozenCwd: string;
   let tmp: string;
 
@@ -378,6 +379,9 @@ describe("streamBusPrompt — per-agent bookkeeping (#213, split per #218)", () 
   });
 
   afterEach(() => {
+    // Restore the default settings path and clear the cache so neither the
+    // override nor cached settings leak into later tests in this file.
+    _setSettingsFileForTests();
     process.chdir(frozenCwd);
     rmSync(tmp, { recursive: true, force: true });
   });
@@ -390,19 +394,22 @@ describe("streamBusPrompt — per-agent bookkeeping (#213, split per #218)", () 
   });
 
   it("WARNs and defers to #227 when the agent crosses the threshold — without rotating", async () => {
-    const settingsFile = join(frozenCwd, ".claude", "claudeclaw", "settings.json");
-    mkdirSync(join(frozenCwd, ".claude", "claudeclaw"), { recursive: true });
+    // Point config at a settings file inside this test's temp dir (removed in
+    // afterEach) so we never read or write the developer's real settings.json.
+    const settingsFile = join(tmp, ".claude", "claudeclaw", "settings.json");
+    mkdirSync(join(tmp, ".claude", "claudeclaw"), { recursive: true });
     writeFileSync(settingsFile, JSON.stringify({ session: { autoRotate: true, maxMessages: 2 } }));
+    _setSettingsFileForTests(settingsFile);
+
+    const warnSpy = spyOn(console, "warn").mockImplementation(() => {});
     try {
       await reloadSettings();
       await createSession("22222222-2222-2222-2222-222222222222", "beta");
       await incrementMessageCount("beta"); // 0 -> 1; the next turn trips the gate
-      const warnSpy = spyOn(console, "warn").mockImplementation(() => {});
 
       await driveOneTurn("beta"); // 1 -> 2, needsRotation(2 >= 2) === true
 
       const warned = warnSpy.mock.calls.some((c) => String(c[0]).includes("deferred to #227"));
-      warnSpy.mockRestore();
       expect(warned).toBe(true);
 
       // The rotation MECHANISM is not wired: the session is counted, not reset.
@@ -410,7 +417,7 @@ describe("streamBusPrompt — per-agent bookkeeping (#213, split per #218)", () 
       expect(after?.messageCount).toBe(2);
       expect(after?.sessionId).toBe("22222222-2222-2222-2222-222222222222");
     } finally {
-      rmSync(settingsFile, { force: true });
+      warnSpy.mockRestore();
     }
   });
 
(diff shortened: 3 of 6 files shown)


TerrysPOV (changes requested):
Thanks for the fast turnaround on this. The bookkeeping side is genuinely solid — the `agentId` threading through `incrementMessageCount` / `peekSession` / `needsRotation` / `rotateSession`, the best-effort try/catch posture mirroring the receipt-chain "instrumentation must not break the bus" pattern, the inner guard for pre-`loadSettings()` boot — all the right shape. Docstrings on the new `rotateSession` overload + the `runner.ts` comment update are also clean.

Holding the merge though. Three critical findings surfaced in the 5-agent review that go to the underlying mechanism, plus one medium. Want to walk through them properly because they reframe what we agreed yesterday too.

## Findings

### 1. CRITICAL — Bus rotation doesn't actually rotate the live conversation

`src/rotation.ts` (169 lines, full file) has zero references to `sessionManager`, `restart`, `spawnAgent`, or `respawn`. The function only does filesystem work: write summary, rename `session.json` to backup, unlink it.

But the bus runtime spawns the agent PTY ONCE at boot with `--session-id <uuid>` and keeps it alive. The conversation accumulates in the SAME JSONL the live `claude` process is writing. Rotating `session.json` on disk doesn't touch the PTY. After "rotation," the next `streamBusPrompt` routes to the SAME live PTY, attached to the SAME claude process, accumulating the SAME context. The session.jsonl keeps growing. Per-prompt latency keeps creeping up.

**Net: #213's underlying claim — "session.jsonl grows unbounded on bus mode" — is not actually fixed by this PR.** The accounting is fixed; the rotation itself isn't.

### 2. CRITICAL — Rotation becomes one-shot per daemon lifetime

After `resetSession(agentId)` unlinks `agents/<id>/session.json`:
- `peekSession(agentId)` returns null on subsequent calls (`loadSession` reads, catches, returns null)
- The rotation gate at `webui-bridge.ts:135` requires a truthy `peeked` → never re-evaluated
- `incrementMessageCount(agentId)` hits `sessions.ts:129` (`if (!existing) return;`) → silent no-op

The file only gets re-created when something calls `defaultResolveSessionId(agentId)` (`agent-resolver.ts:116-122`), which fires on fresh PTY spawn. Since the PTY isn't restarted (Finding 1), `session.json` is never re-created. So even ignoring Finding 1, the rotation gate is single-fire per daemon lifetime — and the counter stays at 0 from rotation onwards.

### 3. CRITICAL — Session-id collision in `generateSummary`

`src/rotation.ts:105-106`:

```ts
const proc = Bun.spawn(
  ["claude", "-p", summaryPrompt, "--resume", sessionId, "--output-format", "text"],
  ...
);
```

`sessionId` here is the LIVE PTY's session id. Per `session-manager.ts:593` comments, claude explicitly refuses an `--resume` against a session that's already locked. So the spawn almost certainly fails — silently, since the failure is swallowed by the surrounding `try/catch (e)`. Operator gets a backup of `session.json` but no actual summary content to feed the next session. Worst case it races against the PTY's JSONL writes.

### 4. MEDIUM — No tests for the observable bookkeeping

The PR adds the entire #213 fix to two non-trivial files and ships zero unit tests. The `incrementMessageCount(agentId)` call site + `needsRotation` / `rotateSession` trigger are exactly the verifiable claims of #213. A test stubbing `runPrompt` to return `{ok:true}` and asserting `incrementMessageCount(agentId)` was called would close the loop; a second test with a session pre-populated past `maxMessages` asserting `rotateSession(config, agentId)` was invoked would close the rotation gate. Per `~/.claude/rules/testing.md` + SDC stage 5, this should land with coverage.

## Reframing yesterday's analysis

R3 also surfaced something important. I framed the `"agent sessions don't rotate"` comment in `runner.ts` as stale guidance yesterday. R3 verified via `git log -S` it originated in commit `ec4040c` (May 2026), and **the design reason was that legacy `execClaude` invocations are stateless one-shots** — rotation is meaningless when each call spawns a fresh `claude --print`. That part of the convention is still correct.

The bus-PTY case IS genuinely different (long-running, accumulating). But the fix here doesn't actually achieve rotation for that case either. So **neither code path delivers rotation for a long-lived conversation**. This is a bigger surface than #213 framed, and that's on me for not seeing it yesterday — apologies for the incomplete recommendation. Issue-body-B + comment-A was necessary but not sufficient.

## Direction — Terry's strong preference is 4a

We talked through four options for the underlying rotation mechanism:

- **4a — Restart the PTY after `resetSession`** by calling `sessionManager.restart(agentId)` so the next `streamBusPrompt` lands on a fresh PTY with a new session id. The summary becomes the bootstrap context for the new PTY.
- 4b — Inject `/compact` into the live PTY via `send_prompt_stream`.
- 4c — Defer rotation to daemon restart; counter becomes WARN-only.
- 4d — Periodic supervisor-driven respawn.

**Terry's strong preference is 4a.** It's the option that actually delivers what #213 promised — rotation on a long-lived agent PTY without requiring daemon bounce. It also keeps the rotation mechanism explicit + deterministic (vs 4b's dependency on `/compact` semantics in the CLI, which iterates) and doesn't surrender the goal like 4c/4d.

Concrete shape of what 4a would need:

1. Move the session-id collision aside: `generateSummary` should run AFTER the PTY restart, against the now-backed-up session.json's saved sessionId (which is no longer the live one). Or copy the JSONL aside and summarize from the copy. Either way, summary input must not be the live session id.
2. After `resetSession(agentId)`, call `sessionManager.restart(agentId)` (new method on the session manager — drops the existing PTY, calls `defaultResolveSessionId(agentId)` which mints a new uuid + writes session.json, spawns a fresh PTY with the new session id).
3. The new PTY's bootstrap prompt includes the summary so the conversation continues with compressed context. This is where the SUMMARY.md content lands.
4. `incrementMessageCount(agentId)` after restart finds the freshly-created session.json (Finding 2's problem goes away).
5. Tests covering: bookkeeping fires, rotation triggers at threshold, PTY restart is invoked, new session.json is written, summary is consumed by the new PTY's bootstrap.

## Options for splitting the PR

The bookkeeping-only piece (the `agentId` threading + the inner try/catch boot guard) is independently useful for diagnostics even if rotation doesn't fire correctly. Two paths forward:

- **A. Split.** Land the bookkeeping piece as-is in a smaller PR (with tests), file a follow-up issue for 4a-shaped rotation. Smaller, lower risk, faster merge.
- **B. Rework this PR.** Stay on this branch, layer 4a on top, add tests for both the bookkeeping AND the restart. Bigger but ships the actual fix.

Lean is yours — both are reasonable. If you split, I'd add a comment in `webui-bridge.ts` noting that the rotation gate fires but the rotation mechanism is incomplete, and reference the follow-up issue so it's not invisible.

## On #217 ordering

I flagged yesterday that #217 (silent-drop synth-reply safety net) must land AFTER #218 because the synthesized reply needs to be attributed to the agent's session. That stands — but with 4a in mind, the attribution path is even cleaner since the restart preserves a clear "before/after" boundary on session id.

Happy to walk through any of this further if useful. The receipt chain is doing exactly what it was supposed to — exposing that even the proposed fix doesn't fix the real thing. That's not a knock on this PR; it's the validation loop working.

Awaiting your direction on split vs rework.

Nibbler1250:
Went with **Split** (option A). Force-pushed this branch down to the bookkeeping-only piece.

What's here now:
- `streamBusPrompt` threads `agentId` through `incrementMessageCount` so per-agent turn counts land on the agent's own `session.json` — the part you flagged as solid.
- The rotation gate still evaluates, but when it trips it **WARNs and defers to #227** instead of running the filesystem rotation. That removes all three criticals at once: no PTY-less rotation pretending to fix #213 (1), no one-shot-gate dead-end (2), and no `--resume` collision against the locked live session (3) — the broken `rotateSession(agentId)` path is gone.
- `rotation.ts` reverted to main (the `agentId` overload went with it).
- Tests added (finding 4): increment fires on success; the gate WARNs at threshold *without* rotating; bookkeeping no-ops cleanly when there's no session to count.

Follow-ups filed:
- **#227** — the 4a restart-based rotation (your 5-step shape); this is the actual #213 fix.
- **#226** — the `sessionManager.restart(agentId)` primitive #227's step 2 calls (also the shared mechanism for the control-plane + wedged-prompt-timeout respawns).

So this PR no longer closes #213 — #227 does. Title/body updated to match. And noted: #217 lands after the rotation work for clean session attribution.

TerrysPOV (changes requested):
### Code review

Re-review of `d81c1f0` — the split worked. All three CRITICALs from the 2026-06-03 review are resolved (no PTY-less rotation, no `resetSession` dead-end, no `--resume` collision). Production code in [`src/bus/webui-bridge.ts`](https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/d81c1f0a5124f9a1b5b66d5e26f2fc588a1e79ea/src/bus/webui-bridge.ts#L122-L162) is clean.

Found 1 issue:

1. The new "WARNs and defers to #227" test writes a stub settings file to the developer's real `.claude/claudeclaw/settings.json` and then `rmSync`'s it. `SETTINGS_FILE` is frozen at module-load `process.cwd()` ([`src/config.ts:12-13`](https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/d81c1f0a5124f9a1b5b66d5e26f2fc588a1e79ea/src/config.ts#L11-L14)) and the test writes to `frozenCwd` (the repo root) rather than `tmp/`. On the success path the original config is unrecoverable; on a mid-test throw the file is left at `{"session":{"autoRotate":true,"maxMessages":2}}`. Anyone who runs `bun test src/bus/__tests__/webui-bridge.test.ts` from the repo root and has a populated `.claude/claudeclaw/settings.json` will lose it.

   https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/d81c1f0a5124f9a1b5b66d5e26f2fc588a1e79ea/src/bus/__tests__/webui-bridge.test.ts#L390-L413

   Fix options: (a) back up the existing file's contents before `writeFileSync` and restore in `finally`; (b) expose a `_setSettingsFileForTests(path)` hook in `src/config.ts` so the test points at `join(tmp, "settings.json")` instead of the real frozen path; (c) skip the test entirely if a real `settings.json` exists at `frozenCwd`.

Two lower-priority nits worth a look while you're in there:

- `reloadSettings()` populates the module-level `cached` and there's no teardown — the third test (`gamma`) inherits `{autoRotate: true, maxMessages: 2}` but passes only because `peekSession("gamma")` returns null and short-circuits. Future tests in this file that call `getSettings()` will get leaked values.
- `warnSpy.mockRestore()` at L402 isn't in a `finally` — if `driveOneTurn` throws between mock and restore, the spy persists across tests in the same worker.

Once the test-file path is fixed, this is good to merge.

TerrysPOV (approved):
All findings from prior review cleared.

Nibbler1250:
Pushed `03d4b02` addressing the review.

- **Test-file path (blocker):** added `_setSettingsFileForTests(path)` in `config.ts` (same shape as the existing `_setDefaultReceiptStoreForTests`) so the rotation-gate test points config at a `settings.json` inside its per-test tmp dir. No more writing/`rm`-ing the real `<cwd>/.claude/claudeclaw/settings.json` — `SETTINGS_FILE` is `let` now, only ever reassigned by that test hook.
- **Cache-leak nit:** `afterEach` now calls `_setSettingsFileForTests()` to restore the default path and clear `cached`, so neither the override nor cached settings leak across tests.
- **mockRestore nit:** moved into `finally`.

Full bus suite green locally (318 pass / 0 fail), CI green. Ready for re-review.

Issue #222, opened 2026-06-03
feat(bus): reconcile agent process-liveness with MCP connection — silent prompt drop on no-connection (control-plane gap behind #207)


Nibbler1250:
## Summary

Behind the intermittent agent wedges that motivated #207, there is a distinct **control-plane** failure that the receipt chain makes visible but nothing yet acts on: a bus agent's **process liveness** and its **MCP/IPC connection** are tracked independently, with no invariant binding them.

When the process is alive but no connection is registered for it (`connectionsByAgent` has no entry — during a respawn window, or after the MCP client drops), `BusCore.sendPrompt` calls `ipcServer.send(agent_id, …)`, which returns `false`; `core.ts` logs `No MCP connection for agent_id=<id>` and **drops the prompt** — no queue, no retry, no respawn. The agent stays **alive-but-deaf** indefinitely; nothing reconciles the two states. Symptom on a live deployment: the agent silently stops responding until a manual restart happens to land a process whose MCP connection establishes cleanly.

**Trigger amplifier:** the session-id collision churn (#213) repeatedly tears the connection down and re-establishes it slowly, widening the deaf window.

### Evidence (live)
- `error: No MCP connection for agent_id=default` (`core.ts:296`).
- With the receipt chain wired into the adapter (#211), the dropped prompt closes its receipt as `timeout` with **`stdin_written` absent** — i.e. it never reached the agent — while the `claude` process is alive and idle.

## Why this is not a duplicate of the in-flight stack

- **#209 / #211 (receipts)** — data-plane *observability* this builds on. The proposal adds a NEW terminal state `no_connection`, **additive** to the existing vocab, not a redefinition.
- **#217 / #215 (synthesize reply)** — fixes the **output** end of the pipe (turn completed, `reply` never called → text lost). This is the **input** end (prompt never reached the agent → no turn). Opposite ends, complementary.
- **#213 (session rotation)** — the trigger (collision churn); a non-goal here (logged, not fixed).

## Proposed approach (full SPEC linked below)

Invariants:
- An agent is `ready` iff *(process spawned)* ∧ *(`hello` received)* ∧ *(connection open)*.
- `sendPrompt` MUST NOT silently drop: a prompt to a not-ready agent resolves to **delivered** (after a bounded readiness wait) or an **observable** distinct terminal state `no_connection` — never silence.
[...]

PR #226, opened 2026-06-04
feat(bus): sessionManager.restart() — shared respawn primitive (fresh-session, post-mortem, backoff)

Nibbler1250:
Per your note on #218 — the shared `sessionManager.restart(agentId)` primitive, built once so all three respawn triggers consume one mechanism: control-plane reconciliation (#222, `stdin_written=false`), auto-respawn on a wedged-prompt timeout (the watchdog case, `stdin_written=true` + no turn), and agent-scoped rotation (#218, the "4a" path).

This PR lands **just the primitive**. #218 will be reworked to call it; #222 and the watchdog issue plug in the same way. No existing caller's behaviour changes until it opts in.

## API

```ts
restart(agent_id: string, opts?: RestartOptions): Promise<AgentProcess>
```

- **`freshSession`** (default `true`) — mint a fresh session id and respawn clean. `false` recycles the same id in place (claude `--resume`, same transcript). The old `restart()` always reused the id; that's the `--resume`-on-a-locked-session footgun you flagged on #218 — `freshSession: true` sidesteps it by not racing the lock.

- **Post-mortem, captured _before_ stop** (`src/bus/post-mortem.ts`): cwd, the last N session-JSONL lines, and the triggering reason → `~/.claude/claudeclaw/post-mortems/<ts>-<agent>.json`, `0600`, best-effort (a capture failure never blocks the restart). Every respawn leaves an audit trail instead of vanishing.

- **Backoff / anti-loop**: bounded to `MAX_RESTARTS_PER_WINDOW` (3) per `RESTART_WINDOW_MS` (1h) rolling window. The 4th restart of one agent inside the window logs a CRITICAL and **throws instead of respawning** — a crash-looping agent flips to operator-intervention mode rather than thrashing. This is the answer to the one-shot-vs-unbounded tension on #218: neither, it's bounded self-heal.

## Tests

26 tests: fresh-id respawn, in-place recycle, post-mortem contents + `0600`, the 4th-refused + recovery-after-the-window-clears path, and the session-id collision interaction. `bun test src/bus/__tests__/session-manager.test.ts` → 26/0. Typecheck delta 0, biome clean. Manifest versions bumped.

## Notes

- Self-contained: adds `restart()` + `post-mortem.ts` + tests only.
- Composes with the per-agent session tailer (#217 follow-up): a respawn stops the old tailer and the fresh spawn re-attaches, so the tailer follows the rotation with no extra wiring.
- Window/threshold and post-mortem retention are constants — happy to expose them as settings or change the defaults to whatever you want as house policy.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.137",
+      "version": "2.2.138",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.137",
+  "version": "2.2.138",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/session-manager.test.ts
+++ b/src/bus/__tests__/session-manager.test.ts
@@ -557,6 +557,106 @@ describe("restart() respawn primitive", () => {
     const recovered = await mgr.restart("rl-1");
     expect(recovered.agent_id).toBe("rl-1");
   }, 30000);
+
+  it("reason='rotation' bypasses the crash-loop budget", async () => {
+    const clock = 2_000_000;
+    const mgr = new SessionManager({
+      commandOverride: "/bin/cat",
+      argsOverride: [],
+      busSocketPath: "/tmp/test-bus.sock",
+      postMortemDir: mkTmp("ccaw-pm-"),
+      persistRotatedSessionId: async () => {},
+      now: () => clock, // frozen — every restart shares one window
+    });
+    cleanups.push(() => mgr.stop("rot-budget"));
+
+    const agent = mkAgent({ id: "rot-budget" });
+    await mgr.spawnAgent(agent, "cron");
+
+    // Far more than MAX_RESTARTS_PER_WINDOW rotations in one frozen window:
+    // rotation has its own backpressure, so it never trips the gate.
+    for (let i = 0; i < 6; i++) {
+      const r = await mgr.restart("rot-budget", { reason: "rotation" });
+      expect(r.agent_id).toBe("rot-budget");
+    }
+    expect(mgr._list()).toContain("rot-budget");
+
+    // ...and the bypass didn't consume budget: a failure-driven restart still
+    // gets its full allowance right after the rotation burst.
+    for (let i = 0; i < 3; i++) {
+      await mgr.restart("rot-budget", { reason: "wedge" });
+    }
+    await expect(mgr.restart("rot-budget", { reason: "wedge" })).rejects.toThrow(/rate limit/);
+  }, 30000);
+
+  it("serializes concurrent restart() of the same agent (coalesces, counts once)", async () => {
+    let clock = 3_000_000;
+    let persistCount = 0;
+    const mgr = new SessionManager({
+      commandOverride: "/bin/cat",
+      argsOverride: [],
+      busSocketPath: "/tmp/test-bus.sock",
+      postMortemDir: mkTmp("ccaw-pm-"),
+      persistRotatedSessionId: async () => {
+        persistCount++;
+      },
+      now: () => clock,
+    });
+    cleanups.push(() => mgr.stop("concur-1"));
+
+    const agent = mkAgent({ id: "concur-1" });
+    await mgr.spawnAgent(agent, "cron");
+
+    // Two overlapping restarts fired in the same tick: the second must
+    // coalesce onto the first's in-flight promise — same process back, a
+    // single mint, and a single budget entry (no undercount, no
+    // `already spawned` throw from the loser).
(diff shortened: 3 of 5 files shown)


Nibbler1250:

A multi-agent review (per-dimension finders → adversarial verifiers) ran over this PR: **5 confirmed**, 3 dismissed as false-positive/intended.
All confirmed findings are **addressed in f5222b5**.

### 🟠 MEDIUM — persist/mint failure between stop() and spawnAgent() leaves the agent permanently down and un-restartable
*src/bus/session-manager.ts:962-985 (restart())*

restart() runs `await this.stop(agent_id)` (which deletes the agent from `this.agents` via finalise()), THEN `await persist(agent_id, fresh)`, THEN `this.spawnAgent(...)`. If `persist` rejects (createSession write hits EACCES/ENOSPC/EROFS, or a custom persistRotatedSessionId throws), restart() rejects AFTER the old process is already torn down and removed from the registry, but BEFORE spawnAgent runs and BEFORE `history.push`. The agent is now dead, de-registered, and not even counted against the rate limit. A caller that catches the error and retries restart() gets `cannot restart unknown agent <id>` because the record is gone — there is no recovery path through restart().

**Fix applied:** Mint + persist the fresh session id BEFORE calling stop() (so a persist failure aborts restart() with the agent still live and registered), OR wrap the persist in best-effort try/catch so a failure logs but still proceeds to agent.session_id update + spawnAgent (matching the best-effort persist posture used elsewhere in this file). Either keeps a persist failure from stranding the agent down with no recovery path through restart().

### 🟠 MEDIUM — Legitimate session rotation shares the same 3/hour budget as crash-loop restarts, so a busy agent can be denied a needed rotation
*src/bus/session-manager.ts:920-985 + const MAX_RESTARTS_PER_WINDOW*

The rate limiter pushes `nowMs` into restartHistory for EVERY restart regardless of `reason`, and the gate refuses the 4th in an hour. But restart() is the shared primitive for THREE triggers, one of which (session rotation by message/age threshold, #213) is a normal, expected event on a high-traffic agent. If an agent legitimately rotates 3 times in an hour, the 4th rotation throws `restart rate limit exceeded` and is refused — leaving the live claude writing the ever-growing JSONL that rotation exists to bound, defeating #213's purpose. Conversely, a few benign rotations can exhaust the budget so a genuine wedge can no longer be auto-recovered.

**Fix applied:** Key the crash-loop guard on failure-driven triggers only. Either (a) skip the budget entirely for reason==='rotation' (don't push nowMs and don't gate on it), or (b) keep separate per-reason histories so wedge respawns and healthy rotations don't share one counter. Rotation has its own backpressure (message/age thresholds), so it doesn't need crash-loop protection; the guard exists for crash-on-boot/bad-config wedge loops.

### ⚪ LOW — Concurrent restart() calls for the same agent are not serialized; both pass the gate and the loser throws mid-teardown
*src/bus/session-manager.ts:restart() — concurrency*

restart() has no per-agent in-flight guard. Two overlapping restart() calls both read `record`, both pass the rate-limit gate (history is only pushed at the very end), both write a post-mortem, both call stop(). After the first completes the respawn, the second's `this.spawnAgent(agent, origin)` trips the `already spawned` guard and throws — but only after it has already written a (now-stale) post-mortem and torn down the freshly-spawned process via its own stop()/mint, leaving the agent in an unpredictable state. This is partly inherited from the old stop()+spawn restart, but the new fresh-session mint + double history.push makes the interleaving more harmful.

**Fix applied:** Add a per-agent in-flight guard so restart() is non-re-entrant: maintain a `private readonly restartInFlight = new Set<string>()`. At the top of restart(), after the unknown-agent check, `if (this.restartInFlight.has(agent_id)) throw new Error(...)` (or coalesce by returning the in-flight promise), then `this.restartInFlight.add(agent_id)` and wrap the body in try/finally that deletes it.

### 🟠 MEDIUM — Concurrent restart() of the same agent races: undercounts rate-limit and the loser throws on the 'already spawned' guard
*src/bus/session-manager.ts:879-986 (restart)*

restart() has no per-agent serialization. The three intended triggers (control-plane wedge watchdog, data-plane model-hang, session rotation) are independent event sources that can fire for the SAME agent id concurrently. With two overlapping restart('x') calls:

1. Rate-limit undercount via array aliasing. Each call does `const history = (this.restartHistory.get(agent_id) ?? []).filter(...)`, which allocates a NEW array per call. Both read the same prior length, both pass the gate, and at the end both do `history.push(nowMs); this.restartHistory.set(agent_id, history)` on their OWN local arrays. The second `set` overwrites the first, so two real restarts get recorded as ONE.

**Fix applied:** Serialize restart() per agent_id with an in-flight promise map: at entry, if a restart for agent_id is already pending, either await/return the existing promise (coalesce) or chain after it (serialize); clear the entry in a finally. Independently, mutate the stored history array in place — e.g.

### ⚪ LOW — restartHistory map grows unbounded — entries are never deleted when an agent is stopped or exits
*src/bus/session-manager.ts:521,923-927,984 (restartHistory)*

restartHistory is only ever read/written inside restart() (lines 923/927/984). It is NOT cleared in stop() (line 840) nor in the onExit auto-cleanup that deletes the agent from this.agents (lines 626-632). So every agent id that is ever restarted leaves a residual number[] entry that lives for the life of the SessionManager, even after the agent is permanently stopped and removed from the registry. Each entry self-prunes its OWN timestamps to <=3 on the next restart() of that same id, but for an agent that is stopped and never restarted again, the entry is stranded forever.

**Fix applied:** Delete the per-agent restartHistory entry on permanent teardown. In stop()'s finalise(): add `this.restartHistory.delete(agent_id);` alongside `this.agents.delete(agent_id)`. Also add `this.restartHistory.delete(agent.id);` inside the spawnAgent onExit auto-cleanup block (next to `this.agents.delete(agent.id)`), so natural exits that bypass stop() are covered too. Both are idempotent.

PR #242, opened 2026-06-08
fix(bus): make the boot-dialog watcher resilient to CLI dialog/render changes

Nibbler1250:
## Problem

A daemon-spawned, headless `claude` can silently wedge at startup: the agent process stays alive and keeps reading its PTY, but it never reaches the REPL — so inbound prompts are read yet never produce a turn (no model call, no reply). It "recovers" only on a restart that happens to skip the dialog and re-wedges otherwise, which presents as an intermittent "the agent went deaf".

## Root cause

The boot-dialog watcher (`PtyAgentProcess.handleBootDialog`) dismisses `claude`'s interactive startup confirmation dialogs by matching **hardcoded raw substrings** against the PTY stream and pressing the right key. Two CLI rendering changes broke that:

1. **Cursor escapes interleaved in dialog text.** The dev-channels dialog title now renders with a cursor-move escape *inside* the text — the raw stream is `development\x1b[32Gchannels`, not `development channels`. So `buf.includes("development channels")` never matches and the dialog is never confirmed. (A terminal resolves the escape and shows clean text, which is why it looks fine to a human but not to a substring match on the raw bytes.)
2. **A new gate dialog.** A trust-folder prompt ("Is this a project you … trust?") now precedes the others and had no handler at all.

Either path leaves the dialog unanswered → the CLI never reaches the REPL → prompts are read but never turned into a turn.

## Fix

Drive the startup dialogs by **structure** instead of fragile per-title strings:

- **Strip ANSI before matching**, so interleaved cursor escapes can't silently break a match.
- Identify the dialog by its selected option (the `❯` row) plus the `Enter to confirm` affordance. When the default is a *proceed* action (trust-folder, dev-channels) → send Enter. The bypass-permissions dialog keeps its specific Down+Enter handler because its default is destructive ("No, exit").
- An unrecognised dialog whose default *looks* destructive is **left alone with a one-shot warning** rather than a blind keypress — so a future CLI change degrades to "stuck + a logged warning" instead of "pressed the wrong button".
- Dedup on the selected-option line → one Enter per distinct dialog, not one per render chunk.
- REPL-readiness is checked first (the `to cycle` footer hint) so a keystroke is never injected into a live REPL.

The only remaining version-specific assumption is the selector glyph; everything else tolerates dialog-text and render changes across CLI builds.

## Test plan

- New unit tests in `session-agent-process.test.ts`, including the regression itself: a dialog whose title is split by a cursor-move escape is still confirmed; the new trust-folder dialog is answered; a destructive-default dialog is **not** auto-answered (and warns once); dedup sends one Enter per distinct dialog; the watcher disengages on the REPL footer and then ignores dialog-looking text.
- Full `src/bus` suite: no regressions (pre-existing failures unchanged).
- Smoke build + Biome lint clean; plugin/marketplace versions bumped.
- Verified live against a daemon-spawned headless agent on a recent CLI build: pre-fix the startup dialog was never dismissed and prompts timed out before a turn; post-fix the agent reaches the REPL and inbound turns complete normally.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.137",
+      "version": "2.2.138",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.137",
+  "version": "2.2.138",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/session-agent-process.test.ts
+++ b/src/bus/__tests__/session-agent-process.test.ts
@@ -149,8 +149,46 @@ describe("PtyAgentProcess boot-dialog watcher (structural / ANSI-resilient)", ()
     } finally {
       console.error = origErr;
     }
-    expect(writes).toEqual([]); // no blind keypress on a destructive default
-    expect(warned).toContain("destructive default");
+    expect(writes).toEqual([]); // no blind keypress on a non-proceed default
+    expect(warned).toContain("non-proceed default");
+  });
+
+  it("fails safe on a destructive default phrased without an exit/cancel keyword (allowlist invert)", () => {
+    // A selected default phrased "Delete everything" matched no proceed verb,
+    // so the watcher must warn-and-wait rather than blind-Enter it — the case
+    // the old destructive-blocklist (exit|cancel|abort|…) would have missed.
+    const { handle, writes, emit } = bootPty();
+    new PtyAgentProcess("zeta", handle);
+    const origErr = console.error;
+    let warned = "";
+    console.error = (...a: unknown[]) => {
+      warned = a.map(String).join(" ");
+    };
+    try {
+      emit(" ❯ 1. Delete everything\r\n   2. Keep files\r\n Enter to confirm");
+    } finally {
+      console.error = origErr;
+    }
+    expect(writes).toEqual([]); // not a recognised proceed action -> no Enter
+    expect(warned).toContain("non-proceed default");
+  });
+
+  it("does NOT fire a second Enter when the bypass dialog re-renders after Down+Enter (Codex F3)", async () => {
+    // The bypass-permissions dialog is answered by the gated Down+Enter branch.
+    // On the redraw chunk the SAME dialog text is still on screen with "❯" now
+    // on the accept row — it must NOT fall through to the generic confirm
+    // branch and fire a second, blind Enter racing the deferred one.
+    const { handle, writes, emit } = bootPty();
+    new PtyAgentProcess("eta", handle);
+    emit(
+      "WARNING: Bypass Permissions mode\r\n ❯ 1. No, exit\r\n   2. Yes, I accept\r\n Enter to confirm",
+    );
+    // redraw after Down: selection moved to the accept row, dialog still up.
+    emit(
+      "WARNING: Bypass Permissions mode\r\n   1. No, exit\r\n ❯ 2. Yes, I accept\r\n Enter to confirm",
+    );
+    await new Promise((r) => setTimeout(r, 250)); // let the deferred Enter land
+    expect(writes).toEqual(["\x1b[B", "\r"]); // exactly Down then one Enter
   });
 
   it("disengages on the REPL footer and ignores later dialog-looking text", async () => {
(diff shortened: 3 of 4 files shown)


Nibbler1250:

A multi-agent review (per-dimension finders → adversarial verifiers) ran over this PR: **4 confirmed**, 4 dismissed as false-positive/intended.
All confirmed findings are **addressed in 59f6c4c**.

### ⚪ LOW — destructiveDefault regex word-matches `exit`/`quit`/etc. anywhere on the selected line → false-positive WARN+stall on a benign proceed option
*src/bus/session-agent-process.ts:264 (destructiveDefault regex)*

`const destructiveDefault = /\b(exit|cancel|abort|quit|decline|reject)\b/.test(selected);` runs against the entire selected ❯ option line. Any proceed option whose label legitimately contains one of these words as a normal word triggers the destructive branch, so the dialog is NOT auto-answered and the boot stalls until the 15s timeout. Reproduced: selected line `❯ 1. Continue and show exit codes` and `❯ 1. Yes, continue (this will not exit)` both classify as destructiveDefault=true → WARN/skip instead of Enter. Today's trust/dev-channels labels are clean, but the PR's whole premise is resilience to FUTURE CLI label changes; a future proceed label containing e.g.

**Fix applied:** Anchor the destructive-default check to the option's leading verb / a known negative prefix rather than scanning the whole line. e.g. after stripping the `❯ N.` prefix, test the first word(s): `/^(no|exit|cancel|abort|quit|decline|reject)\b/` on the de-numbered label, so a proceed label that merely mentions "exit" elsewhere (e.g. "Continue and show exit codes") is not misclassified.

### ⚪ LOW — Disengage marker widened from `shift+tab to cycle` to `to cycle` — broader match can disengage early and re-expose the late-dialog bug #195 fixed
*src/bus/session-agent-process.ts:222-227 (disengage gate broadened to "to cycle")*

The REPL-ready gate was changed to `if (buf.includes("to cycle")) { this.endBootDialogPhase(); return; }`. The comment argues `to cycle` is the version-stable core and "appears in no dialog," but it is a strictly broader substring than the previous `shift+tab to cycle`. If any pre-REPL output during boot (a warning, a dialog body, or model echo on a slow fresh-install boot) contains the phrase "to cycle," the watcher disengages permanently before a later real dialog renders — and per Codex P2 / PR #195 a dialog CAN render after early boot output, leaving it unanswered at the destructive "No, exit" default. The narrower original marker was deliberately chosen to avoid exactly this.

**Fix applied:** Tighten the disengage anchor so it cannot trip on arbitrary prose while keeping the CLI-resilience intent: match `tab to cycle` (covers both "shift+tab to cycle" and a future "tab to cycle permission modes") instead of the bare `to cycle`, or gate on the REPL footer shape (the `⏵⏵` mode glyph / "mode on" footer). Either is a stable core of the footer that does not appear in dialog or warning prose.

### 🟠 MEDIUM — Bypass dialog falls through to generic confirm branch on re-render → redundant/blind second Enter race
*src/bus/session-agent-process.ts:226-244 (handleBootDialog, bypass branch + generic branch)*

The bypass-permissions dialog is answered by the gated branch (`if (!this.answeredBypassPrompt && buf.includes("Yes, I accept"))` → sets the flag, sends Down then Enter 200ms later, returns). But `answeredBypassPrompt` only gates the BYPASS branch. The bypass dialog text stays in `bootDialogBuffer` and the SAME dialog is re-rendered across subsequent chunks (cursor move after Down, redraw). On the next chunk: `to cycle`? no; bypass branch skipped (flag set); it now falls through to the GENERIC confirm branch. That branch sees `Enter to confirm`, finds the `❯` selected row, and:
  (a) If the CLI has already re-rendered selection to `❯ 2.

**Fix applied:** Exclude the bypass dialog from the generic confirm branch. Simplest: after the bypass branch, add a guard before the generic `lastIndexOf("Enter to confirm")` logic — e.g. `if (this.answeredBypassPrompt && buf.includes("Yes, I accept")) return;` (the bypass dialog is still on screen and already handled). Equivalent: gate the whole generic branch on `!buf.includes("Yes, I accept")`.

### ⚪ LOW — Destructive-default detection is keyword-allowlist only → blind Enter on destructive defaults not phrased exit/cancel/abort/quit/decline/reject
*src/bus/session-agent-process.ts:239 (destructiveDefault regex)*

The safety gate that the PR description leans on ('an unrecognised dialog whose default looks destructive is NOT auto-answered') is implemented as `/\b(exit|cancel|abort|quit|decline|reject)\b/.test(selected)` against the selected `❯` option line only. This is genuinely brittle, not robust:
  - FALSE-SAFE (blind Enter on a destructive default): a selected default phrased as `Delete`, `Remove`, `Overwrite`, `Discard`, `Erase`, `Reset`, `Wipe` etc. does NOT match → treated as a proceed action → blind Enter. The PR's own test fixture even uses 'Delete everything' as the dangerous option, yet only passes because the SELECTED row happens to be 'No, exit'.

**Fix applied:** Invert the heuristic from a destructive-blocklist to a proceed-allowlist: only auto-press Enter when the selected `❯` line affirmatively matches an accept/proceed pattern (e.g. /\\b(yes|accept|trust|continue|proceed|allow|i am using this)\\b/), and otherwise warn-and-wait. That fails safe by default (unknown phrasing → stuck+warning rather than blind Enter) and removes the FALSE-WEDGE from benign options that merely contain a blocklist keyword.

PR #243, opened 2026-06-08
fix(bus): hold PTY prompts during session (re)init so they are not lost

Nibbler1250:
## Problem

A prompt sent to a PTY-resident agent can be silently lost: while the agent's session is (re)initialising, the keystroke is written to the PTY but the not-yet-ready TUI swallows it and no turn ever runs. The receipt shows the prompt was written (`stdin_written`) yet no turn starts and no reply comes back. It presents as an intermittent "delivered, no reply", and hits fixed-cadence callers (e.g. a heartbeat) far more than human-spaced, retried ones.

## Root cause

`BusCore.sendPrompt` types the prompt into the PTY via `streamPromptHandler` immediately, with no check that the agent is at a ready prompt. When `sendPrompt` runs during the window between `session.init` and `bus.events.replay_done` (the JSONL tailer's session-(re)init lifecycle), the TUI isn't accepting input yet and the keystroke is dropped.

## Fix

Gate PTY-stdin delivery at the single choke point (`sendPrompt` → `streamPromptHandler`):

- Track per-agent init state from `ingestSessionEvent`: `session.init` → holding, `bus.events.replay_done` → ready (flush).
- While holding, queue prompts (FIFO) instead of typing them into a not-ready TUI; flush them in order on `replay_done`.
- A per-agent backstop (default 4s, anchored to the EARLIEST `session.init` so a reconnect storm can't keep pushing the deadline out) flushes held prompts anyway if `replay_done` is missed — so a held prompt is never stranded (worst case = the previous deliver-immediately behaviour).
- Per-agent and additive: the IPC path is untouched, and delivery is immediate whenever the agent isn't initialising. The backstop interval is injectable for tests.

## Test plan

- New unit tests: prompt held during init; FIFO flush on `replay_done`; immediate delivery when ready; backstop flush when `replay_done` never arrives (never strands); per-agent isolation.
- Full `src/bus` suite green — no regressions. Smoke build + Biome lint clean; plugin/marketplace versions bumped.
- Deployed to a live daemon-spawned agent: boots clean, normal turns unaffected.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.137",
+      "version": "2.2.138",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.137",
+  "version": "2.2.138",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/core.test.ts
+++ b/src/bus/__tests__/core.test.ts
@@ -888,3 +888,98 @@ describe("BusCore IPC", () => {
     client.close();
   });
 });
+
+describe("BusCore delivery gate (session.init / replay_done)", () => {
+  let bus: BusCore;
+  afterEach(async () => {
+    await bus?.stop();
+  });
+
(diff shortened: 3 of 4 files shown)


Nibbler1250:

A multi-agent review (per-dimension finders → adversarial verifiers) ran over this PR: **1 confirmed**, 5 dismissed as false-positive/intended.
All confirmed findings are **addressed in 1603aad**.

### 🔴 HIGH — session.init/replay_done are emitted in INVERTED order for fresh/restart sessions, defeating the gate and adding up-to-backstop latency to the very path it targets
*src/bus/core.ts:480-487 (ingestSessionEvent) interacting with src/bus/jsonl-tailer.ts:155-169 (start) and 264-274 (handleRawLine)*

The gate assumes the order `session.init` (arm hold) then `bus.events.replay_done` (flush). That holds only when the JSONL file already exists at start(). For a fresh session OR a reconciliation-driven restart OR a /clear rotation, the session file is empty/absent, so JsonlTailer.start() skips drainFromOffset() (`if (existsSync(this.filePath))`) and calls emitReplayDone() immediately — the bus receives `replay_done` FIRST.

**Fix applied:** Make the gate order-independent rather than assuming session.init precedes replay_done.

PR #245, opened 2026-06-10
fix(telegram): inactivity-based receipt timeout (do not false-timeout a working turn)

Nibbler1250:
## Problem

The Telegram receipt timeout is a fixed `setTimeout` armed at prompt arrival and
fired after `receiptTimeoutMs` (5 min) **regardless of whether the turn is
progressing**. A turn doing real agentic work — subagents, multiple `Bash`
calls — can legitimately exceed that budget. When it does, the receipt closes as
`timeout` / `no_final_reply_within_timeout` **even though the agent is actively
producing output**.

Anything downstream that treats that receipt as ground truth (e.g. a wedge
watchdog built on the receipt chain) then reads a healthy, working agent as
wedged. Observed in a live deployment: a turn produced its reply **14 s after
the deadline**; the timeout had already fired, and an external watchdog
restarted the agent **3 s before** the reply was delivered — losing both the
turn and the answer to the user.

## Fix

Make the timeout **inactivity-based** instead of a wall-clock budget. Every
observed turn output (`response.text`, progress OR final) rearms the timer, so
the receipt closes as `timeout` only after `receiptTimeoutMs` of **genuine
silence** — a deaf agent or a model hang (the cases the timeout is actually for)
— and never as a hard cap on a live, working turn.

- New `armReceiptTimeout(key)` helper; called from the `response.text` handler
  on every output for an open receipt.
- The `no_final_reply_within_timeout` reason string is **unchanged**, so
  existing consumers keep matching; the semantics simply tighten to "no activity
  within the window".

## Test

Added a unit test: a turn emitting periodic progress past the budget stays open
(a fixed wall-clock timer would have fired); it closes as a timeout only once it
goes silent. Full telegram adapter suite: 37 pass / 0 fail. `biome` clean,
versions bumped.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.137",
+      "version": "2.2.138",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.137",
+  "version": "2.2.138",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/adapters/telegram/__tests__/telegram.test.ts
+++ b/src/adapters/telegram/__tests__/telegram.test.ts
@@ -1206,12 +1206,13 @@ describe("TelegramAdapter — inbound receipts (#211)", () => {
   });
 
   it("closes as timeout(max_turn_budget_exceeded) when activity never stops past the absolute ceiling", async () => {
-    // Ceiling = MAX_TURN_BUDGET_MULTIPLIER (4) * receiptTimeoutMs. With 30ms
-    // that is ~120ms from received_at. Emit progress every 20ms (always under
+    // Ceiling = receiptMaxBudgetMultiplier * receiptTimeoutMs. Pin both small
+    // (4 * 30ms = ~120ms from received_at) so the test is deterministic under
+    // the generous production default. Emit progress every 20ms (always under
     // the inactivity budget) for longer than the ceiling: the inactivity rearm
     // alone would keep the receipt open forever, but the ceiling forces a close
     // with a DISTINCT reason so a chatty-but-wedged agent stays visible.
-    adapter = await startAdapter({ receiptTimeoutMs: 30 });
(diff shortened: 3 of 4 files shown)


Nibbler1250:

A multi-agent review (per-dimension finders → adversarial verifiers) ran over this PR: **2 confirmed**, 1 dismissed as false-positive/intended.
All confirmed findings are **addressed in 15bb16b**.

### ⚪ LOW — response.edit_text turn output does not rearm the inactivity timeout, contradicting the PR's own "any assistant output proves the turn is alive" premise
*src/adapters/telegram/index.ts:557-600 (handleResponseEditText) vs. :480 (handleResponseText armReceiptTimeout)*

The PR establishes that the receipt timeout is INACTIVITY-based and rearms on "any assistant output for this turn — progress OR final" (comment at index.ts:475-479). It implements this by calling this.armReceiptTimeout(key) at the top of handleResponseText (line 480). However, the adapter also subscribes to the response.edit_text topic (index.ts:440-441) and handles it in handleResponseEditText (line 557), which streams updated text into the live turn message — this is also turn output that proves the turn is alive — yet it never calls armReceiptTimeout.

**Fix applied:** Add `this.armReceiptTimeout(key);` near the top of handleResponseEditText (right after `const key = this.convKey(agentId, target.chat_id);`), mirroring handleResponseText, so streamed edit_message output also proves turn liveness. Alternatively, if edit_text is intentionally excluded, add a code comment explaining why.

### 🟠 MEDIUM — Inactivity rearm removes the only upper bound on a receipt — a turn emitting progress forever never times out, blinding the external wedge watchdog
*src/adapters/telegram/index.ts:712-722 (armReceiptTimeout) + ~/agent/scripts/claudeclaw-healthcheck.sh:85-100*

The change converts the receipt timeout from a fixed wall-clock budget (armed once at prompt arrival in openTelegramReceipt) to an inactivity timer rearmed on EVERY response.text (progress OR final) via armReceiptTimeout(). There is no absolute cap. Any agent that emits a progress event more often than receiptTimeoutMs (default 5*60*1000 = 5 min) keeps rearming and the receipt NEVER closes as `timeout`.

**Fix applied:** Keep the inactivity rearm but reintroduce an absolute ceiling anchored at the receipt's received_at. In armReceiptTimeout, cap the rearmed deadline so it cannot exceed received_at + N*receiptTimeoutMs; when that ceiling is hit, close with a DISTINCT reason (e.g. reason:"max_turn_budget_exceeded") so the watchdog can still see the wedge while distinguishing it from a silent no-reply.

Nibbler1250:
Follow-up on the budget-ceiling finding: rather than a tight wall-clock cap (which would re-introduce the false-timeout of legitimate long turns — 15+ min agentic tasks are real), the absolute ceiling now defaults to a **generous 12x `receiptTimeoutMs`** and is **configurable** via `receiptMaxBudgetMultiplier`.

It remains a *coarse* safety-valve for deployments without external monitoring: it closes the receipt with the distinct `max_turn_budget_exceeded` reason — it never drops the turn (the reply still delivers via the reply tool) and is never read as a no-reply wedge. Precise stuck-vs-slow recovery is better handled by an operator-side liveness probe gated on real API-token flow (rchar) than by any wall-clock cap.

PR #248, opened 2026-06-12
fix(bus): reliable PTY submit — confirm turn started + hold through auto-compaction

Nibbler1250:
## Problem

A headless, daemon-spawned PTY agent delivers inbound prompts by typing them into claude's REPL and sending a submitting CR. Two related failure modes let a prompt be **accepted but never turned into a turn** — the agent goes silent and the only signal is a receipt that times out ~5 minutes later:

1. **Delivery race.** When the CR lands during a transient REPL render (bracketed-paste / redraw), the typed text sits in the input box **un-submitted**. No turn starts, no API socket opens. A reproducible wedge dossier showed `stdin_written` but **zero `ESTABLISHED :443`**.

2. **Auto-compaction collision.** When an auto-compaction seizes the REPL the instant a prompt arrives, it **swallows the submit CR**. A captured incident: `/compact` at T, "Compacted" ~100s later, **zero `end_turn`**, the API connection idle (keepalive only) the whole time. The receipt surfaces it as `no_final_reply` only at the 5-minute timeout.

Both are intermittent (~1/day) and, before this change, indistinguishable from a hung model — they were only ever recovered by a watchdog restart.

## Fix

A single, additive **delivery-confirm loop** after the submit CR, entirely at the PTY layer where the signal is real-time. It uses only version-stable TUI invariants:

- **Confirm a turn started.** The idle REPL footer (`"to cycle"`) is replaced by the streaming view the instant a turn begins. If the footer is still idle after a grace window, the submit CR did not land → re-send it (the text is already in the box). Bounded (2 nudges); a stray CR at an idle REPL is a no-op; a genuinely stuck REPL still degrades to the watchdog.

- **Hold through a compaction.** While `"Compacting"` is on screen the submit CR was swallowed, so wait it out — **without spending a nudge** (compaction runs ~100s, far longer than the nudge cadence). When it finishes the footer returns to `"to cycle"` and the idle-footer branch re-submits the still-typed prompt. Bounded by `maxCompactionWaitMs` (default 240s) so a stuck compaction degrades to the watchdog instead of looping.

Re-submit is **CR-only — it never re-types**, so there is no doubling risk if the input box survived compaction. All timings are injectable for tests.

## Why this shape

- **No new event plumbing.** The runner's `auto-compact-*` events only fire on the `claude -p` path, not the PTY/bus path; the `compact_boundary` → `session.compact` tailer event only fires *after* compaction. The live `"Compacting"` frame in the PTY tail is the only real-time start signal, and it's already captured for the boot-dialog watcher.
- **Version-resilience.** The marker is the bare `"to cycle"` core, which survives the CLI rename from `"shift+tab to cycle"` (older) to `"to cycle permission modes"` (2.1.168+). `handleBootDialog` deliberately uses the stricter `"tab to cycle"` to avoid disengaging on boot prose (#195); here the REPL is already up, so boot prose is not a concern and the bare core is the safer live-footer match. (Heads-up: the boot-dialog `"tab to cycle"` marker does not appear in the 2.1.173 footer string, so it may warrant a follow-up — happy to converge both call-sites on a shared constant.)

## Why it matters (delivery integrity)

An operator instruction that is accepted, acknowledged at the transport layer, and then silently lost is the worst failure mode for an agent that runs unattended: no error, no reply, and the loss only visible 5 minutes later as a timeout. This closes the two paths by which a prompt could be dropped between "typed into the PTY" and "a turn actually ran", and every unrecoverable case still degrades **observably** to the watchdog rather than hanging — no silent truncation of the failure.

## Testing

- New tests: re-nudge when the idle footer persists; do **not** re-nudge once a turn streams; wait out a compaction then re-submit; bounded give-up on a stuck compaction (no hang, no spurious submit).
- Bus suite green (`bun test src/bus/`), Biome clean (pinned 2.4.15), `tsc --noEmit` delta 0 vs base.
- Plugin + marketplace manifests bumped to 2.2.144 (version guards).

Markers confirmed against the live CLI bundle (`"Compacting"`, `"to cycle"` present in 2.1.173).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.143",
+      "version": "2.2.144",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.144",
+      "version": "2.2.145",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.143",
+  "version": "2.2.144",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.144",
+  "version": "2.2.145",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/session-agent-process.test.ts
+++ b/src/bus/__tests__/session-agent-process.test.ts
@@ -123,6 +123,52 @@ describe("PtyAgentProcess.send_prompt_stream delivery-confirm (#wedge)", () => {
     clearInterval(iv);
     expect(writes.filter((w) => w === "\r").length).toBe(1); // only the submit
   });
+
+  it("waits out an auto-compaction and re-submits when the REPL returns (socket=yes wedge)", async () => {
+    const { handle, writes, emit } = bootPty();
+    const proc = new PtyAgentProcess("z", handle, {
+      submitConfirmMs: 30,
+      maxSubmitNudges: 2,
+      maxCompactionWaitMs: 5000,
+    });
+    const p = proc.send_prompt_stream("hello");
+    // An auto-compaction seizes the REPL for ~150ms and swallows the submit CR
+    // (no turn). The footer then returns to its idle "to cycle" hint.
+    let compacting = true;
+    const iv = setInterval(
+      () =>
+        emit(
+          compacting
+            ? "\nCompacting conversation… (esc to interrupt)"
+            : "\n⏵ accept edits on (shift+tab to cycle)",
+        ),
+      8,
+    );
+    const stop = setTimeout(() => {
+      compacting = false;
+    }, 150);
+    await p;
+    clearInterval(iv);
+    clearTimeout(stop);
+    // The submit was re-sent AFTER compaction finished: initial CR + >=1 nudge.
+    // (Compaction-wait iterations must not have burned the nudge budget.)
+    expect(writes.filter((w) => w === "\r").length).toBeGreaterThanOrEqual(2);
+  });
+
+  it("gives up on a stuck compaction at maxCompactionWaitMs without hanging or spurious submits", async () => {
+    const { handle, writes, emit } = bootPty();
+    const proc = new PtyAgentProcess("z", handle, {
+      submitConfirmMs: 20,
+      maxSubmitNudges: 2,
+      maxCompactionWaitMs: 80,
+    });
+    const p = proc.send_prompt_stream("hello");
+    const iv = setInterval(() => emit("\nCompacting conversation…"), 8); // never ends
+    await p; // must resolve (bounded), not hang
+    clearInterval(iv);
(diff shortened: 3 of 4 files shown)


Nibbler1250:
Pushed a follow-up commit addressing an adversarial self-review of the confirm loop:

- **Exit mid-confirm now rejects** instead of resolving — the old in-loop `return` reported a phantom success for a prompt whose target had just died (inconsistent with the two pre-CR `_exited` checks, which throw).
- **Anchored the compaction probe** to the CLI status line (`Compacting conversation` / `Compacting at auto window`) instead of the bare word `Compacting`, which also appears in ordinary model output — a live turn streaming that word was stalling the loop (and the serialised writeChain behind it) up to `maxCompactionWaitMs`.
- **No silent give-up**: on an unconfirmed idle give-up the still-typed prompt is cleared (Ctrl-U) so it cannot merge into the next prompt, and any non-delivery emits a one-shot diagnostic — a silently-dropped prompt being the exact failure this loop exists to prevent.

Added regression tests for all three. Bus suite 371/0, Biome clean, tsc delta 0.

Nibbler1250:
## Adversarial review report

Method: 5 independent reviewers, each on a distinct lens (concurrency/state, TUI-version coupling, logic/bounds, test coverage, delivery-integrity/observability), over the diff; **every raised finding was then handed to a separate agent prompted to refute it** before it counted. Raised **36** → confirmed **18** → rejected **18**. The confirmed set collapses into 6 root issues. Four are fixed in the follow-up commit (confirm-loop robustness); three low/version ones are partially addressed + documented below.

### Confirmed → fixed

**1. Exit mid-confirm resolved as success instead of rejecting** · MED
The in-loop `if (this._exited) return;` resolved the promise, while the two pre-CR `_exited` checks `throw`. If the target dies during the confirm window, the caller/receipt records a delivered prompt for a now-dead process — masking the very wedge this loop exists to catch. → **Fixed: throws, parity with the other two checks.**

**2. Bare `"Compacting"` substring matched model output** · MED
The compaction probe matched `"Compacting"` anywhere in the 2000-char tail, which includes the model's own streamed tokens. A live turn streaming that word (an agent discussing log/db compaction, or reviewing this very change) was read as an in-progress compaction → the loop `continue`d, never reached the turn-started break, and held the serialised `writeChain` — blocking every queued prompt — up to `maxCompactionWaitMs` (240 s). → **Fixed: anchored to the CLI status line (`"Compacting conversation"` / `"Compacting at auto window"`) + regression test that streams the bare word and asserts no stall.**

**3. Give-up was silent** · MED
On nudge-budget exhaustion (REPL still idle) the loop resolved with no log/emit — the exact silent-drop the PR sets out to close. → **Fixed: one-shot diagnostic on any non-delivery (`turn-started` / `stuck-compaction` / `unconfirmed-idle`).**

**4. Un-submitted prompt stranded in the input box** · MED
On give-up the typed-but-unsubmitted text stays in the REPL input box; the next prompt's `pty.write(text)` concatenates onto it → a merged, corrupted prompt. → **Fixed: on `unconfirmed-idle` (REPL idle, safe) the line is cleared with Ctrl-U before resolving.**

### Confirmed → partially addressed / documented

**5. Turn-start is inferred negatively (absence of `"to cycle"`)** · LOW
A turn whose output contains `"to cycle"`, or a turn that completes within the grace window (footer re-rendered), can trigger a spurious nudge CR. Bounded: a stray CR at an idle REPL is a documented no-op. The diagnostic from #3 surfaces it if it ever degrades to a real miss. Positive turn-start detection deferred (2.1.173 exposes no stable streaming affordance — `"esc to interrupt"` is absent from the bundle).

**6. Both markers (`"to cycle"`, `"Compacting…"`) are CLI-render strings** · LOW
A future footer rename silently disables detection (same class as #193/#195). Partially covered by #3’s diagnostic (a non-delivery now logs, so a rename surfaces instead of silently wedging). Centralising the markers into one shared predicate used by both `handleBootDialog` and this loop is a sensible follow-up — and note the boot-dialog `"tab to cycle"` marker does **not** appear in the 2.1.173 footer (`"to cycle permission modes"`, no `"tab"`), so a shared constant would want the bare core.

**7. Test quality** · LOW
The original four tests use real timers with tight windows (flakiness risk) and never modelled the cases above. Targeted regression tests added (bare-`Compacting`-in-stream, clear+warn-on-unconfirmed, reject-on-exit). A fake-clock seam to remove real-timer races is deferred.

### Rejected (18)

Refuted on tracing the actual code — kept here so the confirmed set is signal, not noise:

- **Concurrency races on `recentOut`** (several): impossible under JS run-to-completion; `writeChain` serialises confirm loops, so there is no concurrent reset/append race.
- **`recentOut` reset loses a decisive frame between windows**: refuted — the reset is synchronous at the top of each iteration; no frame is dropped mid-window.
- **Double-CR submits an empty input mid-turn**: refuted — the nudge only fires while the idle footer is present, i.e. no turn is running.
- **Compaction-ends-into-streaming / stale-tail-first-iteration / deadline-only-checked-while-compacting**: refuted by the per-iteration reset and the actual control flow.
- **240 s compaction wait vs 300 s receipt timeout “inverts under contention”**: not a real inversion; 240 < 300 with margin and the handoff is intact.
- **`recentOut` injection surface**: verified **clean** — observation-only, never fed back as model input.
- Plus several test-gap findings whose severity was explicitly gated on the (refuted) bugs above.

---

Pushed in the follow-up commit on this branch. Bus suite 371/0, Biome clean, `tsc --noEmit` delta 0. Manifests bumped.

PR #250, opened 2026-06-13
fix(bus): require real output to confirm turn-start (idle-REPL wedge) — follow-up to #248

Nibbler1250:
## Follow-up to #248 — closes the last residual `socket=no` wedge

#248 made PTY submit reliable for the two wedge classes we had reproduced (CR lost to a paste/redraw race, and a turn eaten by auto-compaction). A third, rarer variant slipped through and reproduced **after** #248 was deployed (forensic dossier `20260612T080557`: agent idle ~7h, stdin written, `turn_started_at` absent, `socket=no`, recovered only by the watchdog at the 5-min receipt timeout).

### Root cause
The delivery-confirm loop treated **absence** of the idle footer (`"to cycle"`) as proof a turn had started. On a long-idle, quiet REPL the confirm window can be **completely empty** (no redraw at all), and `!"".includes("to cycle")` is `true` — so an empty buffer was read as *turn-started*, the loop stopped nudging, and the prompt sat un-submitted until the receipt timed out.

This is a silent failure mode: the message is accepted, no error is raised, and the prompt is simply never processed — the worst shape for any audited/operator-facing pipeline.

### Fix
Turn-start now requires **positive evidence** — real output **and** no footer. An empty/whitespace window is treated as inconclusive: it spends a nudge (better recovery odds if a CR was transiently swallowed) and then gives up **honestly** (`Ctrl-U` to clear the stranded input + a one-shot `warn`) rather than reporting a phantom success on a dead/idle process.

### Validation
- `src/bus/__tests__/session-agent-process.test.ts`: **17/17 pass**, including a new regression that drives a silent idle REPL (zero output) and asserts the loop keeps nudging then gives up honestly instead of false-confirming.
- The pre-existing `serialise` test relied on the same false-positive (its fake PTY never emits) — corrected to emit realistic streaming.
- `biome check` clean on the changed file; no new `tsc` errors.

### Scope
One self-contained commit on `src/bus/session-agent-process.ts` (+49/−5) plus the version bump. No behaviour change on the happy path — only the empty/idle window, which previously false-confirmed, now resolves correctly.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.145",
+      "version": "2.2.146",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.146",
+      "version": "2.2.147",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.145",
+  "version": "2.2.146",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.146",
+  "version": "2.2.147",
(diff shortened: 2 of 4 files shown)


Nibbler1250:
Ran a focused multi-agent ultra-review over this PR's confirm-loop (5 dimensions x adversarial verify). It surfaced 15 confirmed findings (2 high, 5 medium, the rest low/nit). Addressed:

**Fixed here (`7b52611`):**
- **Stuck-compaction left the prompt stranded.** The Ctrl-U line-clear ran only for `unconfirmed-idle`; a stuck compaction (`maxCompactionWaitMs` exceeded) also leaves the typed prompt in the REPL box where it concatenates onto the next prompt. Now clears on any non-`turn-started` outcome.
- **One-shot wedge warning silenced the whole process lifetime.** Re-armed after a confirmed turn so a later genuine wedge still warns.
- **Tests:** stuck-compaction now asserts the Ctrl-U clear; added a whitespace-only confirm-window test (the `.trim()` guard); removed the `fakePty` helper this PR orphaned when it switched the tests to `bootPty`.

**Fixed at the bus layer instead (companion PR #252):**
- **HIGH — consecutive-prompt false-positive.** `recentOut` is a single shared PTY tail; a neighbor turn still streaming fills it during the next prompt's confirm window, so the gate reads the neighbor's bytes as this prompt's turn-start (silent drop). The PTY scrape fundamentally can't attribute output to a specific prompt — but the bus can (the per-prompt tailer `prompt` event). So the bus now arms a turn-start verify for a delivery made during an active turn, defers it while the turn runs, and re-delivers once if the prompt never starts. This keeps the confirm-loop PTY-only and version-stable.

**Left as-is by design:** the bare `to cycle` footer substring (a nit: model output containing it could trigger a spurious nudge) — the comment already documents why the bare core is chosen over a stricter phrase (the `shift+tab to cycle` -> `to cycle permission modes` CLI rename), and the bus-level net now catches any resulting drop, so tightening it would trade version-stability for no real gain.

PR #252, opened 2026-06-13
fix(bus): attribute flush-verify per-prompt + compaction-aware re-delivery

Nibbler1250:
## What

After a **backstop flush** (the delivery gate force-flushes held PTY-stdin prompts on a timer because `replay_done` never arrived), the session can still be mid-(re)init and swallow the keystroke, so the prompt never starts a turn — the `socket=no` delivery wedge. This adds a turn-start verification: if the flushed prompt produces no turn within `flushVerifyMs`, it is re-delivered.

The first cut used a single agent-level timer cancellable by any turn activity. A self-review found that design re-introduced, in its own recovery path, the very races it was meant to close. This revision hardens it.

## Design

- **Per-prompt attribution.** Verify timers are keyed by the wrapped prompt string. A pending verify is cancelled only by a tailer `prompt` event whose ingested text matches that exact prompt (the JSONL tailer records the user line = the wrapped string we delivered). An unrelated later prompt's turn no longer satisfies — and silently drops — an earlier still-swallowed prompt. A second flush during the verify window **adds** to the pending set instead of discarding it.
- **Re-deliver through the gate.** Re-delivery goes through `deliverOrQueuePrompt`, so if a fresh `session.init` re-armed the hold, the prompt is re-queued rather than flushed into a still-initialising TUI.
- **Compaction-aware.** A stream handler that hasn't settled is legitimately working — it holds through auto-compaction (`maxCompactionWaitMs`, up to 240 s ≫ `flushVerifyMs`). The verify timer defers re-delivery while a handler is in-flight, eliminating a double-submit when a flush coincided with a long compaction.
- **Socket-close teardown.** Pending verify timers are torn down on socket close (and shutdown), so a stale timer can never type a keystroke into a restarted session that reused the same `agent_id`.

## Safety / correctness notes

- One re-delivery per swallowed prompt; the external health watchdog remains the ultimate net.
- No new keystroke is ever emitted into an initialising or compacting session — the two windows where a keystroke is unsafe.
- Re-delivery is attributed and idempotent per prompt; no path can double-submit.

## Tests

Full bus suite **375 pass / 1 skip / 0 fail**. Added two regression tests:
- re-delivers when only an **unrelated** prompt starts a turn (attribution).
- does **not** re-deliver while a delivery handler is still in-flight (compaction guard).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.147",
+      "version": "2.2.148",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.145",
+      "version": "2.2.146",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.148",
+      "version": "2.2.149",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.149",
+      "version": "2.2.150",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.146",
+      "version": "2.2.147",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.147",
+  "version": "2.2.148",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.145",
+  "version": "2.2.146",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.148",
+  "version": "2.2.149",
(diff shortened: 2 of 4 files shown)


Nibbler1250:
Pushed a follow-up commit (`43f13d4`) that reworks this after a self-review surfaced gaps in the first cut. The single agent-level verify timer was under-attributed and under-guarded — it could drop a swallowed prompt when an unrelated later prompt started a turn, re-deliver into a re-initialising TUI, and double-submit if a flush coincided with an auto-compaction longer than `flushVerifyMs`.

The revision makes verification **per-prompt and attributed** (matched to the tailer `prompt` event's ingested text), re-delivers **through the delivery gate**, is **compaction-aware** (defers while a stream handler is in-flight), and **tears down on socket close**. Two regression tests added; full bus suite green (375 pass / 0 fail). Body updated with the full design.

Nibbler1250:
Added a sibling fix (`06ddbb3`) for a related wedge captured overnight.

**Trigger:** the MCP/IPC socket dropped right at a prompt boundary (`send-failed: no-mcp-connection`) and reconnected ~370 ms later. The prompt fell through to an **immediate** PTY delivery — session was not (re)initialising, so the backstop path never engaged — the keystroke coincided with the reconnect and never started a turn, and the reconciler disarmed on *"reconnected during confirm window"* without verifying a turn had started. The prompt was lost until the external watchdog.

**Fix:** extend the per-prompt turn-start verification to this path. When an immediate PTY delivery coincides with a failed IPC send, verify a turn starts and re-deliver once if not. Held prompts remain covered by the backstop path; healthy deliveries (IPC send OK) are not armed, so there's no overhead on the normal path. One regression test added; full bus suite 376 pass / 0 fail.

This closes the immediate-delivery counterpart to the backstop case already in this PR — both now route through the same attributed, gated, compaction-aware re-delivery.

Nibbler1250:
Ran a second multi-agent ultra-review over this branch (5 dimensions × adversarial verify × synthesis). It surfaced — and `23f6542` fixes — a real HIGH that the test suite was masking:

**Attribution broke for multi-line prompts.** The flush-verify map keyed on the raw wrapped prompt string, but the PTY layer runs `sanitizePtyPromptText` (collapses `\r\n?/\n` → space, strips C0/DEL) before typing. claude records — and the tailer re-emits as the `prompt` event — that *sanitized* form. So for any multi-line prompt, the raw key (with literal newlines) never matched the ingested text, the healthy turn was never attributed, and the prompt was re-delivered (double-submit) in the backstop / IPC-blip windows. Multi-line chat messages are common, so this nullified the attribution guarantee for a frequent case. Every added test used single-line prompts (or echoed the delivered string directly), so the divergence was invisible.

**Fix:** key the verify on `sanitizePtyPromptText(wrapped)` (and sanitize the ingested text before lookup); keep the original wrapped to re-deliver verbatim. Added a regression test whose prompt embeds a newline. Full bus suite 377 pass / 0 fail.

**Known, watchdog-bounded follow-ups (not addressed here, low severity):**
- The one-re-delivery guarantee is per-verify-cycle, not absolute: under a sustained reconnect storm with repeated backstop flushes, a still-swallowed prompt can be re-delivered across cycles. Bounded by the external watchdog.
- Two byte-identical prompts in flight collapse to one verify key (text attribution can't tell them apart); only one is re-delivered if both are swallowed.
- The in-flight deferral re-checks once per `flushVerifyMs`; a sub-tick window exists where a just-settled handler and a not-yet-arrived turn-start event could overlap.

These are inherent to text-based attribution / timer granularity and are tracked as follow-ups rather than expanded scope on this PR.

Nibbler1250:
Closed the remaining confirmed LOW findings from the ultra-review at the source (`253258f`), rather than leaving them to the watchdog:

- **At-most-once (absolute).** After a re-delivery the verify entry is kept in the pending map and not re-armed, so a later backstop flush of the same prompt sees `pending.has(key)` and skips it. A prompt now gets one re-delivery for its whole lifetime — the per-cycle gap is closed.
- **Post-settle grace.** When the delivery handler settles, the verify waits `flushVerifyGraceMs` (default `min(1500, flushVerifyMs/4)`) before declaring the prompt swallowed. The turn-start `prompt` event reaches the bus asynchronously via the tailer and can lag the handler return, so without the grace a turn that started right as the handler returned could be re-delivered (the post-settle/pre-ingest window).
- **onClose teardown** is now covered by a regression test: a socket close clears the pending verify, so a reconciler restart (#222) and the verify never both act on the same prompt.

+3 tests (at-most-once across cycles, onClose clears verify, multi-line attribution). Full bus suite **379 pass / 0 fail**.

One finding is left as a documented inherent limit rather than fixed: two **byte-identical** prompts in flight collapse to one verify key, so if both are swallowed only one is re-delivered. The JSONL tailer records identical user lines identically — it cannot distinguish them either — so text attribution fundamentally can't tell them apart, and the only "fix" (injecting a correlation token into the user-visible prompt) is worse than the bug. The wrapped `ts` attribute already makes this a sub-millisecond-collision edge.

Nibbler1250:
Added a bus-level fix (`ff0fa79`) for the **HIGH** the #250 ultra-review surfaced — the consecutive-prompt false-positive — because it belongs at this layer, not in the PTY confirm-loop.

A prompt delivered while a neighbor turn is still streaming is only a queued keystroke in the REPL input box. The PTY confirm-loop watches a single shared `recentOut` tail, so the neighbor's stream reads as the new prompt's turn-start and the prompt is silently dropped. The PTY scrape cannot attribute output to a specific prompt; the bus can, via the per-prompt tailer `prompt` event already used for flush-verify attribution.

So: a delivery made while a turn is active arms a flush-verify, which is **deferred while any turn stays active** (the prompt legitimately waits behind the live turn — re-delivering sooner would double-submit), then re-delivers exactly once if no turn ever starts for it. Agent turn state is tracked via the `prompt` / `response.turn_end` tailer events and cleared on socket close and shutdown.

+2 tests (re-delivers if it never starts; does not if it starts its own turn after the neighbor ends). The existing unrelated-turn attribution test now ends the neighbor turn so the deferred verify can fire. Full bus suite 381 pass / 0 fail.

PR #255, opened 2026-06-15
fix(plugins): harden the MCP proxy/multiplexer bridge (security audit)

Nibbler1250:
## What

A focused **security + correctness** ultra-review of the MCP proxy/multiplexer subsystem — the tool-execution trust boundary (multiplexes MCP tool calls, persists/replays session state, authenticates identity, exposes HTTP). Reassuring headline: **no critical/high** — the auth model (per-plugin HMAC + bootstrap token) is sound. This PR fixes the one confirmed MED and the clear LOWs.

## Security

- **SSRF (MED)** — `health_url` was never host-allowlisted, even though `callback_url` is (precisely to block SSRF). A plugin registered with the bootstrap token could declare `health_url: http://169.254.169.254/...`; the on-demand health endpoint would then fetch it, follow redirects, and reflect reachability + raw connection errors. Fix: allowlist `health_url` at register like `callback_url`, `redirect: "manual"`, and a generic (non-reflecting) error.
- **`/api/plugin/list` was unauthenticated** — leaked the full plugin/tool inventory. Now requires the bootstrap token (same gate as `/register` and `/multiplexer/metrics`).
- **Audit-log injection / unbounded growth** — the auth-rejection audit logged a raw, attacker-controlled `X-Claudeclaw-Pty-Id`. Now charset-stripped + length-capped before it reaches the log.

## Correctness

- **Idle TTL bug** — session TTL was measured from `issuedAt` while `touch()` only bumps `lastUsedAt`, so an actively-used session was force-expired one hour after first mint. TTL is now an idle timeout (from `lastUsedAt`).
- **Lost update** — `loadAll()`'s prune-rewrite wrote back a pre-computed snapshot, clobbering `record()`/`touch()` writes that landed concurrently. It now re-filters the freshly-read state inside the mutex, dropping only the entries it flagged.
- **tmpfile leak** — `_atomicWrite` now unlinks the tmpfile on chmod/rename failure, and `garbageCollect` sweeps stale `*.json.tmp.*` left by a hard crash.
- **Restart backoff pinned at max** — `crashCount` never reset after recovery, so any later isolated crash got the max backoff. It now resets after the server stays up for a full crash window; a crash cancels the reset, so a genuine crash-loop still escalates.

## Tests

+2 (health_url SSRF guard, list requires token); 3 updated for the new list-auth and idle-TTL semantics. Focused plugin/gateway/persistence suites **252 pass / 0 fail**.

## Deferred (documented, not in this PR)

Lower-value / migration-cost items left as follow-ups: per-call `verifyCall` HMAC is dead code (real auth is the gateway token) — delete or wire; the persisted-record integrity hash is an unkeyed SHA-256 (detects corruption, not tampering — switching to HMAC needs a one-time migration); the plugin-invoke HMAC replay window has no nonce cache; stateless-bucket metrics tuples aren't released. None are reachable-critical.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.150",
+      "version": "2.2.151",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.150",
+  "version": "2.2.151",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/docs/plugin-integration.md
+++ b/docs/plugin-integration.md
@@ -21,7 +21,7 @@ Daemon-style plugins register themselves over HTTP and serve tool calls via a ca
 | `POST` | `/api/plugin/register` | Bearer bootstrap | Register plugin + tools |
 | `POST` | `/api/plugin/<n>/tools/<t>/invoke` | HMAC + timestamp | Invoke tool via callback |
 | `DELETE` | `/api/plugin/<n>` | Bootstrap or plugin token | Unregister |
-| `GET` | `/api/plugin/list` | None | List registered plugins + health |
+| `GET` | `/api/plugin/list` | Bearer bootstrap | List registered plugins + health |
 | `GET` | `/api/plugin/<n>/health` | None | Proxy to plugin health_url |
 
 ### Security model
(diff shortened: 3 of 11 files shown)


Nibbler1250:
Ran an adversarial ultra-review on this PR's own fixes (5 dimensions × verify) — the kind that caught a self-introduced regression in the wedge work. It found real bugs **in the fixes**, now closed in round 2 (`1d51e32`):

- **HIGH (regression)** — the `/api/plugin/list` bootstrap gate broke six tokenless liveness probes in the MCP skill template, including a health sweep that reported a *healthy* daemon as DOWN. The security gate shipped without updating the in-repo consumer. Fixed: SKILL.md probes now send the token, the liveness check treats `401` as up, and `docs/plugin-integration.md` shows the new auth contract.
- **MED (regression)** — the `loadAll` lost-update re-filter trusted a stale key set, so a session re-`touch()`'d/`record()`'d concurrently was still evicted — defeating the idle-TTL fix. The rewrite now re-validates each **fresh** entry (integrity + `lastUsedAt` TTL) inside the mutex.
- **MED (regression)** — the GC stale-tmp sweep used `.includes(".json.tmp.")`, which would delete a real persistence file for a server whose name contains that substring (data loss). Anchored to the actual tmpfile shape `/\.json\.tmp\.[0-9a-f]+$/`.
- **LOW** — `redirect:"manual"` reported a redirect-fronted `health_url` as down; a 3xx now counts as up (host responded, still never followed).
- **NIT** — IPv6 loopback `[::1]` was rejected by the `::1` allowlist; brackets are now stripped before the check.

Also added the replay guard the first audit flagged: the plugin-invoke HMAC accepted a captured request repeatedly within the 15-min window; a bounded seen-signature set now rejects replays.

+1 GC regression test; gateway/persistence tests updated for the new semantics; suites green.

**Still open (LOW, documented):** per-call `verifyCall` is dead code (delete vs document tradeoff — real auth is the gateway token); the persisted-record integrity hash is unkeyed SHA-256 (HMAC needs a one-time migration that drops existing sessions); stateless-bucket metric tuples persist for the process (bounded by tool count). None reachable-critical.

PR #257, opened 2026-06-16
Wire governance layer: budget enforcement, scoped policies, tamper-evident audit log

Nibbler1250:
## Wire the governance layer that was built but never invoked

A security + correctness audit of the governance layer found several features implemented and tested but **never called on the production request path**. This PR wires the ones that are clean wiring, with tests, and adversarially ultra-reviews its own diff (the review caught 4 MEDIUM + 1 NIT regressions in the first cut, all fixed here). Items that need new input plumbing or an operator surface are written up in a tracking issue rather than half-wired.

### What this wires

**CRITICAL — budget enforcement received no usage/cost.** `recordInvocationCompletion` was always called with `undefined` usage, so every budget read 0 spend and no budget could ever trip. Added `extractTranscriptUsage()` which reads the CLI session JSONL transcript (path-agnostic across the legacy stream-json and PTY paths) and sums this invocation's assistant-turn usage, then feeds `calculateEstimatedCost` → `recordInvocationCompletion`.

**HIGH — scoped (per-channel / per-user) policy rules were inert.** They loaded from disk but were never merged into evaluation. `getApplicableRules` now does `mergeScopedPolicies(loadedRules, getScopedRules(request))` (guarded, falls back to global rules on error). E2E test added.

**HIGH — audit log was not tamper-evident.** Each entry now carries `prevHash` + `sha256(prevHash + content)`; writes are serialized so the chain holds under concurrent fire-and-forget callers; `verifyChain()` detects edit/insert/reorder/mid-delete; `cleanupRetention()` re-seals survivors. Known limits (tail-truncation, multi-process, canonical JSON) are documented in-code and in the tracking issue.

**Correctness quick-wins.** Real `toolName`/`toolArgs` passed to `checkToolApproval`; `getBudgetState` filters usage by `policy.scope`; token/cost reads clamped against negative/NaN/Infinity.

### Fixes from the self-ultra-review (folded in)
- **budget fail-open:** `getBudgetState` no longer filters on `channelId`/`userId` — the producer (`recordInvocationStart`) persists them as `undefined`, so the filter would match 0 records and a per-channel budget would silently never block. Aggregates broadly (fail-safe) until the producer threads those fields.
- **policy fail-open:** `mergeScopedPolicies` keeps the **more restrictive** action on an id collision, so a lower-trust scoped ALLOW can no longer delete a same-id global hardened DENY.
- **audit race:** `cleanupRetention` now runs its read/reseal/rename through the same `writeChain` serialization (no clobbered append, no stale `lastHash`).
- nit: `safe()` clamp uses `>= 0` to match its non-negative intent.

### Tests
166 pass in the policy + governance suites (including new coverage for `extractTranscriptUsage`, scoped-deny enforcement via `evaluate()`, the same-id deny-preservation guard, and the audit hash-chain incl. tamper/removal/retention/empty cases). The one pre-existing `Telemetry > should return budget health` failure is unrelated to this diff (fails on `main` too).

### Not in this PR (tracking issue)
Skill-overlay deny rules (no caller + needs `skillName` source), real-time watchdog enforcement (`kill` is a dead branch; mid-turn hook is a feature), HITL approval surface (needs operator UI), the root-cause `channelId/userId/skillName` threading gap in `getPolicyContext`, the audit hash-chain hardening limits, and a deny-precedence doc note.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.151",
+      "version": "2.2.152",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.151",
+  "version": "2.2.152",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/__tests__/governance/usage-tracker.test.ts
+++ b/src/__tests__/governance/usage-tracker.test.ts
@@ -11,8 +11,12 @@ import {
   getAggregates,
   getUsageStats,
   resetUsageTracker,
+  extractTranscriptUsage,
 } from "../../governance/usage-tracker";
 import { join } from "path";
+import { homedir } from "os";
+import { mkdir, writeFile, rm } from "fs/promises";
+import { randomUUID } from "crypto";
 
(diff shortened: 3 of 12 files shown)


Nibbler1250:
Deferred items (skill-overlays, real-time watchdog, HITL surface, channelId/userId/skillName threading, audit hardening limits, deny-precedence doc) are tracked in #258.

PR #261, opened 2026-06-16
fix(bus): enforce the reply tool via a turn-end nudge, synthesized net as fallback (#240/#215)

### LINKED ISSUE #240: bus: refine synthesized-reply altitude when agent skips the reply tool (#217 follow-up)
## Context

Follow-up from the #217 review (note 3, non-blocking / altitude). When a turn ends with text but the agent never called the `reply` tool, `handleTurnEnd` synthesizes a `reply` from the raw concatenated `text` blocks of the whole turn.

## Problem

`thinking` blocks are excluded, but any working/planning prose the agent wrote as `text` — intending it as scratch and expecting to curate the user-facing message via `reply` — now ships verbatim to the surface. The `reply` tool exists precisely so the agent decides what the user sees; the safety net currently bypasses that judgment with a full dump.

This is a deliberate trade-off in #217 (deliver *something* rather than drop silently), and it's the right default for "agent finished but forgot to call reply." But the altitude is worth refining.

## Direction (not prescriptive)

Options:
- Treat a missing `reply` as a **bounded signal** — an operator alert, or a truncated "agent produced output without delivering" notice — rather than a verbatim text dump.
- Or tag the synthesized event (e.g. `synthesized: true`) so surfaces can choose to suppress, truncate, or label it differently from a real curated reply.

Refs #217.

Nibbler1250:
## Problem

When an agent ends a channel-driven turn with text but never calls the `reply` tool, the user gets nothing — its transcript text is not visible to them (issue #215). #217 added a safety net that synthesizes a delivery from the raw turn text, but that text was never curated through `reply`, so working/planning scratch shipped verbatim, masquerading as a curated answer (the altitude follow-up, #240).

## Approach — three layers, each with a distinct role

**1. Root cause first — nudge the reply tool (`bus/core.ts`).**
On the first miss, the bus injects a one-shot `<system-reminder>` telling the agent to call `reply` now, delivered over the same IPC + PTY-stdin seams as a prompt (best-effort — a failed IPC send does **not** trip the #222 respawn reconciler). A curated reply is what the user actually wants; the raw dump is a last resort. Bounded to one nudge per turn (`replyNudged`); the original turn text is stashed (`pendingNudgeText`) so a nudged-but-empty turn never loses the first turn's output. `replyNudge` (default `true`) is a kill switch on `BusCoreOptions`.

**2. Fallback — synthesize + tag (`bus/core.ts`).**
If the nudged turn *also* ends without a reply (or the nudge is disabled), synthesize the delivery and tag the `response.text` payload `synthesized: true` so surfaces can tell it apart from a curated reply.

**3. Presentation — label at the surface (`bus/types.ts` + adapters).**
A shared `withSynthesizedNotice(text, payload)` prefixes a concise notice (`⚠️ The agent ended its turn without sending a reply — raw output below:`), applied in the telegram, slack, and discord adapters (telegram on both the fresh-send and edit-in-place paths). Full text preserved. The bus stays a pure transport that adds metadata; presentation is the surface's call. **webui** (dashboard, different render path) is deferred — the flag is on the payload, adoptable later.

> Note on the nudge: it's an *in-the-moment* correction, not training. The model doesn't learn across sessions, so the nudge and the net are both permanent layers — the nudge isn't a temporary scaffold that "teaches" the model to comply.

## Tests

- core: nudge on first miss (no synthesis, reminder injected); a real reply after a nudge delivers curated text (not `synthesized`) and doesn't double-deliver; fallback synthesizes a tagged delivery when the nudged turn also skips reply; fallback preserves the original text if the nudged turn is empty. Existing synthesis tests pin `replyNudge:false` to target the fallback directly.
- adapters: synthesized delivery is prefixed in telegram (fresh-send **and** edit-in-place — mutation-verified), slack, and discord; reaction-only synthesized final emits no bare notice; a curated reply is unchanged.

596 bus + adapter tests pass, biome clean, no new typecheck errors, no full-suite regressions.

Closes #240. Implements the root-cause fix for #215 (which the #217 net only mitigated).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.187",
+      "version": "2.2.188",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.187",
+  "version": "2.2.188",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/adapters/discord/__tests__/discord-outbound.test.ts
+++ b/src/adapters/discord/__tests__/discord-outbound.test.ts
@@ -48,6 +48,21 @@ describe("DiscordAdapter — outbound response.text", () => {
     expect(h.rest.sent[0]?.text).toBe("done");
   });
 
+  it("prefixes the synthesized notice on a safety-net delivery (#240)", async () => {
+    adapter = await startAdapter(h);
+    h.bus.emit({
+      ts: Date.now(),
+      agent_id: "ops",
+      session_id: "s",
+      topic: "response.text",
+      payload: { text: "raw turn output", intent: "final", synthesized: true },
+    });
+    await new Promise((r) => setTimeout(r, 5));
+    expect(h.rest.sent[0]?.text.startsWith("⚠️")).toBe(true);
+    expect(h.rest.sent[0]?.text).toContain("without sending a reply");
+    expect(h.rest.sent[0]?.text).toContain("raw turn output");
+  });
+
   it("IGNORES the JSONL tailer's observability echo so replies aren't double-posted (#217)", async () => {
     // The silent-drop net (#217) wires the JSONL tailer into the live bus.
     // The tailer re-emits a raw per-block `response.text` for every assistant
--- a/src/adapters/discord/index.ts
+++ b/src/adapters/discord/index.ts
@@ -42,6 +42,7 @@ import {
   type BusOrigin,
   type PermissionRequest,
   isTailerOriginEvent,
+  withSynthesizedNotice,
 } from "../../bus/types";
 import { createDiscordGateway } from "./gateway";
 import { join } from "node:path";
@@ -521,8 +522,11 @@ export class DiscordAdapter {
       const payload = event.payload as { text?: string };
       const text = typeof payload.text === "string" ? payload.text : "";
       if (!text) return;
+      // #240: label a silent-drop safety-net delivery so it is not mistaken for
+      // a curated reply (no-op for normal replies). Full text preserved.
+      const outText = withSynthesizedNotice(text, event.payload);
       for (const channelId of targetChannels) {
-        void this.safeSendMessage(channelId, text);
+        void this.safeSendMessage(channelId, outText);
       }
       return;
     }
--- a/src/adapters/slack/__tests__/slack.test.ts
+++ b/src/adapters/slack/__tests__/slack.test.ts
@@ -552,6 +552,21 @@ describe("SlackAdapter — response.text outbound", () => {
     expect(api.sent[0]?.text).toBe("on it");
   });
 
+  it("prefixes the synthesized notice on a safety-net delivery (#240)", async () => {
+    adapter = await startAdapter();
+    bus.emit({
+      ts: Date.now(),
+      agent_id: "triage",
+      session_id: "s1",
+      topic: "response.text",
+      payload: { text: "raw turn output", intent: "final", synthesized: true },
+    });
+    await waitFor(() => api.sent.length > 0);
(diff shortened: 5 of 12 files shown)


TerrysPOV:
@claude

TerrysPOV (changes requested):
Reviewed the three commits. The core is well-built — I verified no double-delivery on the common path, the nudge path, or the synthesis path; no infinite nudge loop (`replyNudged` guards re-nudging); `agentTurnActive`/`currentTurnReplied`/`currentTurnFinalPublished` are all managed correctly across the new lifecycle sites; the `synthesized` tag + `withSynthesizedNotice` label are consistent across Telegram/Discord/Slack (WebUI passes the raw event, correct). Nice.

One blocker before it lands:

**Nudge silently strands the user when both transports are down.** `nudgeForReply` sets `replyNudged`/`pendingNudgeText`, resets the reply flags, then sends the nudge over IPC (failures swallowed — "best-effort") and via `deliverOrQueuePrompt` (silently returns when `streamPromptHandler` is null). If **both** IPC and PTY are unavailable (the deaf-agent scenario, #222) while the JSONL tailer is still running, the nudge is dropped, no new turn starts, no `response.turn_end` arrives, and the original text sits in `pendingNudgeText` until the reconciler respawns the agent and `onClose` clears `replyNudged`. That's a regression: pre-nudge, `turn_end` → immediate synthesis → the user got the text with zero delay; now they hang for one reconciler cycle.

Fix: track whether the nudge was actually delivered (IPC `send` returned true OR `streamPromptHandler` exists); if neither transport received it, skip the nudge-wait and synthesize immediately, exactly as before. (`replyNudge: false` is a working kill switch, but the hung-state logic should be closed regardless.)

Also, not a blocker but worth doing in the same pass:
- The cross-transport race test ("delivers exactly once when turn_end LOSES the race", `core.test.ts:~1470`) still passes but for a different reason now — with the default `replyNudge:true` + no transports, `turn_end` nudges (drops), and the single delivery comes from the late real reply, so it no longer exercises the synthesizer's dedup. The comment ("turn_end … synthesizes + delivers") is now factually wrong. Either set `replyNudge:false` to keep testing the synthesis dedup, or split into two tests (synthesis-dedup + nudge-then-late-reply) and fix the comment.
- Follow-up nit: `stop()` clears several per-agent maps but not the two new ones (`replyNudged`, `pendingNudgeText`) — same omission as the pre-existing maps; worth adding for symmetry.

It's also conflicting with current main, so it'll want a rebase. Close the nudge-drop hang and this is good.

Nibbler1250:
Thanks for the thorough read — blocker closed.

**Nudge-drop hang fixed.** `nudgeForReply` now returns whether *any* transport actually received the nudge — IPC `send()` returned `true`, OR a `streamPromptHandler` is wired (the PTY-availability signal, since `deliverOrQueuePrompt` no-ops without one). When neither did (the deaf-agent / #222 case), `handleTurnEnd` no longer `return`s to wait for a turn that will never start — it falls through and synthesizes immediately, exactly as the pre-#215 path did (zero delay). `replyNudge:false` remains the kill switch; this closes the hung state regardless.

Also in this pass, per your notes:
- Added a **deaf-agent regression test** (no IPC, no PTY handler → immediate synthesis, `synthesized:true`).
- The cross-transport race test (`#217 finding 2`) now exercises the synthesizer's dedup **for real** again: with the fix, default `replyNudge:true` + no transport synthesizes on `turn_end`, then the late real reply is deduped — so the comment is accurate again without needing `replyNudge:false`.
- `stop()` now clears `replyNudged` + `pendingNudgeText`.

Rebased onto current main (resolved the #240/#215 stack against the merged #265 markdown-HTML + #296 jobHandler additions — kept both sides), version bumped, `bun run lint` clean, bus core + telegram adapter suites 113/0.

TerrysPOV (approved):
Re-reviewed after Nibbler's update (`4ff3d8b`) — approving (supersedes my change-request). My blocker is resolved:

**The nudge no longer strands the user when both transports are down.** `nudgeForReply` now returns `ipcDelivered || ptyAvailable`, and the caller synthesizes *immediately* (zero-delay, exactly the pre-#215 path) when no transport received the nudge — with a log referencing #261/#222. That closes the deaf-agent hang I flagged; the reply-enforcement path is otherwise unchanged and still passes all the dedup/label/no-double-delivery cases. Claude Bot's CI review also LGTM'd it (596→623 tests green).

Maintainer commit from me:
- Rebased onto current main (bus source rebased cleanly — no code conflicts) + bumped to 2.2.188.
- Folded in Claude Bot's one LOW actionable item: wrapped the best-effort IPC `send` in `nudgeForReply` in try/catch, so a synchronous throw can't skip the PTY path or propagate to the turn-end handler (mirrors `sendPrompt`'s existing IPC guard). biome + build clean; bus + telegram + slack suites 623/0.

The earlier test-comment nit and `stop()` clear were follow-up-only and remain fine. Good to merge.

TerrysPOV (approved):

Issue #235, opened 2026-06-07
Root cause of the recurring bus-mode deaf-agent wedge: a bun-pty input-delivery bug (not an upstream model hang)


Nibbler1250:
## TL;DR

The recurring **bus-mode "agent goes silent / deaf agent" wedge** — where a prompt is written to the agent's PTY but never becomes a turn, the agent stays alive and idle in `ep_poll`, and the receipt closes as `timeout` — is **not** an upstream model hang (the #64496-style "model stalls during streaming"). It is a bug in the **`bun-pty` dependency** used by the `pty-stdin` supervision path.

Root cause found, reproduced live, and fixed.

## Root cause

`bun-pty`'s reader thread (`rust-pty/src/lib.rs`) treats **any** `read()` error on the PTY master as EOF (`Err(_) => break`). A *transient* error — `EINTR`, which the daemon hits frequently because it spawns many child processes and every `SIGCHLD` can interrupt the blocking master `read()` — therefore ends the reader, which sets the terminal's internal `exited` flag, after which **`bun_pty_write` silently no-ops** for the rest of the (still-alive) agent's life.

So: agent output keeps flowing (the data already read), but **agent input is permanently dropped** → prompts written via `send_prompt_stream` never reach claude → no turn → `timeout` receipt → "deaf agent."

## Evidence

- `strace` on the live agent: claude reads its pts continuously, but the injected prompt bytes are **never** in any `read()`; and the daemon-side `bun_pty_write` (FFI) produces **no `write()` syscall** at all (it early-returns `CHILD_EXITED`).
- A direct, non-bun-pty PTY (`script`) with the same claude flags delivers input fine — isolating the fault to `bun-pty`'s write path.
- Reading the rust source confirmed the `Err(_) => break` → `Msg::End` → `exited=true` → `bun_pty_write` early-return chain.
- After patching `bun-pty`'s reader to retry transient errors and rebuilding `librust_pty.so`, the agent receives prompts and replies reliably again (confirmed: 30+ real `reply` tool calls, receipts `turn_observed` ~3-4s).

## Upstream fix

Filed against the dependency:

- Issue: `https://github.com/sursaone/bun-pty/issues/41`
- PR: `https://github.com/sursaone/bun-pty/pull/42`

Fix: retry `Interrupted`/`WouldBlock` in the reader loop instead of breaking; only end on real EOF.

## What this means / recommendation

[...]

Issue #233, opened 2026-06-06
bus wedge: end_turn persisted but JsonlTailer misses it → #215 synthesize net never fires → 300s timeout


Nibbler1250:
## Summary

Intermittent "wedge": a bus-resident agent appears hung — it stops answering and only recovers on restart. A forensic capture reclassifies it as a **semantic wedge, not a process/stream hang**, and points at a coverage hole in the #215 silent-drop safety net: the turn's `end_turn` is written to the session JSONL, but the per-agent `JsonlTailer` doesn't publish `response.turn_end` for it, so the `core.ts` synthesize handler never fires → the turn times out at the 300s ceiling.

Filing to coordinate — we're actively on this and will dig further. Looking for input on a reliable repro and on the tailer's resume / seek-to-EOF semantics.

## Evidence (one captured instance)

Per-thread + socket forensic of the wedged process:
- Main thread idle in `epoll_pwait2`; HTTP client thread idle; no thread blocked on a socket `read`/`recvfrom`.
- Two ESTABLISHED `:443` sockets, both Send-Q=0/Recv-Q=0 on keepalive — **no request in flight**. A hung model/stream request would show a non-empty queue + a thread in `recvfrom`. Neither present.
- `compaction_events_in_session: 0`, modest session (~141 KB / 128 events) — no context pressure.

Session timeline:
- Prompt written to stdin; user event recorded; an **assistant turn followed**, then nothing for ~347s.
- The session JSONL **contains `stop_reason: end_turn`** for the wedged turn — the model finished normally.
- The wedged turn ended as **plain assistant text with no `reply` tool call** (unlike adjacent turns that delivered via the bus `reply` tool).
- Receipt: `final_state: timeout`, `reason: no_final_reply_within_timeout`, `duration_ms: 300001`.

**Conclusion of the capture:** execution finished; only the final-reply handoff was missing. This is a reply-delivery contract miss, not "model execution hangs."

## Root cause (current hypothesis)

`end_turn` IS emitted and persisted in the session JSONL, but the `JsonlTailer` does not publish `response.turn_end` for that turn → the `#215` synthesize path in `core.ts` (`response.turn_end` → synthesize `ingestReply`) never triggers → 300s timeout.

So the synthesize code (#217) looks correct; the gap is the **tailer ↔ live-session binding**. Candidate causes (all in `jsonl-tailer.ts` / `session-manager.ts`):

[...]

PR #263, opened 2026-06-16
docs: spec — re-enable claude -p as a per-agent runner toggle

Nibbler1250:
Design doc you asked for. Grounded in two empirical re-tests on the current CLI (2.1.178/2.1.179):

- **`claude -p` long-lived multi-turn works** — the ~3s downshift was the *interactive* REPL with non-TTY stdin (Spike 0.4), not `claude -p`.
- **`claude/channel` is still NOT supported on `-p`** — probed with a raw MCP server: claude advertises `{roots, elicitation}` only, emits zero channel notifications, declines permission-gated tools rather than routing via channel. Spike 0.6 holds.

**Key finding:** the toggle is already ~90% in the data model — `AgentConfig.supervision` overrides the per-origin runner, and `process-stream-json` *is* `claude -p` (already used for non-channel origins). So this is scoping + wiring + graceful degradation, not a new runtime.

Spec covers the parity matrix, the remaining wiring (stream-json inbound seam, hard-hiding `ask`/`request_human`, forcing `bypassPermissions`), config/UX (`runner: "claude-p"|"bus"`), governance posture, the switch story, and 4 open questions for you at the end.

Not wired to any code yet — review/comment away.


TerrysPOV (changes requested):
### Code review

Spec is solid on data — all three core technical claims (`AgentConfig.supervision` override at `session-manager.ts:593`, `process-stream-json` invocation, `defaultSupervisionFor` routing) match the source. Six completeness gaps the Phase B implementer will hit:

1. The spec doesn't acknowledge the deliberate migration history that moved the project off `claude -p`: PR #62 introduced the PTY migration, PR #159 flipped the runtime default to `bus`, PR #162 reframed PTY as a permanent first-class option. Re-introducing `claude -p` without addressing why the prior reversal is now reversible leaves the reader without the architectural rationale.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/351c359527d53967a29c36de36ce47c21c312737/docs/claude-p-runner-toggle-spec.md#L23-L29

2. §6 proposes a friendly `runner: "claude-p" | "bus"` alias but doesn't say how it integrates with the existing `runtime`/`supervision` parsing in `src/config.ts`. A new top-level alias needs its own parser, warn-on-unknown handling, and a precedence rule against the existing `supervision` field.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/351c359527d53967a29c36de36ce47c21c312737/docs/claude-p-runner-toggle-spec.md#L52-L57

3. §4 parity matrix is silent on slash-command availability (`/compact`, `/clear`). PR #106 reviewer noted that `isatty(0)` checks can quietly drop slash-command relay, which is exactly the surface a `claude -p` agent would lose. Worth enumerating in the matrix.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/351c359527d53967a29c36de36ce47c21c312737/docs/claude-p-runner-toggle-spec.md#L31-L44

4. §5.4 says "the JSONL tailer reads claude's session file regardless of runner, so \`response.turn_end\` detection should work — needs a test" without explaining *why* — i.e. that both runners write the same \`~/.claude/projects/<enc-cwd>/<session-id>.jsonl\`. A reader auditing silent-drop coverage for #233 / #235 has to infer the mechanism from one parenthetical, which makes future implementations easy to break.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/351c359527d53967a29c36de36ce47c21c312737/docs/claude-p-runner-toggle-spec.md#L45-L51

5. §5.1 implies per-mode dispatch routing inside \`deliverOrQueuePrompt\`, but \`BusCore.streamPromptHandler\` is a single global handler typed and documented as PTY-stdin-only. Wiring a stream-json seam means a \`BusCore\` interface change (second handler slot or mode-discriminated dispatch), not just routing — should be called out as part of the scope.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/b6900ca3ab6855f1c57bf3c86203dfdb31bde10e/src/bus/core.ts#L140-L210

6. §5.1 separately routes inbound stream-json prompts without referencing PR #140's rationale for the PTY delivery seam (consistent XML-wrapped channel format, security-hardened escaping). Cite or restate the prior-art so fallback-from-\`-p\`-to-\`bus\` agents remain consistent, and to confirm the stream-json seam preserves the same channel-XML wrapping (or explicitly substitutes it).

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/351c359527d53967a29c36de36ce47c21c312737/docs/claude-p-runner-toggle-spec.md#L45-L51

Nibbler1250:
Thanks Terry — these are exactly the gaps a Phase B implementer would trip on. Pushed rev 2 (a4eadd8) addressing all six:

1. **Migration history / why it's reversible** — added a subsection citing #62 (PTY migration), #159 (default→bus), #162 (PTY permanent first-class), and the rationale: the credit-billing driver is *delayed, not gone*, and since #162 already made `runtime`/`supervision` a permanent multi-runner posture, adding `claude -p` as a *parallel* first-class runner is consistent with that design — the Bus stays default, nothing in #62/#159/#162 is undone.

2. **Alias parsing in `config.ts`** — §6 now spells out the new parsing: a `parseRunnerAlias` mapper with warn-on-unknown (mirroring `VALID_RUNTIMES`' warn-and-fallback), a precedence rule (explicit `supervision` wins; `runner` only fills it when absent; conflicting values = parse-time warning), and a note that it's orthogonal to the top-level `runtime` field.

3. **Slash commands** — added a matrix row. Net: `process-stream-json` keeps `/compact`/`/clear`/`/quit` via `send_slash` writing `/<cmd>\n` to stdin (`session-agent-process.ts:180`, Probe 0.6 Q5). Added a review-note paragraph addressing the #106 `isatty(0)` concern explicitly — that risk is the non-TTY interactive path, not the stdin slash relay; only the `process` Windows-fallback drops it.

4. **§5.4 JSONL mechanism** — now states *why*: `claude` writes the transcript to `~/.claude/projects/<encoded-cwd>/<session-id>.jsonl` as a function of `--session-id` + cwd, not stdio mode, so the `JsonlTailer` observes `turn_end` identically under both runners. Called out as the mechanism a #233/#235 silent-drop auditor must rely on, plus an explicit test.

5. **§5.1 is a `BusCore` interface change** — rewritten to say it outright: `streamPromptHandler` is a single global PTY-stdin-only handler (`core.ts:140-210`), so wiring stream-json needs a second handler slot or mode-discriminated dispatch keyed on `SupervisionMode`, reviewed as a core change — not leaf routing. The #240/#215 nudge rides the same seam and follows the same dispatch.

6. **#140 channel-XML wrapping** — added: the stream-json seam must wrap inbound in the identical `<channel …>…</channel>` envelope with #140's hardened escaping (carried as the `text` of a stream-json user message), or explicitly document a substitute, so `-p`↔`bus` fallback agents see a consistent format. Don't hand a `process-stream-json` agent a bare prompt string.

The 4 open questions at the bottom are still genuinely yours to call (granularity, ask/request_human hard-hide vs degrade, alias naming, the capability notice). Happy to fold your answers into the spec.

TerrysPOV (changes requested):
### Code review

Spec rev 2 — all six prior CHANGES_REQUESTED points are genuinely addressed in the text (verified, not just claimed). PASS on each: migration history (§1), `parseRunnerAlias` + precedence (§6), slash-command parity row + stdin-relay note (§4), JsonlTailer mechanism via shared `~/.claude/projects/<enc-cwd>/<sid>.jsonl` (§5.4), explicit `BusCore` interface change (§5.1), and #140 channel-XML envelope mandate (§5.1).

One blocker, unrelated to the spec:

1. `src/adapters/telegram/format.ts` (+79) is a scope leak from open PR #265. The file is byte-identical (sha256 `a894815f70ccfeb8c4b96fa451acb7c69b3eabb648966c49f9a913ca0ed54a35`) and was introduced by commit `82cb262` whose message is purely docs ("docs: rev 2 — address #263 review …"). Strip it from the branch so this PR stays docs-only — otherwise it collides with #265 at merge AND ships the bugs I flagged on #265 (URL `&` double-escape; `$1`/`$&` replace corruption) into the docs PR.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/782a3d22d65747538760c95a451f1116f54e7a7e/src/adapters/telegram/format.ts#L1-L13

Two nit-level line-citation drifts I noticed but won't block on:
- §4 cites `session-agent-process.ts:180` but that's the PTY class's `send_slash`; the stream-json class (which actually writes to stdin as the spec describes) is at ~line 511.
- §5.1 range `core.ts:140-210` overshoots; `streamPromptHandler` is at line 142-143.

TerrysPOV (approved):
### Code review

Rev 3 (`b753bcb`) resolves the only blocker from my rev 2 review — `src/adapters/telegram/format.ts` is gone, no orphan imports remain. Spec text is byte-identical to rev 2, which already PASSed on all six original CHANGES_REQUESTED points (migration history, alias parsing, slash-cmd matrix, JSONL mechanism, BusCore interface, #140 channel-XML).

Version-bump revert (2.2.156 → 2.2.155) is correct policy — `plugin-version-guard.yml` regex `^(src/|commands/|prompts/|\.claude-plugin/)` doesn't match `docs/`, so a docs-only PR shouldn't bump. 3-way merge against current main (already at 2.2.156 post-#270) keeps main's bump cleanly.

Ready to merge.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/b753bcb05001362476e859551ff6510ab2fc7885/docs/claude-p-runner-toggle-spec.md#L1-L10

PR #265, opened 2026-06-17
fix(telegram): render Claude's markdown as HTML on the bus adapter

Magister:
## Problem

The Telegram **bus** adapter sent Claude's replies as raw text. Claude
emits markdown (`**bold**`, `*italic*`, `` `code` ``, `[link](url)`), so
users saw literal `**` and backticks instead of formatting. Sends and
edits also disagreed on `parse_mode`, so even where formatting was
attempted it was inconsistent between a first message and its later edits.

## Fix

- Add `markdownToTelegramHtml()` (`src/adapters/telegram/format.ts`) —
  converts the markdown subset Claude actually emits into Telegram-flavored
  HTML (bold, italic, code/pre, links, strikethrough), escaping the rest.
- Route all agent-reply output through `sendHtml()` / `editHtml()` helpers
  that convert + send with `parse_mode: "HTML"`. Edits now format
  identically to sends — no more raw-markdown edits over formatted sends.
- On Telegram's 400 for malformed HTML, retry as plain text so the message
  still lands (unformatted) rather than being dropped.
- Italic support covers both `_text_` and `*text*`. The `*…*` rule runs
  after `**bold**` is consumed and ignores bullet markers (`* item`) and
  stray asterisks.

## Tests

`bun test src/adapters/telegram/` → **53 pass / 0 fail**. Outbound
formatting test extended to assert bold + italic + code + link all render
as HTML.

> Note that all this code is Claude-generated. It works fine for me, so probably fine.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.184",
+      "version": "2.2.185",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.184",
+  "version": "2.2.185",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/adapters/telegram/__tests__/format.test.ts
+++ b/src/adapters/telegram/__tests__/format.test.ts
@@ -1,0 +1,75 @@
+import { describe, it, expect } from "bun:test";
+import { markdownToTelegramHtml, isTelegramHtmlParseError } from "../format";
+
+describe("markdownToTelegramHtml — code back-reference safety (#265 review)", () => {
+  // A string 2nd arg to String.replace interprets `$&`/`$'`/`` $` ``/`$$`. The
+  // restore step must use a function replacer so code content is inserted
+  // verbatim — otherwise these splice the match / rest of the string in.
+  it("renders `$&` inside inline code literally (no match splice / NUL leak)", () => {
+    const out = markdownToTelegramHtml("See `a$&b` here");
+    expect(out).toBe("See <code>a$&amp;b</code> here");
+    expect(out).not.toContain("\x00");
+    expect(out).not.toContain("IC0");
+  });
+
+  it("renders `$'` inside inline code literally (no rest-of-string splice)", () => {
+    const out = markdownToTelegramHtml("run `x$'y` end");
+    expect(out).toBe("run <code>x$'y</code> end");
+  });
+
+  it("renders `$$` inside inline code literally (no silent $ loss)", () => {
+    const out = markdownToTelegramHtml("cost `$$5` total");
+    expect(out).toBe("cost <code>$$5</code> total");
+  });
+
+  it("renders `$&` inside a fenced code block literally", () => {
+    const out = markdownToTelegramHtml("```\nawk '{print $&}'\n```");
+    expect(out).toContain("<pre><code>");
+    expect(out).toContain("$&");
+    expect(out).not.toContain("\x00");
+    expect(out).not.toContain("CB0");
+  });
+});
+
+describe("markdownToTelegramHtml — link href escaping (#265 review)", () => {
+  it("attribute-escapes a double-quote in the URL", () => {
+    const out = markdownToTelegramHtml('[x](https://a?q="hi")');
+    expect(out).toBe('<a href="https://a?q=&quot;hi&quot;">x</a>');
+  });
+
+  it("keeps `&` in a URL as the correct &amp; entity (Telegram decodes it back)", () => {
+    const out = markdownToTelegramHtml("[ex](https://x.com/a?b=1&c=2)");
+    expect(out).toBe('<a href="https://x.com/a?b=1&amp;c=2">ex</a>');
+  });
+});
+
+describe("isTelegramHtmlParseError — only true for malformed-HTML 400s", () => {
+  it("true for a 400 can't-parse-entities error", () => {
+    expect(
+      isTelegramHtmlParseError(
+        new Error("Telegram API sendMessage: 400 Bad Request: can't parse entities: unclosed tag"),
+      ),
+    ).toBe(true);
+  });
+
+  it("false for a benign 400 (message is not modified)", () => {
+    expect(
+      isTelegramHtmlParseError(
+        new Error("Telegram API editMessageText: 400 Bad Request: message is not modified"),
+      ),
+    ).toBe(false);
+  });
+
+  it("false for 'message to edit not found' and for a 429 flood", () => {
+    expect(
+      isTelegramHtmlParseError(
+        new Error("Telegram API editMessageText: 400 Bad Request: message to edit not found"),
+      ),
+    ).toBe(false);
+    expect(
+      isTelegramHtmlParseError(
+        new Error("Telegram API sendMessage: 429 flood cooldown, 5s remaining"),
+      ),
+    ).toBe(false);
+  });
+});
--- a/src/adapters/telegram/__tests__/telegram.test.ts
+++ b/src/adapters/telegram/__tests__/telegram.test.ts
@@ -633,13 +633,15 @@ describe("TelegramAdapter — response.text outbound", () => {
       agent_id: "triage",
       session_id: "s1",
       topic: "response.text",
-      payload: { text: "**bold** and `code` and [link](https://example.com)" },
+      payload: {
+        text: "**bold** and *italic* and `code` and [link](https://example.com)",
+      },
     });
     await waitFor(() => api.sendMessages.length > 0);
     const sent = api.sendMessages[0];
     expect(sent?.parse_mode).toBe("HTML");
     expect(sent?.text).toBe(
-      '<b>bold</b> and <code>code</code> and <a href="https://example.com">link</a>',
+      '<b>bold</b> and <i>italic</i> and <code>code</code> and <a href="https://example.com">link</a>',
     );
   });
 
--- a/src/adapters/telegram/__tests__/telegram.test.ts
+++ b/src/adapters/telegram/__tests__/telegram.test.ts
@@ -130,6 +130,7 @@ interface SendMessageCall {
   text: string;
   message_thread_id?: number;
   reply_markup?: { inline_keyboard: TelegramInlineKeyboardButton[][] };
+  parse_mode?: "HTML";
 }
 
 interface SetReactionCall {
@@ -151,7 +152,12 @@ interface AnswerCallbackCall {
  */
 class FakeTelegramApi implements TelegramApi {
   public readonly sendMessages: SendMessageCall[] = [];
-  public readonly editMessages: Array<{ chat_id: number; message_id: number; text: string }> = [];
+  public readonly editMessages: Array<{
+    chat_id: number;
+    message_id: number;
+    text: string;
+    parse_mode?: "HTML";
+  }> = [];
   public readonly reactions: SetReactionCall[] = [];
   public readonly callbackAcks: AnswerCallbackCall[] = [];
   /** #201: count inbound polls so send-only configs can assert zero. */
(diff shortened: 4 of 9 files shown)


Nibbler1250:
Thanks for this — the bus adapter genuinely needed markdown rendering, and the overall approach is right: a single `markdownToTelegramHtml()` as the shared source of truth, protecting code spans first, and a plain-text fallback for malformed HTML. I ran it through a thorough pass and everything below is **edge-hardening on the corpus this actually runs against (Claude's output, which hits these edges daily)** — not a design problem. A few are worth fixing before merge.

## Blocking

### 1. Replacement-string `$` corruption in the code-restore step · `format.ts` (steps 11–12)
The restore loops use the matched text as a **replacement string**, so `$`-sequences inside protected code get reinterpreted by `String.prototype.replace`:
```js
text = text.replace(`\x00IC${i}\x00`, `<code>${escaped}</code>`);
```
Any code containing `$&`, `$'`, `` $` ``, `$$`, or `$1` is mangled. Examples:
- `` `$$5` `` (a price) → `<code>$5</code>` — a `$` is eaten
- `` `a$&b` `` → re-injects the NUL sentinel into the output
- fires routinely on shell / `sed` / `awk` / regex / price snippets

**Fix:** use a function replacer so the value is taken literally (and it sidesteps the regex-vs-string ambiguity entirely):
```js
text = text.replace(`\x00IC${i}\x00`, () => `<code>${escaped}</code>`);
// same for the code-block (CB) loop
```
While here, stripping `\x00` from the input up front (`text = text.replace(/\x00/g, "")`) also closes a rare sentinel-collision if Claude ever echoes a literal `\x00IC<n>\x00`.

### 2. Quadratic backtracking (ReDoS) on the link regex · `format.ts` (step 6)
```js
/\[([^\]]+)\]\(([^)]+)\)/g
```
On a run of unmatched `[` (bracket-heavy echoed content — logs, JSON), this is O(n²) and blocks the single-threaded bus event loop *before* the API call. Measured on the branch: 40k `[` → ~1.6s, 80k → ~6.5s, which stalls the whole conversation + poll loop.

**Fix:** bound the classes and/or cap input length (Telegram messages are 4096 chars anyway):
```js
/\[([^\]\n]{1,256}?)\]\(([^)\s]{1,2048})\)/g
```
A `'['.repeat(50000)` regression test asserting sub-100ms would lock this in.

### 3. The spinner edit un-renders the HTML · `index.ts` `startSpinner()` (the `setInterval` body)
The periodic `editMessageText` wasn't routed through the new conversion — every ~2.5s it overwrites the freshly-rendered message with `${FRAME} ${baseText}` as **raw markdown, no `parse_mode`**. So on any `intent: "progress"` turn (the common case for agent replies) the user sees literal `**bold**` / `[x](url)` for the whole turn, self-healing only on the final edit — re-introducing exactly the bug this PR fixes, on the progress path.

**Fix:** route the spinner tick through `markdownToTelegramHtml(`${FRAME} ${baseText}`)` with `parse_mode: "HTML"` (the frame glyph survives conversion).

## Strongly recommended (the PR's headline safety feature is untested)
`FakeTelegramApi` never rejects, so the **400 → plain-text fallback** has no coverage. A small failure-injection hook (`failNextSend`/`failNextEdit` throwing a 400-shaped error) would let you assert: it lands once, the retry has **no** `parse_mode`, `message_thread_id` is preserved, and a double-fault is logged rather than thrown. A direct `describe()` over `markdownToTelegramHtml` (tag families, nested emphasis, unclosed/empty fences, `$`/`&`/`<>`-in-code) would also pin the conversion.

## Minor / nice-to-have
- **`href` isn't quote-escaped** (`format.ts` step 6): a `"` in a URL breaks the `<a href="">` tag → Telegram 400 → the message degrades to plain text. Escape the URL inside the same function replacer (`"`→`&quot;`, plus `& < >`); the link **text** is already escaped by step 5, so only the href needs it.
- **Bare `catch {}`** on the fallbacks (`index.ts`): it fires on *every* error, not just a 400. A transient that reaches Telegram after delivery could re-send and duplicate the message. Gating the fallback on the error actually carrying `400` (log-and-abandon otherwise) keeps it tight.

Really nice contribution — these are all small, localized fixes and the structure is solid. Happy to look again once they're in. 🙌

TerrysPOV (changes requested):
### Code review

Solid fix on the right surface — converter is clean and the test coverage is decent. Six issues, one blocker:

1. **Missing version bumps (CI will fail).** PR touches shipped plugin code under `src/` but `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` are unchanged. The `plugin-version-guard` and `marketplace-version-guard` CI checks will fail. Run `bun run bump:plugin-version` and `bun run bump:marketplace-version` before merge — the project root CLAUDE.md "Development Workflow" section calls this out explicitly.

2. URLs containing `&` are double-escaped because HTML escaping (step 5) runs *before* link extraction (step 6). Input `[ex](https://x.com/a?b=1&c=2)` produces `<a href="https://x.com/a?b=1&amp;c=2">` — the literal `&amp;` ends up in the href and breaks every multi-param URL. Fix: move link extraction *before* the HTML-escape step, or protect URLs with placeholders the same way code blocks already are.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/2bab5c78f7949438d5f0579413b68a03406e5763/src/adapters/telegram/format.ts#L37-L42

3. \`text.replace(placeholder, escapedCode)\` uses string replacement, so \`$&\` / \`$1\` / \`$$\` in code content are interpreted as back-references. A code block containing \`awk '{print $1}'\`, a regex example, or any \`$1\`-style template will silently lose those fragments. Fix: use a function replacer — \`text.replace(placeholder, () => \`<code>${escaped}</code>\`)\` — which receives the string literally.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/2bab5c78f7949438d5f0579413b68a03406e5763/src/adapters/telegram/format.ts#L60-L76

4. The legacy `commands/telegram.ts` path previously rendered italic only on \`_text_\`; now via shared \`format.ts\` it also renders \`*text*\` as italic. Probably a welcome consolidation, but a one-line note in the PR description would prevent surprises on the PTY path.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/2bab5c78f7949438d5f0579413b68a03406e5763/src/adapters/telegram/format.ts#L47-L52

5. \`editHtml\`'s catch retries on *any* error. PR #141 established that "message is not modified" and "message to edit not found" 400s should silently no-op, not retry — otherwise a duplicate plain-text edit fires (or the same edit fails twice and logs). Discriminate on the error description before the plain-text fallback.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/2bab5c78f7949438d5f0579413b68a03406e5763/src/adapters/telegram/index.ts#L543-L560

6. (Follow-up, not blocking.) \`safeSendMessage\` paths — permission-request, unauthorized notice, \`request_human\` prompt — still send hand-crafted strings containing raw Markdown (\`*Permission request*\`, backticks). Same UX issue this PR was filed to fix, just for system messages instead of agent replies. Worth a follow-up that routes those through \`sendHtml\` too.

TerrysPOV:
@Magister are you going to be able to finish-up your work on this PR? If not please say so and we (the maintainers) will pick it up and finalise.

Magister:
@TerrysPOV feel free to pick it up if you want - I realised that ClaudeClaw-Plus is not the ideal fit for my usecase so probably won't continue work on this at least for now

Nibbler1250 (commented):
Took a careful pass over the current head (`2bab5c7`) — @Magister nice work, the converter core is clean and HTML injection is handled correctly (verified `<script>`/`&`/`<b>` are fully escaped, bullets & lone `*` don't become italic, code preserves `<`/`&` literally, no catastrophic ReDoS). But the head is still the commit Terry reviewed, so his CHANGES_REQUESTED stands — and there's one blocker worse than originally framed. Empirically verified each below.

### Blockers

**1. [HIGH] `format.ts:66` & `:75` — code content corrupted by replacement back-references.**
`text.replace(placeholder, \`<code>${escaped}</code>\`)` — a *string* 2nd arg still interprets `$&`, `` $` ``, `$'`, `$$`. Verified:
- `` See `a$&b` here `` → `<code>a␀IC0␀amp;b</code>` — raw NUL + `amp;` leak into the delivered message.
- `` run `x$'y` end `` → `$'` splices the rest of the message body into the code span.
- `` cost `$$5` `` → `$5` (silent `$` loss). Same for fenced blocks at `:75`.

Fix — use a **function replacer** so the string is inserted literally:
```js
text.replace(placeholder, () => `<code>${escaped}</code>`);
```
(Terry's `$1` example happens to survive since there's no capture group, but `$&`/`$'`/`$$` make the class real.)

**2. [MED-HIGH] `index.ts` (~547, 583, 674) — the plaintext fallback catches EVERY error, not just a malformed-HTML 400.**
`sendHtml`/`editHtml` throw on any `!res.ok` (`api.ts:100`) incl. the 429 flood short-circuit. The bare `catch {}` then re-sends as plain text, so:
- a **"message is not modified"** 400 → resends *raw markdown*, downgrading an already-correct formatted message;
- a **"message to edit not found"** 400 → double-fails, regressing the #141 silent-no-op contract.
Fix: only fall back when the error text indicates a parse/entity error; re-throw / no-op otherwise.

**3. [MED] Version bumps missing → CI red.** Diff touches `src/` but not `.claude-plugin/plugin.json` / `marketplace.json`. Run `bun run bump:plugin-version && bun run bump:marketplace-version` and commit.

### Lower
- **[MED] overlapping emphasis** `**a _b** c_` → `<b>a <i>b</b> c</i>` → Telegram 400 → the whole message falls back to plain text (losing *all* formatting, not just the bad span). Hard to fix perfectly; worth a known-limitation note. Fixing #2 is what keeps this from silently dropping messages.
- **[LOW] `format.ts:41`** href isn't attribute-escaped for `"` — `[x](…?q="hi")` → invalid `href="…"hi""`. Low from Claude output; self-heals via the #2 fallback once that's tightened.

### One of the original points looks like a false positive
Terry's #2 (URL `&` "double-escaped") — I couldn't reproduce a defect: `[ex](https://x.com/a?b=1&c=2)` → `href="https://x.com/a?b=1&amp;c=2"`, which is the *correct* Telegram-HTML encoding (Telegram decodes it back to `&`, URL intact). Escaping runs once, not twice. I'd suggest **not** reordering the escape here — the suggested reorder would put a raw `&` in the attribute and actually break it. Worth a 30-sec check against live Telegram before touching it.

Net for @Magister: blockers are **#1 (function replacer)** and **#2 (discriminate the fallback)**, plus the version bumps for green CI. Fix those and I think this is close.

Nibbler1250:
@Magister I pushed a maintainer commit (`d5b7bc7`) folding in the review blockers so this isn't stuck on you — hope that's welcome, shout if you'd rather have driven it:

- **HIGH** — the code / code-block restore now uses **function replacers**, so `$&`, `$'`, `` $` `` and `$$` inside code render verbatim instead of splicing the placeholder / rest-of-message. (`` `a$&b` `` was becoming `a␀IC0␀amp;b`.) Same for the link replace.
- **MED-HIGH** — added `isTelegramHtmlParseError()` and gated all four fallback sites on it: only a genuine malformed-HTML 400 retries as plain text; a benign 400 ("message is not modified" / "to edit not found") or a 429 no longer downgrades an already-formatted message to raw markdown (was regressing #141). `api.ts` now surfaces Telegram's `description` so the discrimination works.
- **LOW** — link href is attribute-escaped (`"` → `&quot;`).
- **Version bumps** (plugin + marketplace) so the guard CI goes green.
- Added a format unit test (back-ref literalness, href escaping, parse-error discrimination). Telegram suite **62 pass / 0 fail**.

Left the overlapping-emphasis case (`**a _b** c_`) as a known limitation — the now-discriminated fallback ships it as plain text rather than dropping it. Terry's URL-`&` point looked like a false positive (the `&amp;` encoding is correct), so I didn't touch it. One thing: the branch is a fair way behind `main`, so it'll want a rebase before merge (that'll also reconcile the version).

TerrysPOV (approved):
Re-reviewed after Nibbler's maintainer commit (`d5b7bc7`) — approving (supersedes my earlier change-request). All prior blockers are resolved and verified against the new tests (62 pass, 0 fail):

- **URL escaping (#2):** link extraction now uses a function replacer with `"`→`&quot;`; `&` in a URL becomes the correct `&amp;` entity (Telegram decodes it back). Pinned by a test.
- **Code back-references (#3):** both inline-code and code-block restores use function replacers, so `$&` / `$1` / `$$` in code (e.g. `awk '{print $&}'`) survive intact.
- **editHtml fallback (#5):** `isTelegramHtmlParseError` discriminates genuine HTML-parse 400s from benign ones ('message is not modified' / 'to edit not found' / 429 flood), so a well-formed message is no longer downgraded to raw markdown on a benign error. `api.ts` now surfaces the Telegram `description` into the thrown message so the matcher works.
- **Italic consolidation (#4):** documented in the PR body.
- **System-message markdown (#6):** acknowledged follow-up, out of scope here.

Maintainer commit from me: rebased onto current main (no source conflicts — telegram files rebased cleanly; only the version files needed re-resolving) and bumped to 2.2.185 so the version guards pass. biome + build + telegram suite all green. Good to merge.

TerrysPOV (approved):
Approved following @nibbler changes and my maintianer commit.

PR #271, opened 2026-06-23
fix(bus): match the REPL footer marker whitespace-tolerant

Nibbler1250:
## What this changes

The bus PTY agent detects the REPL-ready footer via literal substrings:

- `recentOut.includes("to cycle")` — the delivery-confirm re-nudge loop
- `buf.includes("tab to cycle")` — the boot-ready watcher

The CLI renders that status line with **CHA cursor positioning** (`ESC[NG`
between tokens), not literal spaces. After the ANSI stripper runs, the markers
collapse to `tocycle` / `tabtocycle`, so the spaced literals never match.

Captured by driving the real CLI through the bus's own spawn args + ANSI
stripper:

```
[6Gbypass[13Gpermissions[25Gon[28G(shift+tab[39Gto[42Gcycle)
→ stripped: "...permissionson(shift+tabtocycle)..."
```

## Effect

With the marker never matching:

- **delivery-confirm**: `!recentOut.includes("to cycle")` is always true, so any
  post-CR window that carries output reads as `turn-started` — the re-nudge path
  effectively never fires.
- **boot-ready**: the watcher only disengages on its timeout, never on the
  footer marker.

## Honest scope

This is a **robustness/correctness fix, not a fix for an observed incident.** I
found it while debugging an unrelated outage (an expired OAuth token, since
resolved) — the symptoms looked like submit wedges but were auth (401). The
matcher mismatch is real and captured, but I have **no production wedge
attributable to it**: the net it restores guards the documented paste/redraw
race, which I did not reproduce in the wild. Submitting purely because the
matcher objectively cannot match the current render, and the surrounding
comments already intend it to survive footer rewrites.

## Fix

Whitespace-tolerant matchers, each keeping its original guard:

| matcher | before | after |
|---|---|---|
| delivery-confirm | `includes("to cycle")` | `/to\s*cycle/` (bare-core, unchanged intent) |
| boot-ready | `includes("tab to cycle")` | `/tab\s*to\s*cycle/` (keeps the stricter `tab` prefix that avoids tripping on model prose like "to cycle permission modes") |

`\s*` matches both the spaced render and the CHA-collapsed render, on every
CLI version.

## Scope / risk

- 2 lines changed in `src/bus/session-agent-process.ts`. No happy-path change
  (a streaming turn has no footer → still detected as started); the change only
  restores correct detection of the idle/failed case.
- Plugin + marketplace versions bumped (Version Guard).

## Testing

- Reproduced the collapse via a `bun-pty` harness running the real CLI with the
  bus's spawn args + stripper: the stripped footer is `tocycle` (no space), so
  `includes("to cycle")` returns false at an idle REPL.
- `biome check` clean on the changed file.
- `bun build src/index.ts` (smoke / type check) clean — 378 modules, no errors.
- `bun test src/bus/__tests__/session-agent-process.test.ts` → 18 pass / 0 fail
  (incl. the boot-footer disengage test that exercises this matcher).
- `bun test src/__tests__/watchdog.test.ts` (the deploy gate) → 17 pass / 0 fail.
- No regressions; version guards satisfied (plugin + marketplace bumped).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.155",
+      "version": "2.2.156",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.156",
+      "version": "2.2.157",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.156",
+  "version": "2.2.157",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.155",
+  "version": "2.2.156",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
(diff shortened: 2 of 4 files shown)


TerrysPOV (approved):
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

The whitespace-tolerant matchers are correct and well-targeted:

- `/to\s*cycle/` and `/tab\s*to\s*cycle/` both match the spaced render AND the CHA-collapsed render (`\s*` = zero-or-more), so detection is restored on the current CLI without losing the old behaviour.
- Case sensitivity preserved (no `i` flag, matching the original `includes()`).
- The boot-ready matcher keeps the strict `tab` prefix, so the #195 guard against tripping on boot prose ("to cycle permission modes") stays intact.
- No ReDoS risk — single `\s*` between literals is linear.
- History confirms the "robustness fix, not incident" framing: both markers were correct at introduction (#195, #248); the CHA-collapse is a later CLI render change. Prior review intent on #242/#248/#250 supports this approach.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/20b2bc3866b76395303054ccfdd5fa987ac5ea6d/src/bus/session-agent-process.ts#L284-L287

Two non-blocking notes:

1. No automated test exercises the real ANSI-stripped CLI render end-to-end — the existing tests inject fake footer strings, and the collapsed-output sample was captured manually. A `bun-pty` harness test that asserts `/tab\s*to\s*cycle/` matches the actual stripped footer would lock this against future CLI render drift.
2. Version sequencing (merge-time, not a code issue): this bumps to 2.2.156 from a 2.2.155 merge-base, but `main` already advanced to 2.2.156 via #270. The CI guard passes (it compares against merge-base), but the shipped version won't actually advance, and it collides with #272 which also targets 2.2.157. Whichever of #271/#272 merges second will need a re-bump.

PR #273, opened 2026-06-28
fix(governance): getBudgetHealth returns scoped policies (empty query = all)

Nibbler1250:
## Summary

`getBudgetHealth()` silently dropped every **scoped** budget policy — channel/user/model/session/provider-scoped budgets never appeared in the health view; only global (unscoped) policies surfaced.

## Root cause

`getBudgetHealth()` calls `getBudgetState({})` (empty scope = "list every policy"). `getBudgetState` filters each policy through `matchesScope(query, policy.scope)`. With an empty query, `matchesScope` rejects any policy that *has* a scope field — there is nothing in the empty query to match the scoped field against — so scoped policies were excluded from the result set.

## Fix

Treat an empty query scope as "select all" in `getBudgetState`. A non-empty query still filters exactly as before.

- **Enforcement is unchanged** — the evaluate/check path always passes a concrete (non-empty) call scope, which still filters correctly. Only the **health/dashboard listing** is corrected.
- One root cause, one scoped change.

## Tests

- Restores the previously-failing `Telemetry > should return budget health` test (creates a channel-scoped policy, asserts it appears in `getBudgetHealth()`).
- Full governance suite green; no new lint/type diagnostics introduced.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,8 +7,8 @@
     {
       "name": "claudeclaw-plus",
       "source": "./",
-      "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.157",
+      "description": "ClaudeClaw+ \u2014 governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
+      "version": "2.2.158",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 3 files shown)


Nibbler1250:
Follow-up from a self-review pass on this change:

- The empty-scope short-circuit now covers **all six** `BudgetScope` fields (added `userId`), so a future userId-only query is handled consistently with the others rather than being misread as "empty → select all".

- Separately, and **pre-existing / out of scope here**: in `client.ts`, the `getBudgetState(channelId?: string)` wrapper forwards a raw string into `getBudgetState(scope: BudgetScope)`. A string carries none of the scope fields, so that wrapper never actually filters by channel. It is currently unreferenced (no `src` callers), so nothing depends on it today. Happy to send a small separate PR tightening the signature to take a `BudgetScope` (with an undefined guard) if that is useful.

Issue #258, opened 2026-06-16
Governance: built-but-not-wired features + context-threading gap (follow-up to #257)


Nibbler1250:
## Governance layer: built-but-not-wired features + context-threading gap

While auditing the governance layer (policy engine, audit log, budget engine, watchdog, HITL approval) I confirmed several components that exist as well-tested code but are **never invoked on the production request path**, plus one root-cause gap that limits the features that *are* wired. The straightforward wiring has been done in a companion PR; this issue tracks the remaining items that need a design decision or new input plumbing rather than a one-line call.

### Already fixed (companion PR)
- **Budget enforcement** now receives real token usage + cost. `recordInvocationCompletion` was always called with `undefined` usage; added `extractTranscriptUsage()` (reads the CLI session JSONL, path-agnostic across the legacy stream-json and PTY paths) → `calculateEstimatedCost` → `recordInvocationCompletion`.
- **Scoped (per-channel / per-user) policy rules** now participate in `evaluate()` via `mergeScopedPolicies(loadedRules, getScopedRules(request))` in `getApplicableRules`. They previously loaded from disk but were never merged into evaluation.
- **Audit log is now tamper-evident** (hash chain: `prevHash` + `sha256(prevHash + content)` per entry, serialized writes, `verifyChain()`, retention re-seals survivors).
- Correctness quick-wins: real `toolName`/`toolArgs` passed to `checkToolApproval`; `getBudgetState` derives its usage filter from `policy.scope`; token/cost reads clamped against negative/NaN/Infinity.

### Remaining — need a decision or new input, not just wiring

**1. Policy context never carries `channelId` / `userId` / `skillName` (root cause).**
`getPolicyContext()` in `src/runner.ts:1491` hardcodes:
```ts
channelId: undefined, // Will be populated from event context
userId: undefined,
skillName: undefined,
```
(same at `:1628`, `:1780`). The TODO was never done. Consequence: the scoped-policy wiring above is structurally correct but **channel/user-scoped rules can never match in the runner path** because the request context that reaches them has these fields empty. The values exist on the inbound event upstream of the runner — they need to be threaded down into `getPolicyContext`. This is the highest-leverage fix and unblocks item 2.

**2. Skill-overlay deny rules have zero production callers.**
[...]

PR #277, opened 2026-06-28
fix(governance): thread channelId + userId + skillName into runner context (#258 item 1, live budget/usage)

Nibbler1250:
## What

Threads the inbound identity context — **`channelId`, `userId`, and `skillName`** — down into the runner's governance call-sites so per-channel and per-user budget/policy rules can actually match, and so usage records carry channel/user/skill attribution. This closes **#258 item 1** (the root-cause context-threading gap).

## Why

#258 item 1 identified that `getPolicyContext()` and the two live governance objects in `execClaude` hardcode `channelId/userId/skillName: undefined`:

- `selectModel({ …, channelId, userId })` (model router) → `matchesScope` in the budget engine could never match a channel- **or** user-scoped policy → **scoped budget / model-routing enforcement was structurally inert on the runner path**.
- `recordInvocationStart({ … })` → usage records carried no channel/user/skill attribution, forcing the budget engine into broad (over-counting, fail-safe) aggregation for scoped policies.

The values exist on the inbound surface event; this PR threads them through.

## Changes

- **`runner.ts`** — new `PolicyIdentity { userId?, skillName? }` threaded `runUserMessage → run → execClaude` (appended optional param, existing callers unaffected). `channelId` comes from the already-in-scope `threadId`. Populated at: the `selectModel` request (`channelId`, `userId`), the `recordInvocationStart` record (`channelId`, `userId`, `skillName`), and `getPolicyContext` (all three) so the documented helper is complete when its consumer is wired (item 2).
- **`governance/budget-engine.ts`** — `matchesScope` now evaluates `scope.userId` (the field existed on `BudgetScope` but was never checked), mirroring the existing `channelId` check, so **per-user budgets participate once `userId` is threaded**.
- **`governance/usage-tracker.ts`** — `InvocationContext`/`InvocationUsageRecord` carry `userId` + `skillName`; `recordInvocationStart` persists them (per-user spend + skill-level cost attribution).
- **Surfaces** — `telegram` (`message.from?.id`) and `slack` (`event.user`) pass the inbound user id; for slash commands that resolve to a skill, they pass the resolved skill name. `skillName` is empty for ordinary chat — exactly the cases skill overlays (item 2) target.

## Scope boundary

`skillName` is **threaded and persisted** here, but its policy **consumer** (skill-overlay deny rules, `getPolicyContext`'s tool path) is intentionally left to **item 2**, which the issue flags as a design decision ("where in the path should skill overlays be consulted"). This PR provides the `skillName` source that item 2 is blocked on.

## Verification

- `bun test src/__tests__/governance src/__tests__/policy` → all green.
- **Full suite: identical failure set before/after** — the change introduces **0 new failures** (the repo's pre-existing failures in escalation / pause-controller / event-processor / mcp-multiplexer / job-loading are untouched). +1 net pass from the new test.
- Regression guards added: a channel-scoped **and** a user-scoped budget policy each participate only when the respective id is threaded; the invocation record round-trips `userId` + `skillName`.

## Compliance / governance note

This restores the intended **per-channel and per-user budget ceilings and policy scoping** on the runner path — a channel/tenant/user can no longer silently bypass a scoped spend or routing rule because the request context arrived empty.

Completes the **live budget/usage** half of #258 item 1 (channelId/userId now scope model-routing + spend). The policy-path half via getPolicyContext stays inert until a caller exists — the per-tool gate is #284.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.163",
+      "version": "2.2.164",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.159",
+      "version": "2.2.160",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,8 +7,8 @@
     {
       "name": "claudeclaw-plus",
       "source": "./",
-      "description": "ClaudeClaw+ \u2014 governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.158",
+      "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
+      "version": "2.2.159",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.164",
+      "version": "2.2.165",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.164",
+  "version": "2.2.165",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.158",
-  "description": "ClaudeClaw+ \u2014 governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
+  "version": "2.2.159",
+  "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.163",
+  "version": "2.2.164",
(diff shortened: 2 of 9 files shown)


TerrysPOV:
**[COMPLEX — triage]**

Vision: ALIGNED — completes the #258 governance scoping wiring.

Overlapping PRs: #282 (item 2) builds on this and shares `runner.ts` / `budget-engine.ts` / `usage-tracker.ts` / `slack.ts` / `telegram.ts` — merge this first (#282 is author-noted as dependent).

> Housekeeping: the version-file merge conflict was resolved by a maintainer merge+bump commit (`5917e12`, v2.2.163). No action needed there.

### Code review

Found 3 issues:

1. **HIGH — threading `channelId` activates a latent fail-open in budget spend-scoping.** `evaluateBudget` (the live model-router path) builds its usage filter with `channelId: context.channelId` for **every** matched policy, regardless of the policy's scope, and has no `userId` filter (`UsageFilters` has none). Pre-PR, `selectModel` passed `channelId: undefined`, so spend aggregated broadly (fail-safe). This PR passes `channelId: threadId`, so (a) source-scoped / global / monthly budgets now count only the current channel → under-count → the cap trips late or never (fail-open); (b) per-user policies *participate* (the new `matchesScope` userId branch) but spend is keyed by channel, not user → per-user budgets don't actually meter per-user. Fix: derive `evaluateBudget`'s filters from the matched `policy.scope` (only filter `channelId` when channel-scoped; add `userId` to `UsageFilters`/`getAggregates`).

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5917e125f2221dc39fe64645c389d42bfc056abc/src/governance/budget-engine.ts#L484-L496

2. **Stale comment + half-done TODO.** This block still states "recordInvocationStart currently persists channelId/userId as undefined" — no longer true after this PR. Update it, and either complete the "re-enable filtering" TODO (see issue 1) or explicitly defer it.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5917e125f2221dc39fe64645c389d42bfc056abc/src/governance/budget-engine.ts#L603-L613

3. **`getPolicyContext` threading is dead code — "closes #258 item 1" is overstated.** `getPolicyContext` has no caller in `src/`, so threading `channelId`/`userId`/`skillName` into it has no runtime effect; only the budget (`selectModel`) and usage (`recordInvocationStart`) paths are live. The policy half stays inert until something calls it.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/5917e125f2221dc39fe64645c389d42bfc056abc/src/runner.ts#L1491-L1495

Nibbler1250:
Thanks — sharp catch on the spend-side fail-open. Fixed in `1459d7e`.

**Finding 1 (HIGH).** `evaluateBudget` now derives its usage filter from the matched `policy.scope` instead of the request context (mirroring `getBudgetState`): a dimension is only constrained when the policy actually scopes on it, so global / source-scoped / monthly budgets aggregate across all channels again. Multi-value (`string[]`) and unscoped dimensions aggregate broadly (over-count = fail-safe). `UsageFilters` + `getAggregates` gained a `userId` filter so per-user budgets meter per-user.

**Finding 2.** Same area: since channelId/userId are now persisted per usage record (item 1), I re-enabled channelId + userId filtering in `getBudgetState` and refreshed the stale "producer persists undefined" comment.

New tests: a global budget aggregates spend across all channels (the fixed fail-open); a channel-scoped budget meters only its channel; a user-scoped budget meters only its user. Full suite: identical pre-existing failure set.

**Finding 3.** Agreed — `getPolicyContext` has no caller, so its channelId/userId/skillName threading is forward-plumbing with no runtime effect today; only the budget (`selectModel`) and usage (`recordInvocationStart`) paths are live. I'll soften the "closes item 1" framing to "completes the live budget/usage threading"; the policy half stays inert until a caller exists (that's the #258 item-3 gate work, which I've started separately in #284).

TerrysPOV (approved):
Re-reviewed the fixes — the HIGH finding is resolved.

`evaluateBudget` now derives its usage filter from `policy.scope` rather than the request: global / source / monthly budgets aggregate broadly again (fail-open closed), and `channelId`/`userId` are constrained only when the policy actually scopes on them. `getAggregates`/`UsageFilters` gained a real `userId` filter (per-user budgets now meter per-user), `getBudgetState` was made consistent, and the stale "producer persists undefined" comment is corrected. The new tests assert the global / per-channel / per-user metering. Governance suites green (30 pass).

Approving. Per the dependency, merge this before #282 (the item-1 base).

Non-blocking follow-up (fine to defer): `getPolicyContext` is still uncalled, so the policy-context half stays inert until the item-2/3 gate — the live budget/usage paths are correct.

PR #282, opened 2026-06-28
feat(governance): consult skill-overlay deny rules in policy engine (#258 item 2)

Nibbler1250:
## What

Wires `src/policy/skill-overlays.ts` into the policy evaluation path, giving it its **first production caller**. A skill's `deniedTools` overlay now participates in `evaluate()`. This is **#258 item 2**.

> **Depends on #277 (item 1)** for the `skillName` threading. Until #277 merges, this PR's diff also contains item-1's commits; once #277 lands on `main`, only the item-2 changes remain. Please merge #277 first.

## Why

#258 item 2: *"`skill-overlays.ts` is fully implemented and tested but nothing in `src/` calls it. It also keys on `skillName`, which is unavailable until item 1."* Item 1 (#277) threads `skillName`; this connects the consumer.

## Changes

- **`policy/engine.ts`** — `getApplicableRules` folds the active skill's overlay deny rules into the candidate set, alongside the scoped per-channel/per-user rules it already merges. Overlays are **deny-only and scoped by `skillName`**, so they can only ADD a restriction for the matching skill — never widen access. (Import is evaluate-time only; `skill-overlays` imports only *types* from `engine`, so no init cycle — same discipline as the existing `channel-policies`↔`engine` note.)
- **`policy/skill-overlays.ts`** — sync overlay-rules cache (`cacheSkillOverlayFromContent` / `getCachedSkillOverlayRules` / `clearSkillOverlayRulesCache`). `getApplicableRules` is sync while `SKILL.md` reads are async, so a surface parses the overlay from the content it already loaded and caches the derived rules under `skillName` for sync lookup. Mirrors the `channel-policies` cache pattern.
- **`commands/telegram.ts`, `commands/slack.ts`** — populate the cache when a slash command resolves to a skill (the `SKILL.md` content is already in hand).
- **`gateway/index.ts`** — derive `skillName` from the inbound bot command for the policy request.

## Tests

`evaluate()` now: denies a tool the active skill forbids (overlay outranks a base allow), still allows undenied tools, and never leaks one skill's overlay to a different skill or to a skill-less request. Full suite: **identical pre-existing failure set, +4 net pass**.

## Scope boundary (honest)

This connects the **consumer** — the engine consults overlays wherever `evaluate()` runs with a `skillName`. It does **not** yet add a live **per-tool-call gate**: the runner's `evaluateToolForExecution` is still inert, and the gateway gates *message admission* (`toolName: "InboundMessage"`), not individual tool calls. So overlay **enforcement** is wired-but-dormant until that gate exists — issue #258 item 3 ("hook point inside the PTY/stream loop"). This PR is the prerequisite that unblocks it.

Refs #258 (item 2).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.159",
+      "version": "2.2.160",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.165",
+      "version": "2.2.166",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.161",
+      "version": "2.2.164",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.160",
+      "version": "2.2.161",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,8 +7,8 @@
     {
       "name": "claudeclaw-plus",
       "source": "./",
-      "description": "ClaudeClaw+ \u2014 governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.158",
+      "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
+      "version": "2.2.159",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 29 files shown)


TerrysPOV:
**[COMPLEX — triage]**

Vision: ALIGNED — wires skill-overlay deny rules into the policy engine per #258 item 2.

Overlapping PRs: contains #277 (item 1) — merge #277 first (author-noted). Shares the item-1 files.

> Housekeeping: the version-file merge conflict was resolved by a maintainer merge+bump commit (`ea57fea`, v2.2.164). No action needed there.

### Code review

The engine wiring is correct (deny-only, `skillName`-scoped, precedence sound) and faithful to item 2. Found 3 issues:

1. **Overlay deny is runtime-inert (author-disclosed) — tests are false-green.** No live path supplies both a real `toolName` and a `skillName` to `evaluate()` in one request: the gateway evaluates the pseudo-tool `"InboundMessage"`, so a tool-scoped deny (e.g. deny `Bash`) can never match; and the only real-tool path (`evaluateToolForExecution`) has no caller. The 4 new tests call `evaluate({skillName, toolName:"Bash"})` directly — a request no production caller builds — so the suite implies working enforcement that doesn't fire. The PR body discloses this ("dormant until item 3"), so it's accepted scope — but please add one integration test through a real inbound/executor path (or relabel as engine-level participation) so the suite doesn't read as end-to-end.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/ea57fea022d5470ab08c76d905b4eeb356bb174c/src/gateway/index.ts#L240-L244

2. **P2 — Discord surface omitted from overlay-cache population.** `telegram.ts` and `slack.ts` populate the overlay cache + thread `skillName`; `discord.ts` (which also resolves skills here) does neither. Since `overlayRulesCache` is a process-global `Map`, once item-3's gate lands a Discord-invoked skill's deny silently won't apply (nondeterministic, depends on cache warmth from another surface). Mirror the telegram/slack change in `discord.ts`.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/ea57fea022d5470ab08c76d905b4eeb356bb174c/src/commands/discord.ts#L1156-L1160

3. **P3 — overlay deny precedence is not absolute; document it.** Overlay deny rules sit at priority 150 (`basePriority + 50`); a higher-priority (or equal-priority, more-specific) `allow` outranks them via `sortRules`, so the module-header claim "enforced even when globally allowed" isn't strictly guaranteed. #258 itself raised this. A one-line note that overlay denies are priority-150 (not absolute) closes it.

https://github.com/TerrysPOV/ClaudeClaw-Plus/blob/ea57fea022d5470ab08c76d905b4eeb356bb174c/src/policy/engine.ts#L422-L428

Nibbler1250:
Fixed the Discord gap (P2) in `000a7d2` — `discord.ts` now populates the overlay cache on skill resolution, mirroring telegram/slack, so a Discord-invoked skill's deny is no longer dependent on another surface having warmed the process-global cache.

On the other two: **(1)** agreed the 4 tests assert engine-level participation, not end-to-end enforcement (the request shape no live caller builds yet) — I've started the live per-tool gate that closes this in #284, and will add the end-to-end test once the gate threads skillName; happy to relabel these as "engine participation" in the meantime. **(3)** will add the one-line note that overlay denies sit at priority 150 (not absolute — a higher/equal-priority allow outranks them). Both tracked.

TerrysPOV (approved):
Re-reviewed the fixes — all three flagged items addressed.

- Discord parity: `discord.ts` now calls `cacheSkillOverlayFromContent` after skill resolution, with the same slash-stripped key as telegram/slack — closes the process-global cache gap.
- Deny precedence: documented as priority-150 (`basePriority + 50`), not absolute (an equal/higher-priority allow still outranks via `sortRules`).
- Tests relabeled engine-level, matching the disclosed scope.

The overlay remains correctly wired-but-dormant until the item-3 per-tool gate. Policy suite green (28 pass).

Approving. Merge after #277 (the item-1 dependency).

PR #284, opened 2026-06-28
feat(governance): policy gate on forwarded tool permissions (#258 item 3, slice 1)

Nibbler1250:
## What

Wires the policy engine into the **bus permission path** so an operator-authored **deny** rule auto-rejects a tool request *before* it reaches the human-in-the-loop card. This is **#258 item 3, slice 1** — the first real consumer of a per-tool policy gate.

Independent of #277/#282 (it uses only `evaluate()` + explicit rules, already on `main`).

## Why

#258 items 2/3: skill-overlay deny rules (and scoped/global denies) had no live enforcement point — the runner's `evaluateToolForExecution` is inert and the gateway gates message *admission*, not individual tool calls. The bus permission flow (`permission_request` → operator card) is the one place a tool is gated *before it runs*. This injects policy there.

## Changes (`src/bus/core.ts`)

- The `permission_request` handler consults the policy engine before publishing the operator card. **deny-wins**: an EXPLICIT matched deny rule (global, scoped, or — once `skillName` reaches this seam — a skill overlay) short-circuits to a deny response.
- **Critical safety**: the engine's no-match **default-deny** (no `matchedRuleId`) is deliberately NOT treated as an auto-deny — otherwise every tool would be blocked when no rules are configured. Only an operator-authored deny acts.
- **Fail-OPEN**: any evaluation error falls through to the normal operator card, so a broken policy layer can never wedge an agent's tools.
- The policy evaluator is **injected via `BusCoreOptions`** (defaults to the engine singleton `evaluate`), mirroring the existing `eventLogAppend` DI — keeps the gate deterministically testable without the global rule file.

## Scope boundaries (honest)

- **Minimal slice**: gates the FORWARDED (default/plan/auto-mode) permission requests that reach bus core. `skillName`/`userId` are not yet threaded to this seam, so **skill-overlay and per-user rules engage in a follow-up slice** once that context is tracked per agent.
- Agents on the legacy `claude -p` path (e.g. `pty.enabled: false`) don't route permissions through bus core — wiring that spawn to a `--permission-prompt-tool` is a separate phase.

## Tests

Explicit deny → auto-deny + no card; no-match default-deny → falls through to card; evaluator throws → fails open to card. Full suite: **identical pre-existing failure set, +3 net pass.**

Refs #258 (item 3).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.167",
+      "version": "2.2.168",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.166",
+      "version": "2.2.167",
       "keywords": [
         "cron",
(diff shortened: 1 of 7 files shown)


TerrysPOV:
@claude

Issue #275, opened 2026-06-28
RFC: a governed self-tuning layer for Claude Code config — offering to upstream (extends #116)


Nibbler1250:
Hi Terry — following on from #116 (the WiseCron / TunableSubject scaffold), I've had a fuller version of this running locally for a while and it has grown into a complete system. Before I invest in polishing it for upstream, I wanted to open a channel and ask: **would you want this in ClaudeClaw-Plus, and if so, how would you want it scoped and staged?** Your repo, your call — I'd rather shape it to your direction than drop a large PR on you.

### What it is

A **governed self-tuning layer** for the Claude Code surface. The core abstraction is a `TunableSubject` — a config surface the system can safely improve over time. Subjects implemented: model-routing, `CLAUDE.md`, hooks, MCP plugins, prompt templates, **memory**, agents, and the scheduler.

Each subject runs a closed **outcome loop**:

> detect → propose → **human gate** → apply → **measure fitness** (telemetry) → **auto-revert on regression**

with the properties that matter in a governed setting:

- **Tamper-evident audit chain** — every proposal, apply, measurement and revert is recorded; you can read "this change → measurably improved X" straight from the chain.
- **Scope control** — `all` vs `agent`-scoped, surfaced explicitly (no silent blast radius), recorded in the audit.
- **Anti-Goodhart guardrails** — fitness is measured against guardrail metrics, not just the target.
- **MCP-native** — measurement IN (`telemetry__*`) and proposal lifecycle OUT (`tuner__*`) over the plugin MCP bridge, auditable and consumable like any other plugin. Mutating actions stay behind the human gate, never agent self-service.

### Why now — the memory + context-editing angle

Anthropic has just productized **file-based memory (the memory tool / native auto-memory) and context editing** (clearing tool results instead of lossy summarization). The field is converging on the same idea: *manage config and memory quality over time, with governance*, rather than letting it accrete. That's precisely what this layer does — it already includes a `MemorySubject` that keeps the memory files healthy. As operators lean on native memory/context-editing, a governed layer that keeps those surfaces tuned — with consent, provenance, and an audit trail — becomes more relevant, not less. That governance/provenance framing is the part I'd expect to matter most for regulated-finance users.

### The self-improvement loop

[...]

PR #286, opened 2026-06-29
feat(telemetry): host telemetry over MCP — read-only measurement surface (telemetry__*) [#275 brick 1]

Nibbler1250:
**Draft — the read-only measurement brick from the #275 staging (extends #116).**

Per your #275 note ("host-telemetry over MCP first — read-only, non-mutating, builds on the telemetry already in main, a clean easy-to-review first PR"), this is exactly that brick. No OutcomeLoop, no mutating `tuner__*` gate — those are the next PR (model-routing subject first).

## What it adds
A telemetry surface on the existing `PluginMcpBridge`:
- `telemetry__capabilities` + `telemetry__query` (host side, `registerHostTelemetryTools`)
- `McpTelemetryProvider` — the tuner-side **consumer, bundled in the same diff** (talks MCP only; in-process via `bridgeToolCaller` or over a real transport). Graceful degradation: an unreachable endpoint → empty capabilities/results, never a throw.
- Every served query appends a `telemetry_query` record to the tamper-evident `AuditLog` — provenance for each measured number.

Builds on what main already ships: `governance/telemetry.ts`, `orchestrator/telemetry.ts`, `plugins/mcp-bridge.ts`.

## Coverage — one stream per tunable subject
So every subject can measure fitness (the whole point of the loop that comes next):

| stream | feeds subject |
|---|---|
| `mode_dispatch` | model-routing *(the loop's first subject)* |
| `tool_call` | mcp-plugin |
| `memory_access` | memory |
| `hook_exec` | hooks |
| `cron_run` + `session_cost` | scheduler/cron |
| `template_feedback` | prompt-template |
| `agent_dispatch` | agents |
| `mcp.tool_call` | universal per-plugin gateway hub (auto-discovery) |
| `skill_access` | skill usage |

Each reference producer owns one stream, reads a single real source, and advertises `available:false` + reason when the source is absent — **declared-but-inactive by design, never faked data.**

## Scope / what's deliberately NOT here
- The `OutcomeLoop` and the **mutating `tuner__*` gate** (`serveTunerOverMcp`) — next brick.
- The end-to-end-over-real-transport test ships with that brick; the in-process bridge tests here cover the host + client contract.

## Tests
55 pass — host register, in-process client round-trip, graceful degradation, audit provenance/chain, per-producer derivation + availability. Fixtures use now-relative windows so the `capabilities()` 30-day wall-clock probe never drifts.

Marking draft for your read on shape/placement before I polish (notably: these land under `tuner/wisecron/` + `skills-tuner/core/` matching the current tree — happy to relocate if you'd prefer a different home as part of the skills-tuner→wisecron consolidation we discussed in #275).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.176",
+      "version": "2.2.177",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.175",
+      "version": "2.2.176",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.176",
+  "version": "2.2.177",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.175",
+  "version": "2.2.176",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/governance/mode-dispatch-journal.ts
+++ b/src/governance/mode-dispatch-journal.ts
@@ -1,0 +1,104 @@
+/**
+ * mode_dispatch instrumentation — the HOST-side producer of the `mode_dispatch`
+ * telemetry stream (OutcomeLoop Phase B).
+ *
+ * ClaudeClaw routes each request to an `agentic.modes` mode by keyword/phrase
+ * matching (`classifyTask`/`selectModel` in `../model-router.ts`, called from
+ * `./model-router.ts`'s `selectModel`). That resolution IS the dispatch event.
+ * This module records one `{ ts, mode, matched_keyword, reclassified }` line per
+ * dispatch to a journal that `ModeDispatchTelemetryProducer` reads.
+ *
+ * Sink indirection (why it's not a bare `appendFileSync`): the classifier runs
+ * inside unit tests and the `tuner` CLI, neither of which should touch the real
+ * home dir. So emission goes through a module-level sink that defaults to a
+ * NO-OP. The live daemon installs the file sink exactly once at router init
+ * (`ensureGovernanceRouter` in `runner.ts`) via `setModeDispatchSink`. Tests
+ * install an in-memory sink (or a temp-file one) explicitly. No env-sniffing,
+ * no accidental writes.
+ *
+ * `reclassified` is the metric the consumer needs (ModelRoutingSubject's
+ * `routing_reclassify_rate = nonzeroRate`). A fresh dispatch is never known to
+ * be a mis-route at emit time, so it is recorded `false` (value 0) — an honest
+ * baseline that activates the stream with a real dispatch count and a
+ * reclassify-rate of 0, never a fabricated nonzero. A later corrector can append
+ * a `reclassified:true` record for the same dispatch when one is detected.
+ */
+
+import { appendFileSync, mkdirSync } from "node:fs";
+import { homedir } from "node:os";
+import { dirname, join } from "node:path";
+
+/** One dispatch record. Field names match the on-disk JSONL + the producer. */
+export interface ModeDispatchEvent {
+  /** ISO-8601 UTC. Filled by `recordModeDispatch` when omitted. */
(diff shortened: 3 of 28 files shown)


Nibbler1250:
Thanks Copilot — all 5 addressed in 9d53001:
1. `TunerViewProvider.query()` now matches the `TelemetryProvider` signature (stream, range, filters?).
2. Dropped the developer-specific `simon-memory` marker from `MEMORY_PATH_RE` — the generic `CLAUDE.md` / `/memory/` / `/learnings/` markers are portable; fixtures updated.
3. Refreshed the stale doc: `agent_dispatch` IS emitted by `SessionJsonlTelemetryProducer`.
4/5. Removed the private `~/simon-memory/...` path refs from both module headers.

TerrysPOV (dismissed):
Hi mate. Went through this one properly. The shape is right and the read-only safety story holds, so no concerns on the architecture. The injected-runner hermeticity and one-producer-per-stream layout are exactly how I'd want this built. There are a few things I want tightened before it goes in though, mostly resource safety, which matters more here than usual given we're aiming at low-spec boxes.

Two I'd treat as blockers:

**1. Unbounded file reads.** `session-jsonl-provider.ts` walks every `.jsonl` under `~/.claude/projects` with no file-count cap and reads each one whole with `readFileSync` (line 187). On an active install that's potentially hundreds of files, some tens of MB, read synchronously inside the daemon. A 30-day capability probe alone can kick off a pile of those. On a 4GB box that's a genuine event-loop stall or OOM on the cold call. Same pattern in `JournalTelemetryProducer.readAll()` for `operations.jsonl`. Can we cap the file count and read line-by-line rather than slurping whole files? The mtime skip is good but it doesn't help the first call.

**2. `capabilities()` blocking the loop.** `CronRunTelemetryProducer.capabilities()` calls `detectKind()` on every invocation, and that `spawnSync`s journalctl (10s timeout) then crontab (5s timeout) with no caching. Worst case is a 15s synchronous block, and it re-runs on every call because nothing caches the resolved `CronSourceKind`. Since `telemetry__capabilities` is exposed over the socket, any client can call it in a loop and stall the whole daemon. Cache the kind with a TTL so repeat calls cost nothing.

Then a couple of mediums:

**3. `telemetry__query` has no size or range guard** (`telemetry-mcp.ts:116`). A caller can ask for `stream=tool_call, start=2020-01-01, end=2030-01-01` and get the entire history serialised into one MCP response, all in memory. Cap the range in the Zod schema, or cap the returned array with a `truncated` flag.

**4. `args_hash` on the MCP tool-call producer.** No raw args, good. But hashing low-entropy arguments (short filenames, known commands) and emitting the hash is brute-forceable, which cuts against the "nothing sensitive in telemetry" line the whole safety story rests on. Tool name plus status is probably enough signal for fitness. I'd either drop it or document the tradeoff explicitly on the contract.

One scoping thing rather than a bug: `state-db.ts` (+579, all writes) ships in a PR we're calling the read-only measurement surface. It isn't on the read path so it's not a safety problem, but it's write infrastructure that belongs in brick 2 with the OutcomeLoop, not here. If it can move, this PR gets meaningfully smaller and "brick 1 = read-only" stays honest.

Nothing above is a redesign, just tightening. Also a heads up that it's conflicting with main now (main moved under the stack while these were open), so it'll want a rebase before it can merge regardless of the review.

Sort the resource-safety ones and I'm happy to take another pass.

Nibbler1250:
Thanks for the thorough pass — the resource-safety framing is exactly right for the low-spec target. Sorted the four:

1. **Bounded file reads.** `session-jsonl-provider` (and the journal reader) now cap the file count scanned per call (newest-by-mtime) and read **line-by-line** (new `line-reader.ts`) instead of `readFileSync`-slurping — bounded memory, no whole-file read on the cold call; a file past a max size is skipped rather than OOM'd, and the truncation is logged (no silent cap).
2. **`capabilities()` no longer blocks.** The resolved `CronSourceKind` is cached with a TTL, so the socket-exposed `capabilities()` `spawnSync`s journalctl/crontab at most once per TTL instead of every call — no more 15s worst-case stall or loop-DoS.
3. **`telemetry__query` guarded.** The window is clamped to a max range and the returned array is capped with a `truncated: true` flag, so a 2020→2030 query can't pull the whole history into one response.
4. **Dropped `args_hash`** from the tool-call producer — brute-forceable on low-entropy args, and tool name + status carries the fitness signal, so it keeps "nothing sensitive in telemetry" honest.

Tests updated/added for each; suite green, version bumped.

On your scoping note (5, moving `state-db.ts` to brick 2): agreed it makes "brick 1 = read-only" cleaner — leaving that call + the rebase to the stack owner so it's done as one coherent restructure rather than piecemeal. Happy to do the move if you'd rather it come from here.

TerrysPOV (approved):
Re-reviewed after the hardening pass — approving. All four asks from last round are resolved and each has a test behind it:

- **Unbounded file reads** — `MAX_FILES_PER_SCAN = 200` (non-silent, warns with the real count) + the new `line-reader.ts` streaming both `session-jsonl-provider` and `JournalTelemetryProducer.readAll()`. ✓
- **`capabilities()` blocking** — `CronSourceKind` memoised with a 60s TTL + injected clock; no re-spawn on repeat calls. ✓
- **`telemetry__query` guard** — 90-day window clamp + 5000-sample cap, both flagged with `truncated: true`. ✓
- **`args_hash`** — dropped; only `plugin`/`tool`/`status`/`agent_id` emitted, with a test asserting no `args_hash`. ✓

One thing surfaced in the re-review: the same unbounded-read class was fixed in two of the three readers but missed in the third — `mcp-tool-call-producer.ts` `query()` still slurped the whole append-only audit log via `readFileSync` before the sample cap could act (OOM-reachable via `telemetry__query`). Rather than bounce the whole stack for a one-liner you'd already solved next door, I folded it in as a maintainer commit (`b2f6398`): it now streams via `forEachLineSync` with the 64MB byte cap + non-silent `onTruncate`, matching the sibling producers.

I also rebased the branch onto current `main` (version-file conflicts only, no code conflicts) and bumped the version, so the conflict + version-guard are handled. CI is green and it's mergeable. Two non-blocking follow-ups if you ever revisit: `JournalTelemetryProducer.readAll()` materialises the full (byte-bounded) parsed array where `capabilities()` could short-circuit on first match, and the `line-reader.ts:17` doc comment slightly understates the single-very-long-line worst case (it's still `maxBytes`-bounded). Neither blocks.

Nice work on the hardening — the read-only safety story holds. Good to merge.

PR #287, opened 2026-06-29
feat(tuner): OutcomeLoop + model-routing subject — single-subject proof [#275 brick 2, stacked on #286]

Nibbler1250:
**⚠️ Stacked on #286 (host-telemetry).** Until #286 merges, this PR diff also shows #286's files; review/merge #286 first, then I rebase and only the OutcomeLoop delta remains.

feat(tuner): OutcomeLoop + model-routing subject — single-subject self-tuning proof (#275 brick 2)

The validation-and-improvement half of the self-tuning layer, stacked on the
read-only telemetry brick (#286). Per the #275 staging: ONE subject — model-routing
(reversible, measurable) — with the human gate + tamper-evident audit, before any
other subject is touched.

The loop:
  detect -> propose -> HUMAN GATE -> apply -> measure fitness (via telemetry__*) ->
  auto-revert on regression

What it adds:
- OutcomeLoop engine (outcome-loop, proposal-engine, apply-pipeline, adaptive-scheduler,
  outcomes-ledger) + the wisecron registration surface (index.ts) wired to model-routing ONLY.
- The mutating gate over MCP: `tuner__*` tools (gate-mcp) + `serveTunerOverMcp` — proposals
  stay behind the human gate; apply snapshots a baseline and measures through the MCP
  telemetry provider from #286, so the revert decision rides the same audited measurement.
- model-routing subject (mode_dispatch fitness) as the single active subject. The other 7
  wisecron subjects are intentionally NOT wired here — each lands in its own brick once the
  loop has earned its keep (Terry's explicit staging; the registry honours per-subject
  `enabled` flags so adding them later is config + import only).

Safety / governance posture:
- Every proposal/apply/measure/revert is recorded in the tamper-evident AuditLog chain.
- Mutations are always behind the human gate; the gate auto-denies only explicit matches.
- The standalone-server-over-real-transport integration test is deferred with its server
  wiring (same as #286's transport test); the in-process gate-mcp tests cover the contract.

Tests: 93 pass across outcome-loop, gate-mcp, proposal-engine, apply-pipeline,
adaptive-scheduler, outcomes-ledger, model-routing-subject, render-diff. bun build clean;
bun run lint exit 0.

Stacked on #286 — review/merge that first.


TerrysPOV (changes requested):
Hi mate. This is the big one so I went at it hard, especially the mutation path since it's the first thing in the stack that writes real config.

The important news first: the safety model is right and it holds up. I specifically checked the three things I most cared about and they're all correct:

- **Signature gate is fail-closed.** When `bootstrapWisecron` builds the pipeline without passing `verify`, the default isn't permissive, it's a real `verifyProposalSignature` and apply throws on a mismatch (`apply-pipeline.ts:126` and `:164`).
- **Confinement genuinely confines.** `gate-mcp.ts` realpaths the target's directory then checks exact-set membership against `managedTargets()` (`:54-57`), so a symlink or a `..` in an injected `target_path` can't land a write outside the managed surface. That's the strict version, which is what I wanted.
- **The git commit shells safely.** Array-form `spawnSync("git", ["-C", dir, ...args])` (`:68`), no string interpolation, so no commit-message or path injection.

So the architecture is sound. Here's what I want tightened before it goes in:

**1. `gate-mcp.ts` has a stray NUL byte**, so git treats the whole file as binary (`Bin 0 -> 18693 bytes` in the diff). Same thing that bit the stall-watchdog PR. It's ironic that it's the confinement file, because that's the one a reviewer most needs to read in the diff and currently can't. Strip the NUL and it renders as normal text.

**2. The per-target lock is in-process only** (`apply-pipeline.ts:100`, a `Map` of promise chains). That guards concurrent applies inside one process, but not two processes hitting the same target, the daemon and a CLI `apply`, or an overlapping restart. You added the audit-log multi-writer resync precisely because multi-process writes are expected, so the apply lock should match that assumption. A file lock, or wrapping the read → apply → commit → record sequence in a DB transaction, would close it.

**3. Maturation wiring.** I could only find `runMaturation()` called from tests, not from any runtime tick in this diff, and the outcome-loop header says "Phase 1, observation-only". Can you confirm: is the maturation and keep-or-revert pass actually invoked on a schedule at runtime, or is auto-revert deliberately not live yet? I'm happy with observation-first, that's the right staging, but if revert isn't wired the description shouldn't imply it is. If it is wired, point me at where it's ticked.

Two smaller ones:

**4.** `verdict.ts` treats a guardrail with `count: undefined` as well-powered, so a single unlucky sample on an older metric without count tracking can force a revert. Might cause revert churn on first deploy. Worth checking whether any shipped guardrail lacks count tracking.

**5.** `audit-log.ts` `resyncFromDisk()` adopts the tail record's seq and hash without verifying the last link first, so a forged tail record becomes the new chain head. Only reachable by something that can already write the audit file, so low real risk, but it does weaken the tamper-evidence story it exists to uphold. Verifying just the last prev to current link before adopting would fix it.

The fitness math itself I'm happy with. The two-regime noise handling (relative for non-zero baselines, absolute floor near zero), guardrails-before-target ordering, and the DB-date-backed maturation rather than an in-memory timer (so it survives a restart) are all the right calls.

It's COMPLEX, which is unavoidable for the engine, but it's opt-in with `enabled: false` default, so as long as the footprint when disabled is genuinely zero I'm comfortable on the vision side. Same as the rest of the stack, it's conflicting with main now, so it'll want a rebase before it can merge.

Sort 1 to 3 and clarify the maturation question and I reckon this is close.

Nibbler1250:
Thanks — this was a genuinely careful read of the mutation path. Sorted 1, 2 and 5, and here's the maturation answer (3):

**1. NUL byte** — stripped from `gate-mcp.ts`; it renders as text in the diff now.

**2. Cross-process apply lock** — the in-process `Map` fast-path stays, but the read→apply→commit→record sequence is now wrapped in an exclusive lockfile (stale-lock recovery by pid/mtime so a crash mid-apply can't deadlock). It matches the multi-writer assumption the audit log already makes. New test proves two overlapping applies to one target serialize.

**3. Maturation wiring — your read is right, and I've made the header say so unambiguously.** `runMaturation()` is invoked only by the `tuner__mature` gate tool (and the CLI) — there is **no autonomous timer/tick** driving it in this diff. So the defensive revert is a *wired, operator/cycle-triggered* capability, not a self-driving one; "observation-only" means nothing reverts unless something explicitly calls `runMaturation`. The header now states that outright so the description can't be read as implying live auto-revert.

**4. `verdict.ts` unknown count** — good catch. An unknown (`undefined`) count is now treated as **underpowered** (the safe default), so an untracked metric can't clear the gate — and as a guardrail can't force a `regressed`/revert — off one unlucky sample. I checked the shipped path: subjects that don't implement `measureFitnessCounts` degrade to `count: undefined`, which is exactly the case that used to slip through, so this closes the revert-churn risk on first deploy. New tests cover the unknown-count guardrail and target.

**5. `resyncFromDisk` tamper gap** — `resyncFromDisk()` now verifies the last link before adopting the on-disk tail: the tail's hash must recompute over its own content (self-consistent) **and** its `prev_hash` must chain from the real predecessor (or GENESIS for the first record). A forged tail fails both and we keep our own head. New tests: a forged tail is rejected, a genuine cross-writer append is still adopted.

Suite green. The 5 pre-existing `noNonNullAssertion` warnings are untouched (not introduced here). As with the rest of the stack it's conflicting with main, so it'll want the rebase before merge.

TerrysPOV (approved):
Re-reviewed after the "5 tightenings" pass — approving. I checked each of the five asks against `a71c992` and they're all correctly resolved, with tests:

- **NUL byte in `gate-mcp.ts`** — gone; the confinement file is clean UTF-8 and fully diff-able again. ✓
- **Per-target apply lock** — now a genuine cross-process guard: `withTargetLock` → `withFileLock` → O_EXCL lockfile keyed on the resolved target path, with stale-lock breaking on pid liveness. Both `apply()` and `revert()` go through it. ✓
- **Maturation wiring** — confirmed operator-triggered, not self-driving: `runMaturation()` runs only via the `tuner__mature` gate tool / CLI, no autonomous tick. The file comment and code now agree, so "observation-only" is honest. ✓
- **`verdict.ts` undefined count** — `count === undefined` is treated as underpowered → an unlucky single sample yields `neutral`, not a forced revert. No first-deploy revert churn. ✓
- **`audit-log.ts` `resyncFromDisk`** — now validates the last prev→current link (and self-consistency) before adopting the tail, so a forged tail can't become the chain head. ✓

The three things flagged as already-correct still hold: signature gate fail-closed, confinement realpaths + exact-set membership, array-form `spawnSync` for the git commit (no injection).

One non-blocking observation for a future pass: `withFileLock` degrades to in-process-only locking (with an audit event) when a lockfile can't be created — but that degradation isn't surfaced to the caller (`ApplyOutcome` carries no `lock_degraded` flag). Pre-existing, not introduced here, and only bites on paths where lockfiles can't be created (e.g. some network mounts).

I rebased this onto current `main` (the #286 commits dropped since they landed with #286; only #287's own commits remain — verified the hardened #286 is intact, not reverted), bumped the version, and biome-formatted `apply-pipeline.ts` which had a little drift. CI green, mergeable. Solid work on the engine — good to merge.

PR #288, opened 2026-06-29
docs(tuner): migration map + full implementation/telemetry reference [#275]

Nibbler1250:
## Docs: consolidation map + full implementation/telemetry reference [#275]

Documentation brick for the governed self-tuning layer. **No functional code** — docs + `@deprecated` markers only (the code moves land after #286/#287 so nothing breaks).

### What is in this PR
1. **`docs/skills-tuner-to-tuner-migration.md`** — the migration map: the v1 `skills-tuner` engine + single `skills` subject are superseded by the wisecron OutcomeLoop, the `tuner__*` gate, and the 8 subjects; the shared core primitives become the canonical `tuner/core`. Staged: markers now, code moves after #286/#287 merge.
2. **`docs/tuner-implementation-and-telemetry.md`** — the **operational reference** companion to the #275 RFC: implementation order of the PRs (and why staged) + expected end state; what each PR does and why (#286-#293 + app #285); the complete telemetry surface (11 streams, contract 1.2.0, producers, artifact metrics, proposals->outcomes view-manifest) the control-plane app (#285) renders; the proposal payload + choice/alternatives structure; the gate surface (tuner__*) mapped to the app UI; the generic OutcomeLoop decision sequence (reactive + proactive, risk-tier->observation-window, anti-Goodhart guardrails); and the decision sequence per subject — the 3 live (memory / model_routing / skills) in detail and the 5 in reserve (mcp_plugin / hook / cron / prompt_template / agent / claude_md).
3. **`@deprecated`** markers on the v1 `Engine` and the v1 `skills` subject.

### Why
Gives reviewers (and the #285 app work) a single, exact reference for what we are building, in what order, and what data it exposes — without reading across eight PRs. Aligns with the governance posture: provenance + reversibility + no silent mutation, all surfaced and measurable.

PR #289, opened 2026-06-30
feat(telemetry): skill_access producer hook (closes the declared-but-inert stream)

Nibbler1250:
Ships the **producer** the `skill_access` telemetry stream expects.

**Gap:** the consumer (`SkillAccessTelemetryProducer`, reads `~/.config/tuner/skill_accesses.jsonl`) exists, but no producer was shipped — so `skill_access` is *declared but inert* (advertised, never emitted). Its own doc comment even references a `log-skill-access` producer that did not exist in-tree.

**This PR:**
- `hooks/log-skill-access.mjs` — a PostToolUse(Read) hook that appends `{skill_path, accessed_at}` for each read of a skill file under the skills dir. Zero deps (Node only), env-overridable (`CLAUDECLAW_SKILLS_DIR` / `CLAUDECLAW_SKILL_ACCESS_LOG`), aligned on the canonical `~/.claude/skills`, **never blocks or fails the tool** on a logging error.
- `hooks/hooks.json` — wires it (was an empty `{"hooks":{}}` scaffold).
- Exact directory-boundary match (rejects `…/skills-evil` prefix coincidences).
- **14 tests** (pure entry logic, hook integration, symlink-escape, malformed-input safety, log rotation).

Pairs with the telemetry surface in #286 (the consumer): together the `skill_access` stream emits + is queryable. Decoupled via the log path, so it can land independently.

---

**Review addendum (2026-07-03):**
- **Ships opt-in / disabled.** A PostToolUse(Read) matcher is tool-name only and cannot pre-filter to skill paths, so an always-on hook would fork node on every Read for inert telemetry. `hooks/hooks.json` now registers no PostToolUse entry; the producer header documents the snippet to enable once the #286 consumer is live.
- **Log is now bounded.** The producer owns its trim: `rotateIfNeeded()` rotates `skill_accesses.jsonl` to `<path>.1` past 5 MB (best-effort, never blocks the write).
- **Version bump deferred:** stays 2.2.169; after #288 merges this branch takes merge-main + bump to 2.2.170.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,13 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-<<<<<<< HEAD
-      "version": "2.2.182",
-||||||| parent of 85d918b (feat(telemetry): skill_access producer — PostToolUse hook so the stream actually emits)
-      "version": "2.2.168",
-=======
-      "version": "2.2.169",
->>>>>>> 85d918b (feat(telemetry): skill_access producer — PostToolUse hook so the stream actually emits)
+      "version": "2.2.183",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,13 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
+<<<<<<< HEAD
       "version": "2.2.182",
+||||||| parent of 85d918b (feat(telemetry): skill_access producer — PostToolUse hook so the stream actually emits)
+      "version": "2.2.168",
+=======
+      "version": "2.2.169",
+>>>>>>> 85d918b (feat(telemetry): skill_access producer — PostToolUse hook so the stream actually emits)
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,11 @@
 {
   "name": "claudeclaw-plus",
+<<<<<<< HEAD
   "version": "2.2.182",
+||||||| parent of 85d918b (feat(telemetry): skill_access producer — PostToolUse hook so the stream actually emits)
+  "version": "2.2.168",
+=======
+  "version": "2.2.169",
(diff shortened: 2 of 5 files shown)


TerrysPOV (changes requested):
Careful, well-tested code — the pure `skillAccessEntry` split, the exact directory-boundary test (correctly rejects the `…/skills-evil` coincidence), the **symlink-escape resolution via `realpath`** (with a real subprocess test), the failed-read exclusion, and the never-throws wrapper are all exactly right. **No correctness or security blockers.**

**One decision before this lands — enablement, not code.** As wired, this spawns a `node` process on **every `Read`** the agent makes — the matcher is tool-name only, so it can't pre-filter to skill paths; most reads fork node just to return `null`. PostToolUse hooks run in-loop, so that's standing per-read latency on a daemon whose whole pitch is *lightweight / low-spec* — and right now it feeds a stream nothing consumes until #286 lands.

Ask: **land it disabled / opt-in until the consuming subject is actually live**, so we're not paying per-read overhead for inert telemetry. If you think the always-on cost is negligible, put a measured number on it (overhead per read) and I'm happy to be wrong.

**Nits:**
- The log (`~/.config/tuner/skill_accesses.jsonl`) grows unbounded — no rotation/retention. Fine for a producer, but the stream design should own the trim; note where it happens.
- PR body says "7 tests" — there are ~12. Undercount, harmless.

**Version bump:** you're at 2.2.169, same as #288. Once #288 merges (main → 2.2.169), this PR's version guard will fail (HEAD == merge-base). After #288 lands, merge main + bump to **2.2.170**.

Nibbler1250:
Addressed your 3 asks earlier (opt-in default, log rotation, test-count) and now the 3 Copilot nits too: relative-path guard (`resolve()` before `pathToFileURL`), a real malformed-JSON test (raw bytes, not a valid JSON string), and the header doc. 14/14 green. Ready for re-review when you have a sec.

TerrysPOV (approved):
Re-reviewed after the opt-in + log-bound changes — approving (supersedes my earlier change-request). Both asks are addressed:

- **Opt-in / ships disabled** (the enablement ask): `hooks/hooks.json` has NO PostToolUse entry — the producer only logs once an operator wires it in (documented in the hook header). So no per-Read node fork by default. Rebase preserved main's PreToolUse bash-guard and kept the skill_access hook out — verified hooks.json is identical to main.
- **Log bound** (nit): `MAX_LOG_BYTES = 5 MiB` + `rotateIfNeeded` (rotate to `.1`, one segment retained) — the producer owns its own trim, as asked.

The code you'd already vetted (pure `skillAccessEntry`, realpath symlink resolution, failed-read exclusion, never-throws wrapper) is intact; 14 tests pass. Rebased onto current main + bumped past the guard.

One note, not a blocker: the test lives in `src/__tests__/`, which the new CI 'Bun test (module suites)' job doesn't cover yet (tracked in #304) — so this producer's tests won't run in CI until that cleanup. Good to merge.

PR #291, opened 2026-06-30
feat(tuner): memory subject — reactive hygiene + proactive evidence face [#275 subject 1/3, stacked on #287]

Nibbler1250:
## memory subject (reactive + proactive) [#275 — Phase 4, subject 1/3]

> **Stacked on #287.** First of three separate subject PRs.

The `memory` TunableSubject:
- **Reactive**: dead-pointer / oversize / duplicate hygiene patches on the memory index (data-preservation-checked, atomic writes, 3-way revert).
- **Proactive** (`EvidenceDrivenSubject`): a local signal (index load latency / size / dead-pointer ratio, trend off a stable metric with a noise floor) → only when degraded, asks the feeder for **convergent** external evidence → a gated proposal. The technique here (vectorised retrieval) is architectural → routes to a plugin install (see #290), never engine code.

Also lands the shared **`EvidenceDrivenSubject` contract** (`researchSpec` / `localSignal` / `evaluate` / `confirm` + optional `proposeEvidencePatch`) reused by the model_routing and skills subject PRs.

Tests green; biome clean.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.178",
+      "version": "2.2.179",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.178",
+  "version": "2.2.179",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/tuner/__tests__/wisecron/subjects/memory-shrink.test.ts
+++ b/src/tuner/__tests__/wisecron/subjects/memory-shrink.test.ts
@@ -1,0 +1,101 @@
+import { describe, it, expect, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, writeFileSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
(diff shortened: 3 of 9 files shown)


TerrysPOV (approved):
Hi mate. The one thing I cared about here was that a subject rewriting my memory index can never lose a memory, and you've built it so it can't. I'm happy with this one.

The safety is layered properly:

- `.bak` copy before the write, then temp-file plus rename so the write is atomic and a mid-write crash leaves the original intact (`:457`, `:460-462`).
- Hygiene only removes entries whose backing `.md` is genuinely gone (`filter(existsSync(...))` at `:437`), so a live entry is never pruned.
- And the bit that actually sold me: after reconciling external content you diff the live pointers before against the output and hard-throw if any live pointer got dropped (`:441-446`). So even a bad external proposal can't silently delete a memory, the apply just aborts. That's exactly the belt-and-braces I'd want here.

Confinement is the tightest in the stack too, a straight equality check that `target_path` is the one memory index path (`:397`), plus the symlink refusal (`:453`). And the guardrail does its job: `memory_median_reads_per_entry` is guarded by `memory_index_entry_count` (higher-is-better), so the subject can't juice its target by deleting entries, a shrink shows up as a guardrail regression and gets blocked.

Two small things, neither blocking:

- The dropped-pointer guard uses `... as unknown as Array<[string, string]>` on the `matchAll` result. It works because that regex always has its capture group, but it's a type-lie that would bite if the regex ever changed. Worth replacing with a proper `.filter((x) => x !== undefined)` so the safety net isn't resting on a cast.
- `validate()` runs after `apply()` has already written. So a legal-but-oversized index (over the 200-line cap) gets written and then declared invalid after the fact. The `.bak` means it's recoverable and no memory is lost, but writing then failing validation is untidy. Either cap in `applyStrategy` or validate before the write.

Nice work on this one. Same as the rest of the stack, it stacks on #287 and is conflicting with main, so it needs #287 to land and a rebase before it can actually merge, but the code is good from my side.

TerrysPOV (approved):
Rebased onto current main (the #286 + #287 commits dropped since they've landed; only #291's own 7 commits remain — verified #286/#287 files are identical to main, not reverted), version bumped. Full tuner suite green (282 pass, 0 fail, ~2s — no hangs), lint clean. Content unchanged since the prior approval; re-confirming after the rebase. Good to merge.

PR #292, opened 2026-06-30
feat(tuner): model_routing proactive evidence face [#275 subject 2/3, stacked on #287]

Nibbler1250:
## model_routing proactive evidence face [#275 — Phase 4, subject 2/3]

> **Stacked on #287** (extends its reactive model_routing subject).

Adds the **proactive** `EvidenceDrivenSubject` face to model_routing:
- local signal = median per-session cost (USD) trend (outlier-robust, noise-floored);
- when degraded, asks the feeder for convergent evidence on cost-aware routing;
- a `recommendation`-kind technique (architectural) resolves to a plugin (see #290), never engine code.

Includes the apply/revert **confinement** guard and the cost signal. Reuses the shared `EvidenceDrivenSubject` contract.

Tests green; biome clean.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.174",
+      "version": "2.2.175",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.183",
+      "version": "2.2.184",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.181",
+      "version": "2.2.182",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.179",
+      "version": "2.2.180",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.176",
+      "version": "2.2.177",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.182",
+      "version": "2.2.183",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.178",
+      "version": "2.2.179",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.180",
+      "version": "2.2.181",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.184",
+      "version": "2.2.189",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.175",
+      "version": "2.2.176",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.188",
+      "version": "2.2.174",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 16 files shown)


TerrysPOV:
Reviewed the current draft (own contribution only — the model_routing proactive evidence face + cost signal). Interim feedback since it's still a draft.

**The proactive face itself is solid.** The cost-signal math holds up: `costTrend`'s older-half-vs-recent-half comparison with a noise floor (`minAbsUsd`) and `enoughSamples` gate is correct; the absolute-cap path (`recent > maxRecentUsd`) correctly catches consistently-expensive routing; `evaluate()`'s evidence gating (≥3 independent, ≥1 high-trust, confidence scaling) is right; `apply`/`revert` confinement via `assertManagedTarget` is enforced and tested; and the fixtures are hermetic (temp sqlite DBs, `daysAgo()` uses UTC throughout so they won't go stale — the `b06bade` fix is good). No fail-safe gaps on a missing/corrupt DB.

Three things to look at:

1. **`confirm()` fragility (design).** `confirm()` compares a single-day median (`localSignal().value`) against `before.value` (also a single day). Cost naturally varies day-to-day, so one cheap/expensive day post-apply can false-confirm or false-deny. `evaluate()` uses the robust trend+cap signal — `confirm()` should use the same trailing-average logic rather than a point comparison. Low practical impact today (this subject's `evaluate()` returns `kind:"recommendation"`, so `confirm()` isn't driven in the proactive loop), but it's on the interface and worth tightening.

2. **Test-schema drift (nit).** `model-routing-signal.test.ts` builds `session_costs(date, cost_usd)` while `model-routing-evidence.test.ts` uses the production schema `(date, job, model, cost_usd)`. Both work today because the query only selects `date, cost_usd`, but if the query ever adds a column the minimal-schema test would silently return empty instead of failing loudly. Worth a shared fixture on the production schema.

3. **Heads-up, NOT this PR's bug:** the `swapModelInYaml`/`removeKeywordFromYaml`/`renameKeywordInYaml` helpers have a sibling-mode bug — `if (inMode && /^\S/.test(line)) inMode = false` only resets on a column-0 line, so for an indented multi-mode `modes:` block, editing one mode also rewrites sibling modes' models. This is pre-existing (it shipped with the reactive face in #287, so it's in `main` now), not introduced here — but this PR's `proposeChange` exercises all three helpers, so it's worth being aware of. I'm raising the main-side fix separately.

Since it's a draft carrying the merged #286/#287 commits, it'll need a rebase (dropping those) before it's mergeable — happy to do that once you mark it ready.

Nibbler1250:
@TerrysPOV ready for review — the model_routing **proactive** face (#275 subject 2/3), rebased onto current main (only version bumps conflicted; the #286/#287 base dropped since it's landed — verified the proactive face is net-new, model-routing suite green, full tuner-subjects suite 224/0).

**What it is:** a quality-first benchmark reroute. When the local cost signal degrades, the subject reads the operator's real agentic modes (settings.json list format), pulls PUBLISHED benchmarks (Artificial Analysis Intelligence/coding index + Anthropic SWE-bench to fill AA's Opus-coding gap), and proposes a reroute via `proposeEvidencePatch` — **quality is the veto** (never a reroute that regresses quality), cost the tiebreaker, constrained to runnable (Claude) models.

**Budget:** kept under your 800-line rule — `model-routing-subject.ts` ~740; the benchmark provider / quality gate / config adapter / Anthropic source are each their own file (<300).

**Verified on real data:** running against a real settings.json flips the pick from a hallucinated non-Claude model to a genuine `claude-3-5-sonnet → claude-sonnet-5` (better coding AND cheaper); the live run caught 3 bugs unit tests missed (evaluations{} nesting, candidate-filter, Opus-null). Adversarial hardening suite included.

Notes for review: (1) the write-back puts the AA slug into settings.json — worth confirming ClaudeClaw accepts it as `--model` or whether a reverse alias is wanted; (2) the Anthropic SWE-bench table is a curated seed (high-confidence figures only, documented to extend from model cards).

Nibbler1250:
Self-review pass (against the add-evidence-driven-subject skill) — 2 fixes pushed:

1. **Feeder frontier.** The default `benchmarkProvider` fetched Artificial Analysis directly from the subject, crossing the "the subject never fetches the web itself" rule. Default is now non-fetch; the sandboxed feeder / research scout fetches and injects the benchmarks. (AA is a structured numeric API with a defensive parser, so no injection risk — but the role separation now holds.)
2. **Cache poisoning.** `getModelBenchmarks` cached the caller-filtered result, so a `models`-filtered call + `cachePath` shrank the cache for later unfiltered readers. Now fetches + caches the FULL set and filters on the way out.

+2 regression tests. Suite green, lint clean.

TerrysPOV (changes requested):
Reviewed the full PR (it's grown a lot since my last pass) — the cost-signal proactive face and the new Tier-A quality-gate/benchmark-reroute pipeline are both well-built. Verified independently:

- **Quality-veto gate is correctly fail-closed** (`model-routing-quality.ts`): missing benchmark data → `insufficient_data`, never a silent pass; a quality regression vetoes regardless of price; `blendedPrice()` nulls out non-positive prices so a `$0` "free" model can't manufacture a bogus saving. The "Greg" hardening suite (`model-routing-hardening.test.ts`) is genuinely adversarial — boundary deltas, zero/negative pricing, prototype-pollution JSON, 200-candidate scale, coding-vs-intelligence divergence — not padding.
- **Benchmark fetch fails safe**: missing key / non-200 / network throw / malformed JSON all return `[]`, never throw; a fetch failure with a stale cache present falls back to stale-over-nothing. The `a3e7bc4` self-review fix (cache the *unfiltered* set, filter only on the way out) is real and tested.
- **Tests are hermetic** — all network calls go through an injected `fetchImpl`, clocks are injected, no real keys. 226 tests pass, biome clean.
- The sibling-mode YAML fix (#306/#307) is intact and unregressed, including in the new `setModelInYaml`.
- The Opus coding-gap fallback only fills a null `coding_index`, never overwrites a real AA value — safe by construction.

Two compounding bugs in the newer "reconcile with the REAL agentic config" (`settingsPath`) path — I read the code myself to confirm both, not just relying on a review pass:

**1. The write-back writes an Artificial-Analysis slug into the real config, not a launchable model ID.** `proposeEvidencePatch`'s settingsPath branch does `diff_or_content: setAgenticModel(raw, p.key, p.to_model)`, where `to_model = cand.model_id` — and `model_id` is explicitly documented as "Artificial Analysis slug, e.g. `claude-4-8-opus`". Compare `CLAUDE_SLUG_ALIASES` (`agentic-config.ts:17-24`) — `opus: "claude-opus-4-7-non-reasoning"`, `"claude-3-5-sonnet": "claude-35-sonnet"` — against the real launchable IDs used elsewhere in this same repo (`src/skills-tuner/core/config.ts:38-41`): `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5-20251001`. These are different naming conventions; `claude-opus-4-7-non-reasoning` (AA's variant-tagged catalog name) isn't a valid `--model` value anywhere in this tree. `isRunnableModel` only checks the string contains `"claude"` (`agentic-config.ts:34-36`) — it doesn't validate the slug is actually launchable. This is exactly why it wasn't caught in testing: every settingsPath/reroute test uses synthetic `to_model` values (`"c199"`, `"sonnet-5"`, `"sonnet"`), never a realistic AA slug like `claude-opus-4-7-non-reasoning`.

Applying a reroute proposal through this path today would write an unresolvable string into a user's `agentic.modes[].model`, silently breaking that mode until manually fixed.

Fix: add a reverse map from AA slug back to an operator-recognized model value (or carry the operator's own model string alongside the candidate through the pipeline instead of round-tripping through the AA namespace), and add a settingsPath test using a realistic AA slug to prove the written value actually round-trips to something the runner can launch.

**2. `managedTargets()`/`assertManagedTarget()` were never extended to cover `settingsPath` — so the subject's own confinement guard rejects its own new patch type.** Confirmed directly: `managedTargets()` (`model-routing-subject.ts:511-513`) unconditionally returns `[realResolvePath(this.modesConfigPath)]`, and `assertManagedTarget()` throws unless the target equals that path. Both `apply()` and `revert()` call it first. So calling `apply()` on a settingsPath-targeted patch throws `"target_path is not the managed modes config"` today — the confinement guard (correctly, by design) blocks the subject's own newer feature. `model-routing-confinement.test.ts` only ever constructs the subject with `modesConfigPath`, so this has zero coverage.

Fix: extend `managedTargets()` to include `settingsPath` when configured, and add a confinement test that actually applies a settingsPath-targeted patch end-to-end.

**Net effect of #1 + #2 together:** the "reconcile with the REAL agentic config" deliverable is currently non-functional end-to-end — not reachable in production yet (no feeder wires a real benchmark provider; `getModelBenchmarks` has zero production callers today per grep), which is why 226 tests still pass green. But it needs both fixed before this gets wired to a live feeder, not after — right now it would either throw on apply, or (once #2 is patched in isolation) corrupt a user's config with an unresolvable model string.

Rest of the PR (cost-signal face, quality gate, benchmark fetch, hardening tests) is genuinely solid work. Fix the two settingsPath issues above and this is good to merge.

Nibbler1250:
Thanks — both settingsPath bugs fixed, and you nailed exactly why testing missed them (synthetic `to_model` values).

**1. Launchable write-back.** Added `toLaunchableModel()` (`agentic-config.ts`): it maps any Claude slug — including AA's catalog variants like `claude-opus-4-7-non-reasoning` — to its version-agnostic launchable **tier** (`opus`/`sonnet`/`haiku`). Rather than a reverse map to a pinned id (`claude-opus-4-7`, which goes stale on the next release), the reroute now targets a tier, which `claude --model` always accepts and which matches the agentic config's own default convention (`src/config.ts` uses `model: "opus"`/`"sonnet"`). The write emits the tier; a same-tier no-op is skipped. `isRunnableModel` is now defined via `toLaunchableModel`, so a candidate is only proposed if it has a launchable target.

**2. Confinement covers settingsPath.** `managedTargets()` now includes `settingsPath` when configured, and `assertManagedTarget()` checks membership in it (message updated to "not a managed config").

**Tests (the exact gap):** a settingsPath write-back test using the real AA slug `claude-opus-4-7-non-reasoning` asserting the written value is `opus` and round-trips via `toLaunchableModel`, plus a confinement test that applies a settingsPath-targeted patch end-to-end. 227/0, lint clean, biome clean.

TerrysPOV (approved):
Both settingsPath bugs are fixed — verified by reading the actual code, not just the commit message.

**1. AA-slug write-back → fixed, and better than what I suggested.** Rather than a reverse slug-map (which would need updating every time AA renames a catalog entry), `toLaunchableModel()` maps any Claude slug to its version-agnostic launchable tier (opus/sonnet/haiku) — the runner always accepts these and they never go stale. Confirmed at the actual write-back call site (`model-routing-subject.ts:223`): `const launchable = toLaunchableModel(p.to_model)` is what gets written via `setAgenticModel`, with a same-tier no-op correctly skipped. `isRunnableModel` is now correctly derived from the same helper, so candidates are constrained to what's actually launchable.

**2. Confinement guard → fixed.** `managedTargets()` now includes `settingsPath` when configured, and `assertManagedTarget` checks membership in the full list — confirmed directly in `model-routing-subject.ts:523-527`.

**Test coverage closes the actual gap that hid both bugs**: the new evidence test uses the real problematic AA slug (`claude-opus-4-7-non-reasoning`) and asserts it writes as `opus`, and the confinement test applies a settingsPath-targeted patch end-to-end rather than only ever constructing the subject with `modesConfigPath`.

227/0 tests, biome clean, based on current main. Good to merge.

PR #293, opened 2026-06-30
feat(tuner): skills subject — reactive + proactive content-patch face [#275 subject 3/3, stacked on #287]

Nibbler1250:
## skills subject (reactive + proactive) [#275 — Phase 4, subject 3/3]

> **Stacked on #287.** Last of three separate subject PRs.

The `skills` TunableSubject:
- **Reactive**: patches for dead / low-discoverability skills + new-skill scaffolding (confined to scan_dirs).
- **Proactive** (`EvidenceDrivenSubject`): dead-skill-ratio signal (with a stale-log guard) -> convergent evidence on skill-description optimisation.

**Boundary distinction (the key bit):** a skill-description optimisation is a **CONTENT change the subject applies itself** — `evaluate()` returns `kind:"patch"` and `proposeEvidencePatch()` builds a gated patch on the worst dead skill, confined to scan_dirs. It is **not** a plugin and **not** a detect-only note. Only architectural techniques route to a plugin (#290).

Tests green; biome clean.

## What this brings (why it matters)

The value of the `skills` subject is, in order:

1. **Quality — clearer, discriminative skill descriptions.** The matcher routes on descriptions; vague ones (e.g. a real skill described only as *"Gestion documentaire intelligente"*) are effectively invisible. The subject rewrites them to say *what it does AND when to use it*. Measured: description `context_cost` **−23 %** with discoverability held.
2. **Order — a governed, measurable skill set.** This is what turned the v1 skills-tuner (ungoverned) into a governed subject: `fitnessSignals` + `measureFitness` (context-cost, count guardrail, dead-ratio, LLM `description_quality`). Every change is measured → kept or auto-reverted. The skill catalogue stays tidy and legible instead of drifting.
3. **Speed + autonomy (secondary).** Tighter descriptions mean less context loaded per routing decision and fewer mis-routes to fix by hand — a modest, real efficiency gain, but the headline is quality and order, not raw speed.

Honest scope: on the current set only 2 near-duplicate pairs exist and the targeted overlap test was flat, so the proven win is **description quality + order + −23 % cost**, not consolidation. Full evidence + metrics in `docs/tuner-status-overview.md` (#288).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.181",
+      "version": "2.2.182",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.181",
+  "version": "2.2.182",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/tuner/__tests__/wisecron/subjects/skills-evidence.test.ts
+++ b/src/tuner/__tests__/wisecron/subjects/skills-evidence.test.ts
@@ -1,0 +1,87 @@
+import { describe, it, expect, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, writeFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { SkillsSubject } from "../../../subjects/skills-subject.js";
+import type { StructuredEvidence, LocalSignal } from "../../../wisecron/evidence-driven.js";
+
+describe("SkillsSubject — EvidenceDrivenSubject (proactive)", () => {
+  let dir: string, log: string;
+  beforeEach(() => {
+    dir = mkdtempSync(join(tmpdir(), "sk-"));
+    log = join(dir, "access.jsonl");
+  });
+  afterEach(() => rmSync(dir, { recursive: true, force: true }));
+  const setup = (skills: string[], accessed: string[]) => {
+    for (const s of skills) writeFileSync(join(dir, `${s}.md`), "skill");
+    writeFileSync(
+      log,
+      accessed.map((s) => JSON.stringify({ skill_path: join(dir, `${s}.md`) })).join("\n") + "\n",
+    );
+  };
+  const subj = () => new SkillsSubject({ scanDirs: [dir], skillAccessLog: log });
+  const ev = (o: Partial<StructuredEvidence> = {}): StructuredEvidence => ({
+    technique: "skill-description-optimization",
+    independentSources: 5,
+    highTrustSources: 2,
+    provenInProduction: false,
+    citations: ["a", "b"],
+    ...o,
+  });
+  const degraded: LocalSignal = {
+    metric: "skills_dead_ratio",
+    value: 0.8,
+    unit: "ratio",
+    degraded: true,
+    trend: "degrading",
+    sampledAt: "2026-06-30T00:00:00Z",
+  };
+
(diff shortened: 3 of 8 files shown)


TerrysPOV (approved):
Hi mate. This one I'm happy with. I went hard at the three things that worried me most for a subject that rewrites and scaffolds skill files, and they're all done right.

**Confinement is correct on both write paths.** The scaffold path resolves the target, checks it's inside `scan_dirs`, and refuses to write through a symlink (the `lstatSync(...).isSymbolicLink()` guard before the SKILL.md write). The patch path does the same, plus the `.bak` copy first. Having that symlink refusal on both paths is exactly right, and it's actually more complete than the plugin PR, which only guards the settings write and misses the install dir. Nice work.

**The stale-log guard is spot on.** `skillsHealth` only reports `logFresh` when the access log is younger than 14 days and has actually recorded accesses, and the proactive trigger requires `logFresh` before it will propose anything. So a fresh install or a logging outage gives a stale or empty log, `logFresh` comes back false, and the subject won't go and propose killing every skill. That was my single biggest worry with a dead-skill-ratio signal and you've handled it. The comment "a stale/missing log means BROKEN logging, not unused skills, never trust it" is exactly the right instinct.

**Content patches are backed up** with a `.bak` before the write, and the LLM returns the complete revised skill, so the file isn't left half-clobbered and there's a recovery copy.

One housekeeping note, not blocking: `skills-subject.ts` is 1222 lines, well over our 800 max. It's doing reactive detect, the proactive evidence face, the LLM propose path, and the apply mechanics all in one file. I'd like the apply/scaffold mechanics pulled into their own file at some point, same as I said on the plugin one.

To be straight about scope: I focused-verified the confinement, the stale-log guard and the content-patch safety rather than line-by-lining all 1690 lines, but those are the surfaces that actually matter for a self-modifying skills subject and they're sound. So I'm approving the code.

Practical bit though: this stacks on #287, which I've asked for changes on, and the whole stack is conflicting with main. So it can't actually go in until #287 lands and this gets rebased. The approval is on the code, not a merge-right-now. Once #287 is sorted and you rebase, this one's good from my side.

Nibbler1250:
Closing the loop on the five Copilot inline notes here:

- **`skills-signal.ts` — skill name from `basename`** (would record directory-format skills as `"SKILL"`): already handled on this branch — `const name = base === "SKILL" ? basename(dirname(p)) : base;`, so `<skill>/SKILL.md` resolves to the parent-dir name (matching `listSkillNames`). The note was on an earlier revision.
- **`proposal-engine.ts` — `computeDiff` unbounded / `renderDiff` unused** (:13, :215): addressed on the base brick — `computeDiff()` delegates to `renderDiff` (byte-capped) so previews stay under chat limits.
- **`outcome-loop.ts` — maturation measures at `asOf`, not each row's `window_end`** (:151): fair observation — a late `runMaturation` measures a window ending at run-time rather than the intended `window_end`. It's a deliberate "measure at maturation time" tradeoff today (and maturation isn't auto-ticked in this phase), but if you'd prefer the post-window pinned to `window_end` I'm happy to open a follow-up.
- **`tuner-view-provider.ts` — first non-null verdict can be a guardrail row** (:102): display-only (which verdict shows in the timeline); can be tightened to prefer the target row. Follow-up candidate if you want it.

The first two are done; the last two are on the #287 base brick and land here via rebase. Nothing outstanding that blocks — flagging the two design points in case you want follow-ups.

TerrysPOV (approved):
Rebased onto current main (the #286 + #287 commits dropped since they've landed; only #293's own 6 commits remain — verified #286/#287 files identical to main, not reverted), version bumped. Full tuner suite green (258 pass, 0 fail, ~1.7s — no hangs), lint clean. Content unchanged since the prior approval; re-confirming after the rebase. Good to merge.

TerrysPOV (approved):
Re-rebased onto current main after #291 merged. The engine-registry file (index.ts) genuinely conflicted — both #291 (memory) and #293 (skills) register their subject there — resolved to register BOTH (memory + skills), verified type-clean and the registry tests pass. #291/#287 files identical to main (not reverted). Full tuner suite green (255 pass, 0 fail, no hangs), lint clean. Good to merge.

TerrysPOV (approved):
Re-rebased onto current main after #290 merged. index.ts conflicted again — main now registers memory (#291) + mcp_plugin (#290), and this PR adds skills. Resolved to register all four subjects (model-routing + memory + mcp_plugin + skills); verified type-clean, #290/#291 files identical to main (not reverted). Full tuner suite green (317 pass, 0 fail, ~2s). Content unchanged; re-confirming after the rebase. Good to merge.

Issue #296, opened 2026-07-02
Resolution path: eliminate mid-tool-call session stalls (3-PR plan) — for #295


TerrysPOV:
Resolution path / tracking epic for **#295** (main Discord session wedged ~11h on a hung `Bash` tool call → daemon fully unresponsive; the `#272` runaway watchdog structurally cannot catch it).

## Root cause recap (full evidence in #295)
A single tool call that never returns blocks the session's single-threaded loop, taking the whole daemon offline. In the incident, `claude -p` batch jobs backgrounded with `&` **inside** a `Bash` tool call held the persistent shell's output stream open, so the follow-up call never saw EOF and never returned. Underlying smell (raised by Terry): **agent "jobs" have no first-class home in the bus, so Claw improvises `Bun.spawn(["claude","-p",…])` and launches it unsafely.**

---

## Key design decision — how do we tell a *stall* from a *legit long-running tool*?

We do **not** use a flat "no result in N minutes" — that kills legit long jobs and/or misses real stalls. We layer signals:

1. **Ground truth = the tool's own timeout.** Every tool call is bounded: Claude Code's `Bash` tool has a hard max (600s / 10 min); other tools return in ms–seconds. No legit call should exceed its own declared/max timeout, so **a call still outstanding past its max + a grace margin is stalled *by definition*** (its contract was broken — it's not "busy"). This makes the kill-threshold **per-tool**, not a global guess. (PR #2 enforces these caps daemon-side so this ground truth is reliable — in the incident the `Bash` `timeout` was `None` and unenforced, which is why it ran unbounded.)
2. **Corroborating signals (between thresholds):**
   - **Output silence** — any stream output (progress lines) resets the clock; a job that emits progress is provably alive. The stall case (held pipe) emits nothing.
   - **Process liveness** — is the tool's process tree consuming CPU / doing I/O, or blocked idle forever?
3. **Escalation ladder, not a guillotine:** at a **soft** threshold → *warn / notify* (ping Terry: "session X looks stalled") without killing; only past a **hard** ceiling (well beyond any legit tool) → *kill + respawn*. A legit 12-minute job gets a heads-up, not a kill.

**Defaults (all configurable):**
- `stallWarnSeconds` ≈ **300s** — notify only, never kill.
- `stallKillSeconds` = **max-legit-tool-duration + grace** (~15 min pre-PR#2; = the enforced Bash cap + grace once PR#2 lands).
[...]

Issue #295, opened 2026-07-02
Main session wedges indefinitely on a hung Bash tool call → daemon fully unresponsive; #272 watchdog can't recover it


TerrysPOV:
## Summary

On **2026-07-02** the daemon went **fully unresponsive on Discord for ~11 hours** (last reply 09:10 local / 08:10 UTC; recovered only by a manual `systemctl restart`). Root cause: **the main Discord session blocked indefinitely inside a single hung `Bash` tool call**, so its event loop could never process new messages. The `#272` runaway watchdog is deployed but **structurally cannot detect or recover this class of failure**. This is a reliability hole that will recur.

## What actually happened (evidence, UTC)

The main session was asked to kick off two background research/draft jobs. From its transcript (`-home-claw-project/3389a969….jsonl`):

- **08:13:09** `Bash` (`timeout: None`, `run_in_background: None`):
  ```
  nohup bun run /tmp/fire-reg-aissure-draft.ts > /tmp/reg-aissure-draft.log 2>&1 &
  nohup bun run /tmp/fire-reg-aissure-sa.ts    > /tmp/reg-aissure-sa.log    2>&1 &
  ```
  Each `fire-*.ts` does `Bun.spawn(["claude","-p", …], {stdout:"pipe",stderr:"pipe"})` — i.e. spawns a **long-lived headless `claude -p` job** (60-min cap), **bypassing the PTY/bus**.
- **08:23:34** `Bash` (`timeout: None`): `sleep 15 && tail -5 …draft.log … tail -5 …sa.log` — **this call never produced a `tool_result`.** It is the last event in the transcript. The session was wedged from here on.

**The jobs themselves were fine** — this was not a stuck delegation:
- `reg-aissure-sa` → **exit 0 at 09:13**, "Done. 3 South Africa proposals written to the AIssure tracker and posted to Discord."
- `reg-aissure-draft` → hit its own 60-min cap (**exit 143 at 09:10**).

So the delegated work completed/terminated on its own by ~09:13, yet the **main session stayed blocked for another ~10 hours** and never observed the results or answered later messages.

## Root cause

[...]

PR #297, opened 2026-07-03
[WIP] feat(bus): stall watchdog for wedged PTY sessions — SPEC + detection core (PR 1/3, #296)

TerrysPOV:
**Status: WIP — early-review checkpoint.** This is PR 1/3 of the resolution path in #296 (fixing the #295 outage: a bus PTY session wedged ~11h on a hung `Bash` tool call). I'm opening it now — with the SPEC + the detection core landed and tested — so the *design* and *core logic* get reviewed **before** I build the mechanical plumbing on top. Not merge-ready yet (see "Not here yet").

@Nibbler1250 — Terry's asked you to review this checkpoint. The two things I'd most value your eyes on are (a) the **detection model** and (b) the **auto-discovery** design.

## The problem (full detail in #295)
A persistent bus PTY session drives a single-threaded loop. If a tool call never returns — e.g. a `Bash` call whose `&`-backgrounded children hold the shell's output stream open — the loop blocks forever and the session ignores **all** later messages. The `#272` runaway watchdog is structurally blind to this (it trips on *too much* activity; a stall is *idle*), and `src/watchdog.ts` only guards the one-shot `runClaudeOnce` path, not the long-lived PTY sessions.

## What's in this PR now
- **`.planning/stall-watchdog/SPEC.md`** — the full design (6-section house style) incl. the auto-discovery decision.
- **`src/bus/stall-watchdog.ts`** — the detection core:
  - Subscribes to the bus event stream (the JSONL tailer already emits `response.tool_use` / `tool_result`); tracks outstanding tool calls per session.
  - Per-tool ceilings keyed on the `tool_use` name (fast `Read`/`Edit`/… `60/120s`, `Bash` `300/900s`, MCP/unknown `300/900s`), warn→kill ladder. **The stall-vs-legit-long-tool call is made by the tool's own bound, not by silence** (silence is identical for both at the transcript level).
  - Recovery is delegated via an injected `restart` dep so it **reuses `SessionManager.restart`** (inherits the kill+respawn + `MAX_RESTARTS_PER_WINDOW` rate-limiter); kill-dedup prevents double-restart across overlapping sweeps.
  - **Auto-discovery:** every kill captures a pre-restart liveness snapshot and is classified `genuine_wedge` vs `suspected_false_positive`; a false positive flags the operator with a suggestion to raise the exact ceiling — so the conservative defaults self-correct instead of silently harming legit long tools.
- **`src/bus/__tests__/stall-watchdog.test.ts`** — 17 unit tests (classify, `evaluateStall`, ingest transitions, kill orchestration incl. dedup / warn-only mode / failed-restart surfacing). All green, biome-clean.

## Not here yet (follows after this review)
1. `src/bus/stall-forensics.ts` — the real `/proc` CPU probe (`null` off-Linux), `classifyKill`, `appendStallKillAudit` + tests.
2. `src/config.ts` — `parseStallWatchdogConfig` + `DEFAULT_SETTINGS` wiring.
3. `src/bus/session-agent-process.ts` — track `lastDataAt` + expose `pid` for the probe.
4. Wiring point (`runtime-mount`/`wiring`) — instantiate `StallWatchdog`, bind the real deps + an integration test.

## Review asks
- **Detection model:** is "outstanding `tool_use` past a per-tool ceiling" the right signal? Any tool that legitimately blows past `Bash`'s 900s that I should pre-seed a ceiling for?
- **Auto-discovery:** the `genuine_wedge` vs `suspected_false_positive` classification (CPU-tree progress + output recency, best-effort/degrades off-Linux) — sound? Over-engineered?
- **Reuse of `SessionManager.restart` as the 4th trigger** (`reason: "stall"`) — right integration point?

Refs: #296 (epic), #295 (incident).


Nibbler1250 (commented):
Reviewed the two issues, the SPEC, the implementation, and ran the suite (17/17 green on the checkpoint). Short version: **the core is sound, ship the direction** 👍 — the stall-vs-runaway framing in #295/#296 is exactly right, and the two design calls that matter most are correct.

I verified the two assumptions the whole thing rests on, since a mismatch would make the detector a silent no-op:

- **Event shapes match exactly** — the tailer emits `response.tool_use → {id,name}` (`jsonl-tailer.ts:470`) and `tool_result → {tool_use_id}` (`:449`), which is precisely what `ingest` keys on. ✔
- **Clock domain is consistent** — `publish()` stamps `ts` from the transcript line's own `timestamp` (epoch ms, `Date.now()` fallback), same domain as `deps.now()`, so `now − startedAt` is real elapsed wall-clock. ✔ (this was my main worry and it holds)

### Your three asks
**(a) Detection model — right signal.** "Outstanding `tool_use` past a per-tool ceiling" is the only reliable discriminator given silence is identical for stall vs. legit-long. Warn→kill ladder + most-severe/longest-outstanding is clean, and `evaluateStall` being pure is exactly how I'd want it for testing.

**(b) Auto-discovery — sound, not over-engineered.** "Suggest, never auto-apply" + degrade-to-`unknown` off-Linux is the correct posture, and it's a genuinely nice future feed into the #275 tuner (a gated, evidence-backed ceiling-raise proposal is literally the tuner's shape). One caveat: the actual `classifyKill` truth-table — the part most likely to have edge-case bugs — lands in the follow-up (`stall-forensics.ts`), so this checkpoint reviews the orchestration around it, not the classifier itself. Fine for a 1/3, just flagging the riskiest logic is still to come.

**(c) `restart({reason:"stall"})` as the 4th trigger — right integration point.** The payoff is inheriting `MAX_RESTARTS_PER_WINDOW`, which is exactly what stops a stall→respawn→stall infinite loop. A fresh kill path would have re-introduced that risk.

### Findings

**1. `Task`/sub-agent + long MCP tools will be false-positive-killed at the 900s `default` — and this is your "anything to pre-seed" ask.** A sub-agent dispatch (`Task`/`Agent`) or a workflow/research `mcp__…` tool legitimately runs 10–30 min; both currently fall into `default` (kill 900s). Given the incident itself was long agent jobs, and PR #3 makes agent-jobs first-class (likely surfaced as a `Task`-like or `mcp__` tool), I'd pre-seed a generous `task` class (~ the `claude -p` 60-min cap) and split `mcp` from `default`. Auto-discovery backstops it, but killing a running 20-min research job is a real regression you can design out. This is the one change I'd want before it goes live.

**2. Restart-failure → critical-notification storm.** In `handleKill`, if `deps.restart` throws (rate-limiter exhausted), the session isn't dropped and `killing` is cleared in the `finally` → the next 30s sweep re-kills → re-notifies `critical`, every 30s indefinitely. The rate-limiter bounds *respawns* but not *notifications*, so an unrecoverable wedge becomes a Discord flood. A per-session cooldown (or dedup) on repeated kill failure fixes it.

**3. 🐛 A stray `NUL` byte makes the whole file git-binary.** `key()` uses a literal `\0` as the agent/session separator, so git flags `src/bus/stall-watchdog.ts` as **binary** — it shows as `Binary files differ` / `0 +0 −` in this PR's diff (the core file is currently un-reviewable in the web UI; that's why the file list shows `0+/0-`). Replacing the `\0` with the space it visually reads as makes it a normal 423-line text addition. Worth fixing regardless of the above.

### Offer
I've got all three implemented locally against the checkpoint (per-tool `task`/`mcp` ceilings + `restartFailureCooldownMs` + the NUL fix), with 4 added tests (22/22 green, biome-clean). Since this is your active WIP branch I didn't push to it — happy to hand it over as a patch or a side branch you can cherry-pick, or you fold the feedback in your own style for #2. Your call.

Net: good to land as the 1/3 checkpoint, ideally with #1 seeded (or explicitly deferred to PR#2's config) and the NUL stripped so the diff is reviewable. Then go on 2 and 3 and mark them for me. Nice work — the escalation-ladder design is the right shape.

TerrysPOV:
@Nibbler1250 thanks so much for the review. We are a team and I'm not precious about who owns what/does what. my only interest is what's best for the project and moves us forward. As you've done the work, by all means do a maintainer commit and I'll run the final review, final fixes (if any) and move onto 2. Thanks again for your work and support on this.

Nibbler1250:
Thanks Terry — done, maintainer commit is in (`c3a8d5f`):

- **`task` + `mcp` ceiling classes** — `Task`/`Agent` → warn 1800 / kill 3600 (≈ the `claude -p` 60-min cap from #295); `mcp__…` → warn 600 / kill 1800; both split from `default`. No more false-positive kill of a long sub-agent/research tool.
- **`restartFailureCooldownMs`** (default 300s) — after a failed restart the session backs off instead of re-killing + re-alerting every sweep.
- **Stripped the stray `NUL`** in `key()` — the file now renders as a normal 423-line text addition instead of `Binary files differ`, so your final diff is reviewable in the web UI.

+4 tests (22/22 green, biome-clean). The 17 unrelated `src/bus` failures (`BusCore IPC`, `session-id collision`…) are pre-existing on the branch, not from this change. All yours for the final review + PR #2 — shout if you want anything adjusted.

TerrysPOV:
@claude /review

TerrysPOV:
## ✅ Approved to merge

All review feedback is in and verified:

- **Nibbler1250's 3 findings** — task/mcp ceiling classes, `restartFailureCooldownMs`, and the stray-`NUL` strip — landed in `c3a8d5f`, each backed by a real test, and independently confirmed by the `@claude` review. Re-ran the suite on the current branch: **22/22 green, biome-clean**.
- **SPEC reconciled** (`c689258`) — the `@claude` review's one substantive note (the SPEC still described `ceilings: {default, perTool?}` while the code landed a fixed 5-class model) is fixed. §3/§4/§5 now describe the `fast/bash/task/mcp/default` classes and the `restartFailureCooldownMs` decision, so the merged doc matches what ships.

### Deferred follow-up (placed)
The `@claude` review's second, non-blocking note — `restartFailedAt` has no explicit reset path (fine today because a fresh `session.init` always builds a new `SessionStallState`, but a hazard if later wiring reuses a state object across a wedge→recovery cycle without `ingest`) — is captured in **SPEC §7** and **placed in the stall-watchdog *completion* PR** (the remaining PR #1 plumbing: `stall-forensics.ts` + `config.ts` + `session-agent-process.ts` + wiring). That completion is tracked as a **new PR**, not #296's PR #2 (Bash-exec hardening) or PR #3 (agent-job primitive), which keep their scope. The action there is a one-line invariant + a test that a re-observed session starts clean after a failed-then-succeeded restart.

Merge whenever ready; then I'll open the completion PR and mark it for @Nibbler1250.

Issue #268, opened 2026-06-22
governance: watchdog activeInvocations leak suspends healthy long-lived agents


TerrysPOV:
## Symptom

Healthy cron-driven agents (Suzy, Reg, Publisher) appeared "stuck" in production for ~9 days. The user-facing diagnosis layer attributed it to PTY timeouts; the actual fault is the governance watchdog suspending invocations that are completing turns cleanly.

## Root cause

`src/governance/watchdog.ts:checkLimits()` evaluates `elapsedSeconds = Date.now() - record.startedAt` against `limits.maxRuntimeSeconds` (default 7200s = 2h). **No code path removes records from `watchdogIndex.activeInvocations` on turn-end.** A single long-lived agent process therefore accumulates a record per cron firing, and once a record is > 7200s old every subsequent `checkLimits` call returns `state: "suspend"` and an escalation handoff fires.

Production daemon had:
- **1,096 active invocations** in `watchdog-index.json` (oldest dated `2026-04-05`, when #257 first shipped).
- **140 handoff files** in `.claude/claudeclaw/handoffs/`, every single one with `reason: "Escalation from watchdog: Watchdog suspend: Limits triggered: maxRuntimeSeconds=N/7200"`.
- Daemon kept generating new suspend handoffs ~every 2 hours, masquerading as agent failures in the user-facing layer.

Meanwhile, the agents themselves were healthy — their JSONL transcripts (`~/.claude/projects/-home-claw-project/<sid>.jsonl`) showed clean `end_turn` + `stop_hook_summary` events on every cron firing.

## Secondary defect

`WatchdogConfig.enabled` is parsed (`watchdog.ts:59,87,110`) but never consulted. `checkLimits` runs to completion regardless. There is no path for operators to opt out via config.

## Reproduction

1. Start daemon with any agent on a cron schedule.
2. Let it run for 2h+ (one cron interval is enough if the cron is hourly).
3. Subsequent invocations trip `maxRuntimeSeconds`, watchdog emits suspend handoff for every check.

## Workaround (immediate)

Hot-patch in PR #<this PR number>: short-circuit `checkLimits` when `watchdogConfig.enabled === false`, default `enabled` to `false`. Deployed on production server `2026-06-22T17:00 UTC`; watchdog index reset to empty. No new suspend handoffs since.

## Real fix (follow-up)

1. Add `completeInvocation(invocationId)` to remove a record from `activeInvocations` on `end_turn`.
[...]

PR #298, opened 2026-07-03
fix(governance): implement watchdog kill escalation (#280)

### LINKED ISSUE #280: Watchdog: checkLimits never returns "kill" — dead branch / no kill escalation (from #272 review)
## Summary

`checkLimits` never returns `"kill"`, so the runner's `=== "kill"` handling is dead code and there is no kill-escalation path. Decide: implement kill escalation, or remove the dead branch and document kill as governance-layer only.

## Context

`checkLimits` in `src/governance/watchdog.ts` only ever returns `healthy` / `warn` / `suspend` — its own comment notes "kill state escalation requires explicit handling at execution layer; suspend is the highest state returned by checkLimits".

Consequently the watchdog-decision handling in `src/runner.ts` includes a `decision.state === "kill"` branch that can never fire, and the conceptual "kill" escalation (terminate a truly runaway invocation) is unimplemented.

This is **pre-existing** — surfaced during the PR #272 5-agent review, not introduced by it.

## Proposed follow-up (pick one)

- **Implement kill:** have `checkLimits` escalate to `kill` at a hard ceiling well above the `suspend` thresholds (e.g. a multiple of `maxRuntimeSeconds`/`maxToolCalls`), and wire/keep the runner's kill handling; add tests.
- **Remove the dead branch:** delete the unreachable `=== "kill"` handling in the runner and document that `kill` is a governance-layer outcome only (mapped elsewhere), keeping the type but not the runtime branch.

## Acceptance criteria

- Either `kill` is reachable from `checkLimits` and covered by a test, **or** the dead runner branch is removed and the kill semantics are documented.

## Source

PR #272 5-agent review. Files: `src/governance/watchdog.ts` (`checkLimits`), `src/runner.ts` (the `=== "kill"` watchdog-decision branch).

Nibbler1250:
Resolves #280.

`checkLimits` only ever returned up to `suspend`, so the runner's `decision.state === "kill"` branch — and `handleTrigger`'s full `kill` case (audited terminate: records the kill, drops the active-invocation record, emits a `killed` event) — were **dead code**. A genuine runaway past the suspend line had no distinct outcome.

Per the issue's two options, this takes **"implement kill"** (the chosen path):

- `checkLimits` escalates to `kill` when any metric reaches `limit * killCeilingMultiplier` — new `WatchdogLimits.killCeilingMultiplier` (default **2**; set ≤1 to opt out and keep suspend as the ceiling). Applies to `maxToolCalls` / `maxTurns` / `maxRuntimeSeconds` / `maxRepeatedTools`, each as a tier above its existing suspend line.
- `recommendedAction` maps `kill → "terminate"`; removed the stale "suspend is the highest state" comment.
- `runner.ts`: `kill` is now reachable — both suspend and kill route through `handleTrigger`, which dispatches the audited terminate path for kill.

**Scope note:** this makes the governance `kill` outcome *reachable + audited* (the #280 acceptance criterion). Actually terminating a wedged subprocess mid-tool-call stays the stall-watchdog's job at the bus layer (#295/#297) — deliberately out of scope here.

**Tests:** +4 — escalates past the 2× ceiling; reachable end-to-end via `handleTrigger` → `"terminated"`; just-under-ceiling stays `suspend`; `killCeilingMultiplier ≤ 1` opt-out keeps suspend. Full watchdog suite **29 pass / 0 fail**, biome-clean, version bumped.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.186",
+      "version": "2.2.187",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.186",
+  "version": "2.2.187",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/__tests__/governance/watchdog.test.ts
+++ b/src/__tests__/governance/watchdog.test.ts
@@ -14,7 +14,7 @@ import {
   getWatchdogConfig,
   resetWatchdog,
 } from "../../governance/watchdog";
-import { join } from "path";
+import { join } from "node:path";
 
 const WATCHDOG_DIR = join(process.cwd(), ".claude", "claudeclaw", "watchdog");
 
@@ -24,7 +24,7 @@ describe("Watchdog", () => {
 
     // Clear watchdog directory for test isolation
     try {
-      const { rm, readdir } = await import("fs/promises");
+      const { rm, readdir } = await import("node:fs/promises");
       const files = await readdir(WATCHDOG_DIR);
       for (const file of files) {
         if (file !== ".gitkeep") {
@@ -85,9 +85,9 @@ describe("Watchdog", () => {
 
     const metrics = await getActiveInvocation(invocationId);
     expect(metrics).not.toBeNull();
-    expect(metrics!.toolCallCount).toBe(5);
-    expect(metrics!.turnCount).toBe(2);
-    expect(metrics!.invocationId).toBe(invocationId);
+    expect(metrics?.toolCallCount).toBe(5);
+    expect(metrics?.turnCount).toBe(2);
+    expect(metrics?.invocationId).toBe(invocationId);
   });
 
   test("should increment tool call count", async () => {
@@ -98,8 +98,8 @@ describe("Watchdog", () => {
     await incrementToolCall(invocationId, "write_file", { path: "/tmp/output.txt" });
 
     const metrics = await getActiveInvocation(invocationId);
-    expect(metrics!.toolCallCount).toBe(3);
-    expect(metrics!.toolCalls.length).toBe(3);
+    expect(metrics?.toolCallCount).toBe(3);
+    expect(metrics?.toolCalls.length).toBe(3);
   });
 
   test("should increment turn count", async () => {
@@ -109,7 +109,7 @@ describe("Watchdog", () => {
     await incrementTurnCount(invocationId);
 
     const metrics = await getActiveInvocation(invocationId);
-    expect(metrics!.turnCount).toBe(2);
+    expect(metrics?.turnCount).toBe(2);
   });
 
   test("should return healthy when under limits", async () => {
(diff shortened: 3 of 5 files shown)


TerrysPOV (changes requested):
Reviewed the commit. The escalation-tier plumbing is right: `escalateState()` fixes the old bug where `worstState` got stuck at `"warn"` after a warn-then-hard-limit sequence, `killCeilingMultiplier` (default 2) makes the previously-unreachable `"kill"` state reachable, the boundary/opt-out tests are coherent, and the bundled #268 lifecycle-leak fix (short-circuit metric recording when disabled, `finally`-block `clearInvocation`, 24h startup eviction) is sound. Default flipped to `enabled: false` — good.

One blocker, and it's the headline feature:

**The "kill" escalation never actually kills the process.** `handleTrigger("kill")` removes the invocation from `watchdogIndex.activeInvocations`, writes an audit record, and returns `{ action: "terminated", success: true }` — but it sends **no signal to the running claude subprocess**. The runner (`runner.ts`) calls `checkLimits` → `watchdogHandleTrigger` for both `suspend` and `kill`, and neither path calls `killActive()` / `proc.kill()` / `killAllPtys()` (the actual termination primitives). So the watchdog correctly detects a runaway, escalates to `kill`, audits it, clears the index — and the runaway keeps running. The `"terminated"` return value is misleading to any caller that trusts it.

Fix: on the `kill` path, actually terminate — call `killActive()` (or the runtime-appropriate primitive) before returning `"terminated"`. If termination is intentionally deferred elsewhere (e.g. the bus stall-watchdog / session-manager owns process kills and this legacy governance watchdog is audit-only), then don't return `"terminated"` — return something like `"flagged"` and document that the escalation is advisory, so nobody assumes the process was stopped.

Two minor notes (not blockers):
- The `maxRepeatedTools` kill ceiling (`killsAt(maxRepeatedTools)`) compares against `totalRepeatedCalls` — the sum across ALL repeated patterns, not any single pattern exceeding the ceiling. So `3×2=6` can fire from two 3-call patterns, which isn't really "a runaway pattern". Same counting as the 0.7× warn branch (internally consistent), but the semantics differ from the per-metric tool/turn kill logic — worth documenting or aligning.
- `handleTrigger("kill")` and the runner's `finally` `clearInvocation` both `saveWatchdogIndex()`. Harmless under single-threaded JS (sequential), but the double-write is fragile; a note would help.

It's conflicting with current main, so it'll want a rebase too. Wire the actual termination (or make the advisory nature explicit) and this is good.

Nibbler1250:
Thanks — addressed the blocker (chose to actually terminate, per option a).

**The kill now really terminates.** Rather than importing `killActive` into `watchdog.ts` (which would create a governance→runner import cycle, since `runner.ts` already imports `handleTrigger`), the terminator is **injected**: `handleTrigger(context, decision, { terminate })`. The runner passes `killActive` at both call sites (the main post-completion check and the post-retry check). Governance stays decoupled from the runtime, and it's unit-testable.

- `success` now reflects reality: if the injected terminator runs and reports nothing killed, `success:false` and the audit event records `kill-failed`. Audit-only callers with no terminator stay `success:true` (unchanged).
- Added two tests: the terminator **is** invoked on kill, and `success:false` surfaces when it finds nothing to kill.

Honest note on timing: this watchdog check fires just after `recordInvocationCompletion`, so it's retrospective — `killActive()` tears down whatever is active for that agent (same semantics as the existing telegram `killActive` kill command), which is the right primitive to stop a runaway from continuing into the next turn/retry. If you'd prefer the kill to move earlier in the loop, happy to follow up.

Rebased onto current main, version bumped (2.2.186), `bun run lint` clean, full watchdog suite 31/0. The two non-blocking notes (kill-ceiling counting semantics, double `saveWatchdogIndex`) I left as-is to keep this focused — can do a follow-up if you want them closed.

TerrysPOV (approved):
Re-reviewed after Nibbler's update (`075d366`, "kill escalation actually terminates the runaway") — approving (supersedes my change-request). My blocker is fully resolved:

The kill path now **actually terminates the process**. `handleTrigger` takes an injected `terminate` (the runner passes `killActive` at both call sites), and the `kill` branch calls it — so the runaway's subprocess is signalled (`killActive` → `proc.kill()` on every `mainActiveProcs` entry + PTY disposal), not just audited. The dependency-injection shape is the right call: governance stays decoupled from the runtime (no import cycle), and audit-only callers with no terminator still work. The reporting is honest too — `executedAction` records `killed` vs `kill-failed`, and `success` is `false` when the terminator ran but found nothing to kill (vs `true` for audit-only). Both paths are covered by the new tests (terminator-invoked → success/terminated; terminator-returns-false → success:false). Watchdog suite 31/0.

The two minor notes from my earlier pass (cross-pattern `maxRepeatedTools` counting; the double `saveWatchdogIndex` on the kill path) were explicitly non-blocking and remain fine as-is.

Maintainer commit from me: rebased onto current main (past #307) + bumped to 2.2.187 — the branch was level with main at 2.2.186 after #307 merged, so the version needed to advance. Good to merge.

Nibbler1250:
Heads-up / apology: my force-push here raced your `chore: bump version after rebase onto main` commit and overwrote it — that was my mistake (I re-checked the lease and pushed instead of stopping when the first `--force-with-lease` correctly rejected). I've since rebased the branch onto current `main` (so #307 is in the base, same as your rebase intended) and kept the version bump (2.2.187, above main's 2.2.186). Net state is what your commit was aiming for; sorry for stepping on it.

PR #299, opened 2026-07-03
feat(ui): live Kanban board of subagent activity (#294)

### LINKED ISSUE #294: Kanban board: review — remove or properly scope + wire (currently a manual, agent-unwired shell)
## Decision needed

The dashboard ships a **Kanban** tab, but it is a **manual, agent-unwired shell** — half-built. Decide: **remove it**, or **properly scope + wire it** into something that earns the tab.

It is **not** a plugin and **not** an MCP (asking the daemon correctly returns "no kanban plugin/MCP"). It is a native web-UI widget backed by a local JSON file.

## What it is today

- **Storage:** a plain file, `.claude/claudeclaw/kanban.json` (relative to the daemon cwd), three columns — `todo` / `in_progress` / `done`. Empty by default; no seed/demo cards.
- **Server surface:** exactly two routes — `GET /api/kanban` (read the file) and `POST /api/kanban` (overwrite the whole board with the client payload). `src/ui/server.ts:821-830`.
- **UI mutation surface (manual only):** add a card (always lands in **To do**) and clear the **Done** column. That's it. `src/ui/page/script.ts` (~L1688+, `addKanbanTask` / `clearKanbanDone` / `saveKanban`).
- **No move support:** there is no drag-and-drop and no other move UI — a card added to "To do" cannot be moved to "In progress" or "Done" from the interface.

## What's half-built (the tell)

`src/ui/services/kanban.ts` exports `addCardToColumn()` and `moveCard()`, and `KanbanCard` carries `agent_type` / `started_at` / `completed_at` — clearly scaffolded to **auto-track agent/subagent tasks** (spawned → in_progress → done). But **those helpers have zero callers anywhere in the codebase**, and nothing on the daemon/session/cron side ever writes to the board. The intended auto-population was never wired. So it shipped as a manual sticky-note board wearing an agent-tracker's clothes.

## Options

**A. Remove.** Drop the tab, the two routes, `services/kanban.ts`, and the front-end block. A manual board with no move support and no agent integration is thin — arguably dashboard cruft. Lowest maintenance; aligns with "lightweight, features users actually use."

**B. Scope + build the intended feature.** Wire the dead helpers so the board becomes a live view of what Claw is doing — e.g. drive cards from subagent spawns / `TodoWrite` / job runs (todo → in_progress → done, stamping `agent_type` + timestamps), and add move/complete transitions. Turns a decorative tab into a genuine operator signal. Needs a real spec (what events feed it, retention/clear policy, single-operator scale) before any code.

## Key files

- `src/ui/services/kanban.ts` — service + JSON storage; the unwired `addCardToColumn` / `moveCard`.
- `src/ui/server.ts:821-830` — `GET`/`POST /api/kanban`.
- `src/ui/page/script.ts` (~L1688+) — front-end (add-to-todo + clear-done; no move).
- `src/ui/page/{template,styles}.ts` — the tab markup + styles.
- `src/commands/start.ts:554-558` — note listing kanban as part of the documented dashboard.

## Recommendation

Default to **A (remove)** unless there's a concrete operator use for **B**. If B, it should get a short SPEC first (which events populate it, and why a board beats the existing sessions/logs views) so we're building earned breadth, not a decorative tab.

Nibbler1250:
Resolves #294 — takes **option B** (wire it), the choice made over the issue's default "remove".

The Kanban tab shipped with working `addCardToColumn`/`moveCard` mutators and a card shape carrying `started_at`/`completed_at`/`agent_type` — but **zero callers**: a static sticky-note board wearing an agent-tracker's clothes. This wires it to the bus so it becomes a **live view of subagent activity**.

## What it does (V1 = subagent-only feed)
A global `BusCore` subscriber (`src/bus/kanban-tracker.ts`, structured like `StallWatchdog`):
- `response.tool_use` where name is `Agent`/`Task` → a card in **in_progress** (id = tool_use id, title = description/prompt, `started_at`, `agent_type`).
- matching `tool_result` → `moveCard(…, "done", { completed_at })`, then caps the `done` column to the newest 50.
- `session.end` / `session.init` → sweeps that session's still-open cards to **done (orphaned)** so the board never accumulates zombies (mirrors `StallWatchdog` clearing `outstanding`).

Global feed = covers cron/Telegram/Discord-triggered work, not just dashboard chats. Cron cards / per-session cards are deliberately **phase 2** (SPEC §out-of-scope).

## Design notes
- **Never throws into the bus**: `ingest()` is sync + try-wrapped; every kanban write is swallowed+logged.
- **Serialized writes (load-bearing)**: `addCardToColumn`/`moveCard`/`capDone` are read-modify-write on one JSON file, so concurrent bus events would lose updates — all writes chain through a single in-tracker promise queue.
- **Injected deps** (subscribe/addCard/moveCard/capDone/now/log) → unit-testable without disk or a real bus.
- **Gated** on `settings.kanban.enabled` (defaults ON; `{ enabled: false }` disables the feed, manual "+ Add task" unaffected).
- **No front-end changes**: the existing 10s GET poll + `renderKanbanCard` (already renders timestamps) pick up server-side moves.
- **Wiring**: mounted in `start.ts` after the bus+agents are up (deferred-spawn block), non-fatal; `stop()` in `shutdown()` and the rollback path.

## Tests / checks
- `src/bus/__tests__/kanban-tracker.test.ts` — **17 pass** (spawn→in_progress; finish→done+cap; non-Agent tool ignored; unknown result ignored; session.end + session.init orphan sweep; session isolation; agent_name fallback; idempotent dup spawn; **write serialization / no-lost-update**; ingest-never-throws; disabled → no subscribe; start/stop; `capDoneColumn`).
- Biome clean; `tsc` adds **no new errors** (start.ts stays at its pre-existing count). Version bumped.

SPEC: `.planning/kanban-live/SPEC.md`.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.183",
+      "version": "2.2.184",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.183",
+  "version": "2.2.184",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.planning/kanban-live/SPEC.md
+++ b/.planning/kanban-live/SPEC.md
@@ -1,0 +1,139 @@
+# SPEC — Live Kanban board of subagent activity (#294 option B)
+
+## 1. Problem statement
+
+The Web UI ships a **half-built Kanban board**: `src/ui/services/kanban.ts` defines the
+`KanbanBoard`/`KanbanCard` model, the `readKanban`/`writeKanban` persistence, and two
+mutators — `addCardToColumn(column, card)` and `moveCard(id, toColumn, patch?)` — but those
+two mutators have **zero callers**. The only writes today come from the manual `POST /api/kanban`
+"+ Add task" path (`src/ui/server.ts:825`). So the board is a static scratchpad, not a view of
+anything the daemon is actually doing. Issue #294 option B: make it a **live view of subagent
+(Task/Agent) activity** driven off the bus, reusing the mutators that already exist.
+
+## 2. Current behaviour (as-is, with refs)
+
+- **The mutators are unwired.** `addCardToColumn` / `moveCard` (`src/ui/services/kanban.ts:49,58`)
+  each do a read-modify-write of `.claude/claudeclaw/kanban.json`. Nothing in the codebase calls
+  them; the board only changes when a human POSTs a whole board to `/api/kanban`.
+- **The bus already observes every subagent spawn/finish.** The JSONL tailer emits
+  `response.tool_use` (`{id, name, input}`) and `tool_result` (`{tool_use_id, …}`) per session
+  (`src/bus/jsonl-tailer.ts`), plus `session.agent_name`, `session.init`, `session.end`.
+  `BusCore.subscribe(filter, handler)` (`src/bus/core.ts:897`) fans these to any subscriber.
+  A subagent dispatch surfaces as a `response.tool_use` with `name === "Agent"` (or `"Task"`),
+  and its completion as the matching `tool_result` (`tool_use_id === id`).
+- **The proven signal, but per-chat only.** `src/ui/server.ts:778` already maps the runner's
+  `AgentStreamEvent` to SSE `agent_spawn`/`agent_done` — but only for the single interactive
+  chat runner, not globally. The bus path sees *every* session, which is what a board wants.
+- **No serialization guard exists** on the two file mutators: two near-simultaneous bus events
+  would each read-modify-write the same JSON and lose one update.
+
+## 3. Target behaviour (to-be)
+
+A new **`KanbanTracker`** (`src/bus/kanban-tracker.ts`), a pure `BusCore` subscriber (mirrors
+`StallWatchdog`'s structure/quality), that turns subagent lifecycle into board cards:
+
+1. **Spawn → card in `in_progress`.** On `response.tool_use` where `payload.name` is `"Agent"`
+   or `"Task"`: create `{ id: payload.id, title: (input.description ?? input.prompt ?? "subagent").slice(0,120),
+   started_at: ISO(ts), agent_type }` and `addCard("in_progress", card)`. The tool_use `id` is
+   tracked in an in-memory live set keyed by id, and remembered against its session.
+2. **Finish → card to `done`.** On `tool_result` whose `tool_use_id` matches a **tracked** live
+   agent card: `moveCard(id, "done", { completed_at: ISO(ts) })` and drop it from the live set.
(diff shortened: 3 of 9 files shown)


Nibbler1250:
Validated live end-to-end on a running daemon (not just the unit suite): deployed the tracker + wiring onto a live instance, restarted it clean, then injected a prompt that made the agent dispatch a `Task` subagent. Result in `kanban.json`:

```json
{ "id": "toolu_…", "title": "Count .ts files in src/bus/",
  "started_at": "…", "agent_type": "general-purpose", "completed_at": "…" }
```

The full lifecycle fired against real bus events: `response.tool_use` (Task) → card created in `in_progress` with `started_at` + `agent_type` resolved from the tool's `subagent_type`; the matching `tool_result` → moved to `done` with `completed_at`; `kanban.json` created by the tracker (didn't exist before). No wiring bug — the daemon booted clean with the tracker mounted, and the board populated with zero manual steps. Reverted the instance to its prior state afterward.

TerrysPOV (changes requested):
Reviewed the commit. Nicely built and lightweight in the way it should be: the `KanbanTracker` ingests bus events (spawn → in-progress, tool_result → done), the board is polling-only (10s, only while the Kanban tab is active — no SSE/per-client connection), writes are serialized through an in-tracker mutex, the `done` column is capped, and `session.end`/`session.init` correctly sweep orphaned in-progress cards so a restart/rotation doesn't leave zombies. The tracker has comprehensive unit tests. Read-side recovers from a corrupt/missing `kanban.json` (returns the default board), and `ingest()` never throws into the bus.

One blocker:

**`saveKanban()` uses plain `fetch`, not `mutatingFetch`, so its POST has no CSRF token and the server 403s it.** (`src/ui/page/script.ts` ~L1752; server `POST /api/kanban` calls `requireCsrf` at `src/ui/server.ts` ~L826.) Every other state-changing call in `script.ts` uses `mutatingFetch` (heartbeat toggle, job fire/delete, quick-job, LLM-router save) — this one was missed. And because `saveKanban` swallows all errors (`catch (_) {}`), the failure is silent: clicking "add task" or "clear done" updates the UI locally but never persists, and the user gets no error. One-line fix — swap `fetch` → `mutatingFetch`.

Two non-blocking notes:
- `POST /api/kanban` writes `req.json()` cast straight to `KanbanBoard` with no structural validation, so an authenticated `{}` or malformed body corrupts `kanban.json`. Pre-existing pattern (the read side recovers to the default board), self-inflicted only, so not a blocker — but a small shape check before `writeKanban` would be tidy.
- Latent XSS trap: `renderKanbanCard` correctly escapes `title`/`description`/`id` via `escKanban`, but `agent_type` is stored and never rendered. If a future change renders it without `escKanban`, it's an injection vector (subagent type / agent name is content-influenceable). No action now — just flagging for whenever `agent_type` gets rendered.

Vision: PARTIAL — a genuinely useful operator view for multi-agent orchestration, marginal for single-agent users; implementation stays lightweight (no new deps/process, bounded memory), so no real objection.

Conflicting with current main, so it'll want a rebase. Fix the `saveKanban` CSRF one-liner and this is good to go.

TerrysPOV (approved):
Maintainer fix applied — approving (supersedes my change-request). Fixed the one blocker in `acfd`-style maintainer commit: `saveKanban` now uses `mutatingFetch` so its POST carries the CSRF token (raw fetch was 403'ing and silently dropping the write). I verified `mutatingFetch` and `saveKanban` share the top-level script scope (empirically — no wrapping IIFE — so it's a safe direct swap, not a ReferenceError).

The rest stands as reviewed: lightweight polling, session-sweep orphaning, serialized writes, comprehensive tracker tests. The two non-blocking notes (POST body has no structural validation — pre-existing pattern, read-side recovers; and the latent `agent_type` XSS trap if it's ever rendered) are left as-is. Rebased onto current main + bumped (2.2.184); biome + build + kanban tests green. Good to merge.

Issue #227, opened 2026-06-04
bus session rotation: restart-based mechanism (4a follow-up to #218)


Nibbler1250:
Follow-up to #218 (split per the review there): #218 lands the per-agent **bookkeeping**; this issue tracks the actual rotation **mechanism** for the long-lived bus PTY.

## Why the bookkeeping alone isn't enough

The #218 review surfaced that filesystem-only rotation (write summary, rename `session.json`, unlink) doesn't rotate the live conversation. The bus spawns the agent PTY once at boot with `--session-id <uuid>` and keeps it alive, so the session JSONL keeps growing and per-prompt latency creeps up regardless. Three criticals all trace back to "the PTY is never restarted":

1. **Rotation doesn't touch the live PTY** → #213's "JSONL grows unbounded on bus mode" isn't actually fixed; the next prompt routes to the same live PTY/process/context.
2. **Rotation is one-shot per daemon lifetime** — after `resetSession` unlinks `session.json`, the gate never re-arms and `incrementMessageCount` no-ops (`if (!existing) return;`), so the counter stays at 0 from rotation onward.
3. **Session-id collision in `generateSummary`** — `claude -p --resume <live-session-id>` against the locked live session fails silently (swallowed by the surrounding try/catch).

## Direction: 4a (maintainer preference)

Restart the PTY after `resetSession` so the next `streamBusPrompt` lands on a fresh PTY with a new session id; the rotation summary becomes the bootstrap context for the new PTY. Concrete shape:

1. `generateSummary` runs **after** the restart, against the backed-up `session.json`'s saved session id (no longer the live one) — or summarizes from a JSONL copied aside. Summary input must never be the live session id (fixes critical 3).
2. After `resetSession(agentId)`, call `sessionManager.restart(agentId)`: drop the PTY, mint a new uuid via `defaultResolveSessionId` (writes `session.json`), spawn a fresh PTY. **This is the `restart()` primitive in #226.**
3. The new PTY's bootstrap prompt includes the summary (compressed-context continuation).
4. `incrementMessageCount(agentId)` after restart finds the freshly-created `session.json` (fixes critical 2).
5. Tests: bookkeeping fires; rotation triggers at threshold; PTY restart invoked; new `session.json` written; summary consumed by the new PTY's bootstrap.

## Depends on

[...]

PR #300, opened 2026-07-04
feat(bus): activate the stall watchdog — forensics + config + wiring (completes PR 1/3, #296)

TerrysPOV:
Completes **PR 1/3 of #296** — the stall watchdog that fixes the #295 outage (a bus PTY session wedged ~11h on a hung `Bash` tool call). #297 landed the detection core as dead code; **this PR wires it in — it is the commit that actually makes the watchdog run.**

> ⚠️ **Merging this activates the watchdog on the daemon** (and deploys). It ships **enabled** by default. That's intended, but reviewers should know the stakes are "live behaviour change", not "more dead code".

@Nibbler1250 — requesting your review again (thanks for the #297 pass; all of it is folded in here).

## What's in this PR
- **`src/bus/stall-forensics.ts`** (already reviewed as the first commit) — `probeProcessTreeCpu` (best-effort `/proc` CPU-tree sampler, `null` off-Linux), pure `classifyKill`, `appendStallKillAudit`. 9 tests.
- **`src/config.ts`** — `parseStallWatchdogConfig` (per-field validation + clamps, falls back to `DEFAULT_STALL_CONFIG`) wired into `Settings` / `DEFAULT_SETTINGS` / `parseSettings`. Ships **enabled**; every ceiling + cooldown overridable in `settings.json`.
- **`src/bus/session-agent-process.ts`** — track `lastDataAt` on both `PtyAgentProcess` and `ChildAgentProcess` (the forensic "output recency" liveness signal). `pid` was already exposed.
- **`src/bus/runtime-mount.ts`** — instantiate `StallWatchdog` after a *successful* mount (so a failed mount never leaves a dangling sweep timer), binding the real deps:
  - `subscribe` → `BusCore.subscribe` (topics: tool_use / tool_result / turn_end / init / end)
  - `restart` → `SessionManager.restart({ reason: "stall" })` — **inherits `MAX_RESTARTS_PER_WINDOW`**, so a recurring stall can't respawn-loop
  - `captureForensic` → `probeProcessTreeCpu(pid)` + `lastDataAt`
  - `recordKill` → `classifyKill` + `appendStallKillAudit`
  - `notify` → structured `logger` (see follow-up below)
  - torn down first in `stop()`.

## #297 review follow-up — addressed
The `@claude` review's deferred note (`restartFailedAt` reset invariant) is now **implemented behaviour + tested**: a re-observed session kills cleanly after a failed→(cooldown)→succeeded restart cycle (no leaked cooldown). See `stall-watchdog.test.ts`.

## Tests
64 stall/forensics/mount tests green; `session-agent-process` (20) and config (81) unaffected; `bun run lint` clean. New: `parseStallWatchdogConfig` validation, the `restartFailedAt` invariant, and the mount test updated (mount now starts exactly one subscriber — the watchdog).

## Security posture (self-review)
Minimal surface: reads `/proc/<pid>/stat` (pid is an integer from `getAgent`, read-only, no injection); appends a `JSON.stringify`'d record to `.claude/claudeclaw/stall-kills.jsonl`; recovery reuses the existing `SessionManager.restart`. No shell exec, no user-input interpolation, no secrets. All best-effort paths swallow errors so forensics/audit can never block recovery.

## Known follow-up (not in this PR) — tracked in #301
`notify` currently routes to the structured `logger` (operator-visible in daemon logs). **Delivering the false-positive flag to Discord** is a deliberate follow-up, tracked in **#301** (same shape as #227's deferred summary-injection) — the audit trail + logs capture it today; the chat surfacing needs the operator-alert path.

Refs: #296 (epic), #295 (incident), #297 (detection core, merged).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.170",
+      "version": "2.2.171",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.169",
+      "version": "2.2.170",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.169",
+  "version": "2.2.170",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.170",
+  "version": "2.2.171",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/runtime-mount.test.ts
+++ b/src/bus/__tests__/runtime-mount.test.ts
@@ -315,8 +315,9 @@ describe("mountBusRuntime — real UDS path", () => {
     try {
       expect(handle.bus).toBeDefined();
       expect(handle.sessionManager).toBeInstanceOf(SessionManager);
-      // BusCore.state() works post-mount.
-      expect(handle.bus.state().subscriberCount).toBe(0);
+      // BusCore.state() works post-mount. The one subscriber is the stall
+      // watchdog's session-event subscription, started by mountBusRuntime (#296).
+      expect(handle.bus.state().subscriberCount).toBe(1);
     } finally {
       await handle.stop();
     }
--- a/src/bus/__tests__/stall-forensics.test.ts
+++ b/src/bus/__tests__/stall-forensics.test.ts
@@ -1,0 +1,114 @@
+import { describe, it, expect } from "bun:test";
+import { mkdtempSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import {
+  classifyKill,
+  probeProcessTreeCpu,
+  appendStallKillAudit,
+  RECENT_OUTPUT_MS,
+  type StallKillAuditRecord,
+} from "../stall-forensics";
+import type { ForensicSnapshot, ToolCeiling } from "../stall-watchdog";
+
+const BASH: ToolCeiling = { warnSeconds: 300, killSeconds: 900 };
+const snap = (cpuAdvancing: boolean | null, outputRecencyMs: number | null): ForensicSnapshot => ({
+  cpuAdvancing,
+  outputRecencyMs,
+});
+
+/* ── classifyKill truth table ──────────────────────────────────────────── */
+
+describe("classifyKill", () => {
+  it("CPU advancing → suspected_false_positive", () => {
+    const o = classifyKill(snap(true, null), 950_000, BASH);
+    expect(o.classification).toBe("suspected_false_positive");
+    expect(o.suggestedKillSeconds).toBeDefined();
+  });
+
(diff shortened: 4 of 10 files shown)


Nibbler1250 (commented):
Reviewed the full PR (thanks for folding the #297 pass in). **Verdict: no blocking defects — safe to ship enabled.** I traced every path I flagged as risky in #297 as "the part that lands later", and they all hold up.

### Verified correct
- **`stall-forensics.ts` — clean.** `/proc/<pid>/stat` parsing splits on the **last** `)` — the right fix for a `comm` containing spaces/parens; `utime`/`stime` indices are correct. Every parse failure (missing `)`, truncated line, non-numeric, **pid that dies mid-probe**, missing root) funnels to `null` — the probe path never throws (tests cover dead-pid + live-pid). Tree walk is a `Set`-guarded iterative stack, so a ppid cycle / huge tree can't loop forever.
- **`classifyKill` truth-table — correct** for all 6 `cpuAdvancing{true,false,null} × outputRecency` combos. Critically, **probe-unavailable (`null`) + no recent output → `unknown`, not `genuine_wedge`** — so an unconfirmable kill doesn't silently suppress the operator flag. `suggestedKillSeconds` is sane (≥2×kill, whole-minute, no NaN/negative).
- **`appendStallKillAudit`** swallows all errors → auditing can't block recovery.
- **`parseStallWatchdogConfig` — boot-safe.** Pure per-field validation with `?? {}` fallbacks; non-object → `structuredClone(default)`; a garbage per-tool ceiling falls back to *that* ceiling only. `mountBusRuntime` also wraps `getSettings()` in try/catch → a malformed `stallWatchdog` block cannot break daemon boot.
- **`runtime-mount.ts` — correct.** Watchdog constructed+started only *after* the successful-mount point (no dangling sweep timer on a failed mount); exactly **one** subscriber added; the bus `tool_use`/`tool_result` payload field names match what `ingest` reads (so outstanding-tool tracking clears — no false kill from a field mismatch); `stop()` clears the timer + unsubscribes, called first in teardown.
- **`session-agent-process.ts`** — `lastDataAt` stamped per data chunk on both process types, exposed via getter; a genuinely good wedge signal (blocked loop stops repainting → stale; live-but-slow tool keeps repainting → recent). Hot-path cost = one `Date.now()`.
- **`restartFailedAt` invariant — holds.** A successful restart deletes the whole session state, so the cooldown can't leak into a later stall; a failed restart bounds the backoff (one alert per cooldown window).

### Non-blocking notes (ranked, none justify holding the merge)
1. **(Low, pre-existing from the detection core — not this PR) `handleKill`'s `decision` param resolves to `never`.** `Extract<StallDecision, {action:"kill"}>` filters out the only data-bearing member because that member declares `action: "warn" | "kill"` jointly, so `tsc` can't check the function body. `tsc` isn't the CI gate and the runtime value is always a kill, so behaviour is fine — but splitting the union into separate `warn`/`kill` members would restore type-safety. Follow-up, not this PR.
2. **(Low) No `warn ≤ kill` clamp.** A misconfig with `warn > kill` silently makes the warn dead. Not harmful; a clamp or a log would catch the footgun.
3. **(Low) `restartFailureCooldownMs` min is `0`** → a user setting `0` re-alerts every sweep (the storm the cooldown exists to prevent). Consider flooring at `sweepIntervalMs`.
4. **(Low) `DEFAULT_SETTINGS.stallWatchdog` holds a live reference to `DEFAULT_STALL_CONFIG`.** Nothing mutates it today; a `structuredClone` at the default site would harden against a future in-place edit.
5. **(Info) The new suites aren't in the CI gate.** CI runs only the one watchdog test + biome; the 64 new stall/forensics/mount tests (all green locally) aren't executed — worth adding for something enabled-by-default.
6. **(Very low) A hung `deps.restart()`** would pin the session in the `killing` set; a hard timeout around the restart await would make it robust. Degrade-safe today.

Happy to fold #2–#4 (the cheap config-hardening) in as a maintainer commit if that's useful to you — they're 5-minute changes with no happy-path behaviour change. #1 and #5 I'd file as their own follow-ups. Your call. Nice work — this is a clean activation.

Nibbler1250 (approved):

TerrysPOV:
Yes pls fold in as maintainer commits and I will get on with reviewing the Tuner stack

Nibbler1250:
Folded in the config-hardening notes (#2–#4) as a maintainer commit (`679744d`):

- **`warn ≤ kill` clamp** in `parseCeiling` — a `warn > kill` misconfig now pins to the kill line instead of a silently-dead warn.
- **`restartFailureCooldownMs` floored at `sweepIntervalMs`** — a sub-sweep cooldown (e.g. `0`) would re-kill + re-alert every sweep; it now falls back to the default.
- **`structuredClone(DEFAULT_STALL_CONFIG)`** at the `DEFAULT_SETTINGS` site so a future in-place settings edit can't mutate the shared exported default.

+2 tests (clamp; sub-sweep cooldown → default). 28 stall tests green, biome-clean, version bumped.

Left for their own follow-ups (as flagged): **note #1** (the pre-existing `Extract<…,{action:"kill"}>` → `never` on `handleKill` — a union-split fix that would restore type-checking of that function; pre-existing from the detection core, so I kept it out of this PR per "don't fix pre-existing in the same diff") and **note #5** (adding the new stall/forensics/mount suites to the CI test step). Happy to open issues for both, or send a small PR for #1 if you'd like — your call. Otherwise this is good to go from my side.

TerrysPOV:
@claude review

PR #302, opened 2026-07-04
feat(hooks): block unsafe inline Bash backgrounding — PreToolUse guard (PR 2/3, #296)

TerrysPOV:
**PR 2/3 of #296.** PR 1 (the stall watchdog, #297 + #300) is the *recovery* net; this is the *prevention* — it stops the #295 root cause at source.

@Nibbler1250 — requesting your review.

## What it does
A `PreToolUse(Bash)` hook that **denies** an unsafe inline background job before it runs and steers the agent to the pipe-safe primitive.

The #295 outage: a `Bash` call ran `nohup … &`; the backgrounded process held the tool's output pipe open, so a later `Bash` call never saw EOF and the single-threaded session wedged for ~11h.

## Why a hook, not a timeout (and why block, not rewrite)
- **Not a timeout:** I verified against the Claude Code docs that `BASH_*_TIMEOUT_MS` bounds the *foreground command*, **not** the tool's pipe-drain — a backgrounded child holding fd1/fd2 keeps the pipe open past the timeout. So the timeout lever provably wouldn't have caught this hang.
- **Block, not rewrite** (Terry's call): Claude Code hooks *can* rewrite (`updatedInput`), but safely rewriting arbitrary shell is fragile, and a blind detach-to-`/dev/null` would destroy output the agent wants (Claw specifically wanted to `tail` its job's log). Block-and-steer teaches the correct primitive (`run_in_background: true`), which is exactly what was missing.

## How it decides
`isUnsafeBackground(command)` (pure, unit-tested): strips quoted strings, neutralises tokens that legitimately contain `&` (`&&`, fd-dups `2>&1`/`>&2`, redirect-all `&>file`), then flags a remaining lone background `&`. **Safe forms pass untouched:** `&&`, `cmd & wait`, `setsid … </dev/null … &`, and anything with no background operator. On a hit it returns `permissionDecision: "deny"` + guidance (use `run_in_background: true`, or fully detach with `setsid <cmd> </dev/null >log 2>&1 &`).

**Fails OPEN:** any parse/logic error allows the command — a guard must never break the tool it protects.

It's a heuristic (a zero-dep hook can't fully parse shell), tuned to catch the common footgun (`nohup … &`, `cmd &`, `cmd & echo`) while erring toward allowing on ambiguity — PR 1's stall watchdog remains the backstop for anything that slips through.

## Files
- `hooks/guard-bash-backgrounding.mjs` — the hook (zero-dep Node, mirrors #289's `.mjs` convention).
- `hooks/hooks.json` — wires the `PreToolUse` Bash matcher.
- `src/__tests__/bash-backgrounding-guard.test.ts` — 9 tests: truth table (incl. quote/URL false-positive guards + the `wait`/`setsid` safe forms), deny/allow integration, fail-open on malformed input.

## ⚠️ Merge overlap
**#289 (still open) also edits `hooks/hooks.json`** (adds a `PostToolUse(Read)` entry). Whichever merges second must **merge the JSON** (keep both hook blocks), not clobber. Trivial, just flagging so it isn't lost.

## Review asks
- Is the `isUnsafeBackground` heuristic the right balance — catching the footgun without annoying false positives? Any common safe pattern I'd wrongly block?
- The block-vs-rewrite call — agree block is the right posture here?

Refs: #296 (epic), #295 (incident), #297/#300 (PR 1).


TerrysPOV:
@claude review

Nibbler1250 (commented):
Reviewed post-merge (thanks for the ping). **The heuristic strikes the right balance and the block-not-rewrite call is correct** — rewriting arbitrary shell is fragile, and block+steer teaches `run_in_background` which is the actual fix. The footgun is caught where it matters: `nohup cmd &`, `cmd &`, `cmd 2>&1 &` (still holds fd1 → correctly flagged), `( … ) &`, `cmd & disown` all deny; `&&`, fd-dups, quoted `&`, `cmd & wait`, and the `setsid … </dev/null >log 2>&1 &` form all pass. Fail-open + the stall watchdog backstop is the right safety posture for a heuristic.

Two things, both non-blocking:

**1. Safe-form-2 is stricter than pipe-safety needs → a mild false-positive.** The #295 hazard is specifically a backgrounded child holding **fd1/fd2** (the tool's output pipe). That's neutralised by redirecting stdout+stderr off the pipe — `setsid` and `</dev/null` (fd0) aren't part of *that* hazard. So a genuinely pipe-safe command like `cmd >log 2>&1 &` or `nohup cmd >log 2>&1 </dev/null &` gets **blocked** today because the exemption requires `setsid` + `stdin</dev/null`. If minimising false positives matters, the exemption could key on `stdoutRedir && stderrRedir` alone. If the stricter canonical form is an intentional best-practice nudge (accept the small false-positive to teach the robust pattern), that's a defensible call too — just flagging the tradeoff. Either way the deny message hands them the accepted form, so it's recoverable.

**2. Quote-stripping means a backgrounded job inside a quoted `-c` string slips through — and still holds the pipe.** `bash -c "sleep 1000 &"` (or `sh -c '… &'`) strips to `bash -c ""`, so no `&` remains → allowed, yet the inner background inherits fd1/fd2 and can wedge exactly like #295. Stripping quotes is the *right* call for avoiding URL/string false positives, so this is an inherent limitation of a zero-dep heuristic, not a bug — I'd just add it to the SPEC as a known gap that the stall watchdog (PR 1) is the backstop for, and maybe a test asserting the `-c` case is knowingly allowed so a future reader doesn't think it's covered.

**Merge-overlap note you flagged (#289 also edits `hooks/hooks.json`)** — noted; when #289 comes up I'll merge the JSON (keep both hook blocks), not clobber.

Nice piece — prevention + recovery is the right shape for #296.

PR #303, opened 2026-07-05
feat(bus): first-class agent-job primitive — dispatch_job (PR 3/3 of #296)

TerrysPOV:
## PR 3/3 of #296 — first-class agent-job primitive (root-smell fix)

Epic **#296** is the 3-PR fix for the #295 daemon outage (an agent hand-rolled a `/tmp` fire-script that `Bun.spawn(["claude","-p",…])` + `nohup … &`, holding the Bash tool's pipe and wedging the session for ~11h).

- **PR 1 — stall watchdog (recovery):** merged (#297, #300).
- **PR 2 — Bash-exec guard (prevention):** merged (#302).
- **PR 3 — agent-job primitive (this PR):** removes the *need* to improvise jobs at all.

### What it does

Gives agents a supported `dispatch_job` tool so one agent can fire a long background job to another named agent (`reg`/`suzy` research/draft runs), get a `job_id` back **immediately**, and have the result delivered back when it finishes — instead of shelling out to `claude -p` by hand.

- `dispatch_job({ agent, prompt, model?, timeoutMs? })` → `{ job_id, status }` immediately (fire-and-return; never blocks the caller's turn — that awaiting is exactly what wedged #295).
- `job_status` / `list_jobs` / `cancel_job` for observability + control.
- The finished job's result is delivered back to the **dispatching** agent (via the bus) and stays queryable.

### Architecture (why it's wired this way)

The `AgentJobRunner` lives in the **daemon** — only it can spawn/track the headless `claude -p` job processes. The four tools run in the per-agent **MCP subprocess**, which reaches the daemon over the existing unix socket. So each tool call routes over IPC as one correlated request/response (`IpcJobRequest`/`IpcJobResult`), mirroring the existing `request_human` round-trip. Fire-and-return, in-memory registry (no cross-restart persistence — matches the scheduler + restart-rotation model).

### Security posture

The daemon spawns agent processes in response to an IPC message, so this went through a dedicated security review. Confirmed sound and hardened:

- **No dispatcher spoofing** — the job's `dispatcher` is stamped from the socket-authenticated `agent_id`, never a client-supplied value.
- **No shell injection** — `Bun.spawn(args, …)` uses an args array (no shell); prompt/model are separate argv elements; the executable is fixed.
- **Sandbox escape closed (HIGH, fixed)** — `agentExists` now requires the kebab-case name charset before the `existsSync` check, so `agent:"../../etc"` can't spawn `claude -p` with cwd outside `agents/`.
- **Model allow-listed** — caller model validated (`validateModelString`) before spawn; no silent tier escalation.
- **Bounded** — caller-uninfluenceable concurrency cap + hard timeout cap; queue-depth cap + prompt-length cap + oldest-terminal record eviction (a looping `dispatch_job` can't OOM the daemon).
- **No turn-wedge** — the MCP round-trip is bounded by a 30 s timeout, and the daemon always sends exactly one `job_result`.

### Test plan

- `src/bus/__tests__/agent-jobs.test.ts` — registry transitions, fire-and-return, concurrency/queue caps, prompt-length cap, retention eviction, cancel/stop.
- `src/bus/__tests__/agent-jobs-ipc.test.ts` — real-UDS socket round-trip: dispatch/status/list/cancel, result delivery, unknown-agent + not-enabled paths.
- `src/__tests__/agents.test.ts` — `agentExists` path-traversal rejection.
- Full bus suite green.

Refs #296, #295.

https://claude.ai/code/session_01RJZianxrwQWLhYGmALm7L9

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.174",
+      "version": "2.2.175",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.172",
+      "version": "2.2.173",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.173",
+      "version": "2.2.174",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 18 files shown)


TerrysPOV:
@claude /review

Issue #304, opened 2026-07-05
Test hygiene: make root src/__tests__/ CI-able (pollution + infra deps), then widen the CI Test scope


TerrysPOV:
## Context

CI now runs `bun test` on the clean module suites (`src/bus`, `src/tuner`, `src/observability`, `src/skills-tuner`, `src/governance` — ~700 tests, all green) via the new `Test` workflow. The root `src/__tests__/` suite is **excluded** because a full `bun test` on main is **2518 pass / 41 fail**, and all 41 failures live in `src/__tests__/`.

## The 41 failures split two ways

1. **Cross-test pollution** — tests that pass in isolation but fail in the shared process (e.g. `Phase 18: runner modelOverride wiring` passes alone, fails in the pack). Root cause is shared module/singleton/env/file state between test files. This is the same class the SDC `mark-tests-passing.sh` works around with `TEST_ARGS` scoping.
2. **Infra-dependent** — tests that fail even in isolation because they need a DB/queue/service not present in CI (e.g. `Event Processor` → 9 fail alone).

Representative failing files: `event-processor`, `jobs`, `escalation/*`, `gateway/*`, `policy/*`, `mcp_proxy_*`, `pty-*`, `runner`, `integration/*`.

## Goal

Make `src/__tests__/` green (fix pollution via proper per-test isolation/teardown; gate or mock the infra-dependent tests), then **widen the CI `Test` job's paths** toward the full `bun test`. Landmark: CI runs the whole suite green.

## Why it matters

The reason this surfaced: PR #290's 3 hanging `collectObservations` tests showed **green in CI** because CI ran no tests at all. The scoped `Test` job catches that class in the module dirs now; closing this issue extends the safety net to the whole codebase.

TerrysPOV:
## Re-baselined at `36321ad` — the original diagnosis was wrong, and the scope is smaller than it looks

Full `bun test` on current main: **2893 pass / 6 skip / 41 fail across 2940 tests in 180 files.** The failure count is unchanged since this issue was filed, but the *composition* is not what the issue describes.

### Finding 1 — all failures live in 10 files, not scattered

| Failures | File |
|---|---|
| 9 | `src/__tests__/event-processor.test.ts` |
| 9 | `src/__tests__/runner.test.ts` |
| 7 | `src/__tests__/mcp-multiplexer-integration.test.ts` |
| 7 | `src/__tests__/session-persistence-integration.test.ts` |
[...]

PR #305, opened 2026-07-05
ci: run bun test on the clean module suites

TerrysPOV:
## What

Adds a **`Test`** GitHub Actions job (`.github/workflows/test.yml`) that runs `bun test` on the module suites — `src/bus`, `src/tuner`, `src/observability`, `src/skills-tuner`, `src/governance` (~700 tests).

## Why

Until now CI ran **no unit tests** — the "Bun Smoke Test" only does a build/parse/type-check. That's how PR #290's 3 hanging `collectObservations` tests showed green. This wires the tests into CI so that class of regression is caught.

## Scope (deliberately not the full suite yet)

A full `bun test` on main is **2518 pass / 41 fail**, and all 41 failures are in root `src/__tests__/` — a mix of cross-test pollution (pass in isolation, fail in the shared process) and infra-dependent tests (DB/queue not in CI). Folding those in would make CI permanently red. The scoped module suites are pollution-free and green (700/0, ~2 min).

Cleanup of `src/__tests__/` and widening the scope toward the full suite is tracked in #304.

Refs #304.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.179",
+      "version": "2.2.180",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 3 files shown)


TerrysPOV:
Addressed the Codex P2 in `e98be14`: `src/governance` holds no `*.test.ts` (bun treats the args as path patterns, so it silently matched nothing), so governance regressions wouldn't have failed the job. Repointed to `src/__tests__/governance` — the real governance tests (77) pass cleanly when scoped, and the combined run stays green (822 pass, 0 fail). Comment updated to spell out the path-pattern gotcha.

PR #307, opened 2026-07-06
fix(tuner): model-routing YAML editors leak into sibling modes (#306)

### LINKED ISSUE #306: bug(tuner): model-routing YAML edits corrupt sibling modes (inMode never resets on indented modes)
## Summary

The three YAML-edit helpers in `src/tuner/subjects/model-routing-subject.ts` — `swapModelInYaml`, `removeKeywordFromYaml`, `renameKeywordInYaml` — only exit the target mode's block when a line starts at **column 0**:

```ts
if (inMode && /^\S/.test(line)) inMode = false;
```

In a normal indented `modes:` block, sibling mode headers are indented (e.g. `  slow:`), so `/^\S/` never matches and `inMode` never resets. Everything after the target mode — including **sibling modes** — is then edited as if still inside the target mode.

Shipped via #287 (the model-routing reactive face), so it is **already in `main`**. Surfaced during the review of the proactive face (PR #292).

## Impact

This is the model-routing **apply path**. Applying a routing change to a multi-mode `agentic.yaml` silently rewrites sibling modes' models. The corrupt output still passes `validate()` and `healthProbe()`, so nothing catches it. Since the tuner is opt-in and this only bites multi-mode configs, severity is medium — but it's silent config corruption on a governed apply, which is exactly what the confinement/validation story is meant to prevent.

## Reproduction

Given:
```yaml
modes:
  fast:
    model: claude-sonnet
  slow:
    model: claude-opus
```
`swapModelInYaml(content, "fast")` produces:
```yaml
modes:
  fast:
    model: claude-haiku   # correct
  slow:
    model: claude-sonnet  # WRONG — slow's model swapped too
```

## Suggested fix

Capture the target mode header's indentation when `inMode` is set, and reset `inMode` when a subsequent non-blank line appears at the same-or-lesser indentation:

```ts
let inMode = false;
let modeIndent = "";
for (const line of lines) {
  const m = line.match(new RegExp(`^(\\s*)${escapeRe(mode)}:\\s*$`));
  if (m) { inMode = true; modeIndent = m[1]; out.push(line); continue; }
  if (inMode && line.trim().length > 0 && !line.startsWith(modeIndent + " ")) inMode = false;
  // ...existing edits...
}
```

Apply to all three helpers, and add a regression test with a **two-mode** block asserting the sibling mode is untouched (none of the current tests cover the multi-mode case, which is why it's undetected).

Locations: `removeKeywordFromYaml` (~L567–588), `renameKeywordInYaml` (~L590–619), `swapModelInYaml` (~L621–641) in `src/tuner/subjects/model-routing-subject.ts`.

cc @Nibbler1250

Nibbler1250:
## What

Fixes the sibling-mode corruption in the three line-based YAML editors in `src/tuner/subjects/model-routing-subject.ts` — `swapModelInYaml`, `removeKeywordFromYaml`, `renameKeywordInYaml`.

Closes #306.

## The bug

Each editor only exited the target mode's block on a **column-0** line:

```ts
if (inMode && /^\S/.test(line)) inMode = false;
```

In a normal indented `modes:` block the sibling mode headers are indented (`  slow:`), so `/^\S/` never matches, `inMode` never resets, and every edit after the target mode is applied to the **sibling modes** too. It's the governed **apply path**, and the corrupt output still passes `validate()` + `healthProbe()`, so nothing downstream catches it — silent config corruption, exactly what the confinement/validation story exists to prevent.

## The fix

Capture the target header's indentation when `inMode` is set, and exit the block on the first non-blank line indented **no deeper than the header** (a sibling mode or a dedent):

```ts
const modeHeader = line.match(new RegExp(`^(\\s*)${escapeRe(mode)}:\\s*$`));
if (modeHeader) { inMode = true; modeIndent = modeHeader[1] ?? ""; out.push(line); continue; }
if (inMode && line.trim().length > 0 && !line.startsWith(`${modeIndent} `)) inMode = false;
```

Applied identically to all three editors. Also unifies the two header regexes into one that captures the indent.

## Test

Adds a **two-mode** regression (`fast` + `slow` sharing a `quick` keyword) asserting all three alternatives from `proposeChange` leave the sibling `slow` mode untouched — its model, and its shared keyword, survive.

Verified the test **fails on the pre-fix code** (the swap-model alternative rewrote `slow`'s `claude-opus` → `claude-sonnet`) and **passes after**. Full `model-routing-subject` suite: **29 pass / 0 fail**. `bun run lint` clean (no new errors), versions bumped.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.185",
+      "version": "2.2.186",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.185",
+  "version": "2.2.186",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/tuner/__tests__/wisecron/subjects/model-routing-subject.test.ts
+++ b/src/tuner/__tests__/wisecron/subjects/model-routing-subject.test.ts
@@ -463,3 +463,63 @@ describe("ModelRoutingSubject — healthProbe (post-apply artifact check)", () =
     expect(probe.failed).toBe(false);
   });
 });
+
+describe("ModelRoutingSubject — YAML editors don't corrupt sibling modes (#306)", () => {
+  it("proposeChange edits only the target mode in a multi-mode block", async () => {
+    const yaml = [
+      "modes:",
+      "  fast:",
+      "    model: claude-sonnet",
+      "    keywords:",
+      "      - quick",
+      "      - rapid",
+      "  slow:",
+      "    model: claude-opus",
+      "    keywords:",
+      "      - quick",
+      "      - deep",
+    ].join("\n");
+    writeFileSync(configPath, yaml, "utf8");
+    const s = new ModelRoutingSubject({ modesConfigPath: configPath });
+    const obs: Observation[] = [
+      {
+        session_id: "t",
+        observed_at: new Date(),
+        signal_type: "orphan",
+        verbatim: "{}",
+        metadata: {
+          subject: "model_routing",
+          mode: "fast",
+          keyword: "quick",
+          triggers: 0,
+          reclassify_rate: 0,
+          expensive: false,
+          age_days: 200,
+        },
+      },
+    ];
+    const clusters = await s.detectProblems(obs);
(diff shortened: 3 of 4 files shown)


TerrysPOV:
@claude /review

TerrysPOV (approved):
Approving. This fixes #306 (the sibling-mode YAML corruption I filed after reviewing #292), and the fix is exactly right.

Confirmed the `modeIndent`-capturing exit condition is applied identically across all three editors (`removeKeywordFromYaml`, `renameKeywordInYaml`, `swapModelInYaml`): a sibling header at the same indent exits the block, deeper-nested lines stay in, blank lines are skipped, and `modeIndent === ""` reduces to the original column-0 behaviour so single-mode configs are untouched. The regression test (two-mode `fast`/`slow` with a shared keyword, asserting all three `proposeChange` alternatives leave the sibling untouched) is the right shape and passes. Full model-routing suite green (29/0); Claude Bot's CI review also LGTM'd it.

Maintainer commit from me: rebased onto current main + bumped to 2.2.185 (the branch was level with main, so the version guard needed it).

One non-blocking follow-up (Claude Bot flagged it too): a comment line indented at *exactly* `modeIndent` strictly inside a mode block would prematurely close it — low risk since hand-written configs rarely do that, but a `&& !line.trim().startsWith("#")` guard on the exit condition would close it fully if we ever see it in the wild. Not worth holding this. Good to merge.

Issue #178, opened 2026-05-27
fix(deploy): systemd cgroup hardening is bypassed by su re-parenting


TerrysPOV:
## Problem

The Hetzner systemd unit `/etc/systemd/system/claudeclaw.service` declares Tier 1 hardening:

```
TasksMax=50
MemoryMax=2G
CPUQuota=150%
```

But the launcher `/usr/local/bin/claudeclaw-start` ends with:

```
exec su -s /bin/bash claw -c "$LAUNCHER"
```

`su` opens a PAM session which creates a new login scope under `user.slice`. The daemon and all its children end up in `/user.slice/user-1000.slice/session-X.scope` — **not** in `claudeclaw.service` cgroup.

Verified on production: `pids.current=0` on the service cgroup despite the daemon and 3 long-lived `claude` processes being alive.

## Impact

The Tier 1 hardening is non-functional. If a runaway claude leaks RAM, the 2G `MemoryMax` does not fire. If the daemon spawns more than 50 tasks (we currently have 33+ threads across 3 claudes), the `TasksMax` does not fire. CPU is uncapped.

This is the *safety net* that was supposed to catch the symptoms behind the PTY crash chain. It is currently inactive.

## Fix — Option A (recommended)

Run the daemon directly as the `claw` user. Remove the `su` indirection:

```
[Service]
User=claw
Group=claw
ExecStartPre=/usr/local/bin/claudeclaw-fetch-doppler-env  # writes the env file as root if needed
ExecStart=/bin/bash -c 'set -a; source /home/claw/.claudeclaw-env; set +a; exec /usr/local/bin/bun run /opt/claudeclaw/src/index.ts start --web'
```

Doppler can be invoked directly as `claw` if its credentials are accessible to that user, dropping the need for the ExecStartPre split.

## Fix — Option B (lighter)

Keep the `su` indirection but ALSO declare the limits on `user-1000.slice` via `/etc/systemd/system/user-1000.slice.d/limits.conf`.

Less clean — couples ClaudeClaw resource policy to the system-wide user slice — but doesn't require changing how the daemon boots.

## Test plan

- Pre-deploy: `systemctl cat claudeclaw.service` shows new unit.
- Post-deploy: `cat /proc/<daemon-pid>/cgroup` shows `/system.slice/claudeclaw.service` (not `user.slice`).
[...]

PR #309, opened 2026-07-09
fix(deploy): systemd memory hardening — freeze-zone docs + cgroup pressure probe on /api/health (#178)

Nibbler1250:
## What

Complete handling of the #178 class — both the "limits don't apply" failure you found (su re-parenting) and a second, nastier variant we hit in production on 2026-07-08: **limits that apply but freeze instead of kill**.

## The incident (why #178 needs one more line than Tier 1)

An agent Bash tool call (a grep variant over a multi-hundred-MB JSONL) ballooned to 1.3 GB RSS + **3.9 GB swap** inside the service cgroup (`MemoryHigh=1.5G`, `MemoryMax=2.5G`, swap unlimited). `MemoryHigh`/`MemoryMax` bound RAM, not swap — so the hog's growth spilled into swap and `memory.current` sat between High and Max indefinitely: never OOM-killed, always throttled (`mem_cgroup_handle_over_high`, whole tree in `D` state). Symptoms that fooled every existing layer:

- Gateway port LISTENing, `/api/health` timing out — looks dead, isn't.
- Liveness heuristics saw trickling I/O ("slow turn, defer") — throttled ≠ stalled, so #297/#298-style detection never fires.
- Daemon restarts were futile: the fresh process lands in the same saturated cgroup (two observed before root-cause).

Recovery was `kill -9` of the hog: throttle lifted instantly (health 200 in 4 ms).

## Changes

1. **`docs/deploy-systemd-hardening.md`** — your Option A unit (no `su` indirection) with Tier-1 limits **including `MemorySwapMax`** (the line that closes the freeze zone), the verification steps that catch the su-bypass (`/proc/PID/cgroup` must show the service, `pids.current > 0`), and the operator remediation: find + kill the largest child inside the cgroup, never restart-loop the daemon.
2. **`src/observability/memory-pressure.ts`** — self-probe of the daemon's own cgroup v2 memory files (`current/high/max/swap` + the `memory.events` `high` counter). Injectable paths for tests; `supported: false` on cgroup v1/non-Linux at zero cost.
3. **`/api/health` exposes the discriminator** — `{ memory: { overHigh, currentBytes, highBytes, highEvents } }` when supported. External monitors can now tell *freeze* (kill the hog) from *death* (restart). `ok` deliberately stays `true` so naive ok-based restart loops don't trip on it.

## Tests

- 6 new tests (`src/observability/__tests__/memory-pressure.test.ts`): freeze-zone detection, healthy, no-soft-limit, cgroup-v1/non-Linux unsupported, partial files.
- Full CI suite list locally: 1023 pass / 0 fail; build smoke clean.

We run the operator-side counterpart live (watchdog layer that kills the largest cgroup child on `overHigh` instead of restarting — validated against a simulated freeze), and the unit now carries `MemorySwapMax=1G`.

Closes the memory half of #178; the su re-parenting half is deployment-side (your box) — the doc's verification section covers it.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.191",
+      "version": "2.2.192",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.191",
+  "version": "2.2.192",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/docs/deploy-systemd-hardening.md
+++ b/docs/deploy-systemd-hardening.md
@@ -1,0 +1,113 @@
+# Deploying under systemd — cgroup hardening that actually fires
+
+This is the companion to issue #178. It covers the two failure modes found in
+production:
+
+1. **The limits don't apply at all** — `su` re-parenting moves the daemon out
+   of the service cgroup (#178's original finding).
+2. **The limits apply but freeze instead of kill** — the "memory freeze zone":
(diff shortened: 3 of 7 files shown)


TerrysPOV:
@claude /review

Nibbler1250:
Review findings addressed: the doc now shows both health shapes (unauthenticated = boolean-only, token = full detail), and the swap fields carry an in-code note that they are module-only by design (the operator-side watchdog consumes them directly; the HTTP surface stays minimal). The sync-read cost note is acknowledged — procfs-style reads, fine at health-check cadence.

TerrysPOV:
@claude /review

TerrysPOV (approved):
good to merge

PR #311, opened 2026-07-09
fix(tuner): memory shrink relocates trimmed detail into topic files (closes #310)

### LINKED ISSUE #310: bug(tuner): memory shrink truncates index detail without guaranteeing it exists in the topic file
## Summary

The memory subject's `shrink` alternative promises **"detail stays in topics"** — but nothing in the code enforces it. At apply time, over-budget index lines are shortened (LLM rewrite at propose-time, or deterministic truncation with `…` in `reconcileToLive`/`shortenIndexLine`) **without ever checking that the trimmed detail exists in the entry's topic file**. The gate notification tells the operator "nothing is really deleted (verified)" — that claim is declarative, not enforced.

## How this bites in production

Watched over several days of daily shrink proposals on a live index:

- The agent writes updates **into the index line itself** (multi-clause hooks growing past 200 chars overnight), then the subject proposes the same cleanup every morning (same `pattern_signature`), the operator approves, the line is truncated.
- In the observed cases the detail happened to be duplicated in the topic files — so the truncation only cut duplicates. But that was **luck, not guarantee**: the day the agent writes a detail *only* in the index line (it happens), the shrink silently discards it. Recovery would be `.bak`/git archaeology.

## Fix (PR attached): relocate-then-shorten

Make the label true by construction. At apply time, for every index line the reconciled content shortens:

1. Compute the trimmed-away detail (old hook minus kept prefix).
2. Check (whitespace-normalized containment) that it exists in the entry's topic file.
3. If missing → **append the full original hook to the topic file first** (additive-only block), then accept the shortened line.
4. If the topic file is missing, is a symlink, or the pointer escapes the memory dir (traversal) → there is nowhere safe to hold the detail: **keep the line long** (no information loss, ever) and leave the entry to dedup-dead / a human.

Ordering: topic appends happen **before** the index write, so a mid-apply failure duplicates information instead of losing it.

Follow-up (separate, if wanted): recurrence awareness at the gate — a signature that keeps re-firing after repeated applies indicates an upstream producer re-creating the signal; the subject could say so instead of re-proposing the identical cleanup daily.

Nibbler1250:
## What

Closes #310 — makes the memory shrink's promise ("detail stays in topics") true **by construction** instead of by hope.

## The change: relocate-then-shorten

At apply time, for every index line the reconciled content shortens:

1. The trimmed-away detail is computed (old hook minus kept prefix, `…` stripped).
2. It's containment-checked (whitespace-normalized) against the entry's topic file.
3. **Missing there → the full original hook is appended to the topic first** (an additive-only `**Index overflow (relocated by tuner):**` block), then the shortened line is accepted.
4. A topic that is missing, a symlink, or reached via a traversal-shaped pointer has nowhere *safe* to hold the detail → the line is **kept long**. The invariant is: no information loss, ever — worst case the index stays verbose.

Ordering matters: topic appends happen **before** the atomic index write, so a mid-apply failure duplicates information (index still long + note in topic) rather than losing it.

## Security notes

- Topic writes are confined: `resolve()` within the memory dir + `startsWith(dir + sep)` (the pointer comes from index content, so it's treated as untrusted — same class as the existing M3 checks).
- `appendFileSync` follows symlinks, so a symlinked topic is refused outright (M3-style), not just resolved.
- Appends are additive-only; no file creation, no rewrites.

## Tests

6 new (relocation to topic, normalized containment skip, missing topic keeps line long, traversal pointer refused, symlinked topic refused, append is additive and ordered after existing content) — memory-subject suites 45/45, full CI suite list locally 1023 pass / 0 fail, build smoke clean.

Also validated against a copy of a real production index: a 250-char entry shortened to 197 with its unique detail landing in the topic file.

Follow-up idea (not in this PR, mentioned in #310): recurrence awareness at the gate for signatures that re-fire after repeated applies.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.191",
+      "version": "2.2.192",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.191",
+  "version": "2.2.192",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/tuner/__tests__/wisecron/subjects/memory-subject.test.ts
+++ b/src/tuner/__tests__/wisecron/subjects/memory-subject.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach } from "bun:test";
-import { mkdtempSync, writeFileSync, readFileSync, existsSync, rmSync } from "node:fs";
+import { mkdtempSync, writeFileSync, readFileSync, existsSync, rmSync, symlinkSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { MemorySubject } from "../../../subjects/memory-subject.js";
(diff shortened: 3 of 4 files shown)


TerrysPOV:
@claude /review

Nibbler1250:
All three findings addressed: (1) the already-there containment check now also covers blocks queued for the same topic in the current pass — two index lines sharing one topic land a single overflow block (test added); (2) isFile() guard beside the symlink check — a directory named *.md now falls into keep-long instead of throwing EISDIR (test added); (3) the paraphrase tradeoff is documented in place: non-prefix LLM shortenings relocate the full hook by design — zero-loss without semantic diffing. memory-subject 40/40, src/tuner suite green.

TerrysPOV:
@claude /review

TerrysPOV (approved):
Good to merge

PR #317, opened 2026-07-19
fix(runner): gate turn completion on the idle REPL footer (closes #316)

### LINKED ISSUE #316: PTY runner: turn-completion detection mis-fires on multi-step tool/sub-agent turns (job timeouts + premature completion)
## Summary

The PTY runner's turn-completion detection (`src/runner/pty-output-parser.ts`) unreliably detects turn boundaries for turns that go **quiet mid-turn** — i.e. while Claude is waiting on a tool call or a Task sub-agent. This breaks scheduled agent jobs: **~91% of jobs (10/11) since 2026-07-17 hit `PtyTurnTimeoutError` (max retries exhausted, exit 1)** across every agent — `reg`, `suzy`, and `publisher` (publisher has published nothing in ~2 months). Shorter multi-step tool turns instead complete **prematurely** with truncated output.

## Root cause

Since Claude 2.1.89 dropped the OSC progress markers, `runTurn` detects completion via a sentinel-echo round-trip gated on an activity indicator (`pty-output-parser.ts`):

1. `tick()` will only emit `quiet` once `sawActivityIndicatorThisTurn` is true (a spinner glyph `✻ ✶ ✳ ✢ ✽ ✺ ✷ ◉` was seen), then after `quietWindowMs` (500ms) of stream silence.
2. On `quiet`, the supervisor writes the sentinel; the echo is matched to end the turn.

The flaw: **once the activity glyph is seen early in the turn, any later 500ms-quiet is treated as "turn done."** But multi-step turns legitimately go silent mid-turn while waiting on a tool (`Bash`, MCP) or a `Task` sub-agent. The detector fires `quiet` mid-turn and completes the turn incorrectly.

## Reproduction

Instrumented `pty-output-parser.ts` with env-gated (`CCAW_PTY_DIAG`) trace logging of the gate state machine, then ran the **real** `run()` path (opus, `reg` MCP/security profile) with a 4-step Bash task: `echo STEP-ONE` / `sleep 5 && echo STEP-TWO` / `sleep 5 && echo STEP-THREE` / `date`, then reply `DONE-ALL-STEPS`.

Result: cut off at STEP-TWO — `sawStep3=false`, `sawDoneMarker=false`, exitCode 0. Trace:

```
tick gate:no-byte totalBytes=1506
ACTIVITY-INDICATOR-FOUND
tick gate:not-quiet-yet totalBytes=2736
tick READY-to-emit-quiet totalBytes=29367     <- 500ms quiet fired during the 2nd `sleep 5` tool wait
QUIET-EMITTED
SENTINEL-FOUND offset=29392                    <- turn completed prematurely, steps 3-4 discarded
```

By contrast, 7 single-step / pure-generation turns (including opus and resumed sessions) all detected cleanly — confirming the defect is specific to **mid-turn quiet** (tool / sub-agent waits).

## Caveat (confirmed vs inferred)

The reproduction produces *premature completion* (exit 0, truncated); production shows *timeouts* (exit 1). The exact production timeout path was not captured standalone because real deep-research jobs spawn sub-agents that OOM the RAM-constrained box when run outside the daemon. Both are manifestations of the same root — the quiet-based heuristic cannot distinguish "mid-turn tool wait" from "turn finished."

## Proposed fix

The interactive **bus** path (`src/bus/session-agent-process.ts`) already solved the analogous Claude-2.1.89 detection problem by watching for the **idle REPL footer marker** to reappear as the "turn truly done" signal, made whitespace/CHA-tolerant (commits `20b2bc3` "match the REPL footer marker whitespace-tolerant", `c118f74` "pin CHA-collapsed footer matching"). The runner never received that treatment. Port footer-marker completion detection to the runner so mid-turn tool-waits are not mistaken for completion — the idle footer only appears when the turn is genuinely done.

## Impact

- `reg`: topic research / deep-research jobs time out daily
- `suzy`: daily-digest times out (work partially completes before the timeout)
- `publisher`: publish pipeline times out — no articles published in ~2 months

TerrysPOV:
Closes #316.

## Problem
The PTY runner's turn-completion detection treated any 500ms stream-quiet (after a spinner glyph was seen) as "turn done". But multi-step **tool / sub-agent** turns legitimately go quiet mid-turn while a tool runs, so the detector fired `quiet` mid-turn and completed the turn **prematurely** (losing the rest of the work) — and for the daemon's long agent jobs it desynced into the outer turn timeout. ~91% of scheduled jobs (reg/suzy/publisher) were timing out; the publisher published nothing for ~2 months.

Reproduced with an instrumented parser + the real `run()` path (opus, MCP): a 4-step Bash task was severed at step 2 by a quiet during a `sleep` tool-wait.

## Fix
Gate `quiet` on claude's idle REPL footer being present in the recent output, matching the anchored, version-stable, **prose-safe** core `tab to cycle` (`shift+tab to cycle` in older CLIs, `tab to cycle permission modes` in 2.1.168+). That hint only paints when claude is idle at the prompt — never during a tool/sub-agent wait — so a mid-turn quiet no longer completes the turn. Mirrors the bus adapter's footer-marker approach and its prose-safe matcher; reuses the parser's existing `expandCursorForwardToSpaces` + `stripAnsi` so CUF/ANSI-interspersed footers still match. The outer turn timeout remains the backstop if a future CLI renames the hint.

## Tests
154 PTY tests green. New/updated coverage: mid-turn tool-wait suppression, CUF-tolerant footer matching, the **prose false-positive regression** (bare `to cycle` in response text must NOT trip the gate — the hole caught in review), and the 2048-byte `footerTail` eviction boundary. Existing quiet tests were updated to emit the footer (same pattern they already use for the `✻` spinner).

## Review
- **Security** (everything-claude-code:security-reviewer) — CLEAN, no CRITICAL/HIGH. Buffer bounded (2048B), no crash/ReDoS on adversarial bytes, outer-timeout backstop verified against `pty-process.ts`.
- **Code review** (feature-dev:code-reviewer) — found a CRITICAL unanchored-substring match (bare `to cycle` would false-positive on response prose / tool output). **Fixed** by anchoring to `tab to cycle` + the regression test above.

## Known tradeoff / follow-up
`tab to cycle` is now a hard prerequisite for clean turn completion. If a future CLI renames the footer, turns slow-fail via the outer timeout (retryable) rather than hang. Suggested follow-up: a CI golden-fixture assertion on the footer string so a CLI bump fails a test instead of silently degrading in production.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.192",
+      "version": "2.2.193",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.192",
+  "version": "2.2.193",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/__tests__/pty-output-parser.test.ts
+++ b/src/__tests__/pty-output-parser.test.ts
@@ -71,7 +71,10 @@ describe("pty-output-parser — sentinel flow (synthetic)", () => {
     // one of claude's spinner glyphs. Markers (`●`/`⏺`) are
     // deliberately NOT in the gate (Codex P1 on PR #124) because they
     // appear in the TUI status box + scrollback redraws.
-    feed(parser, new TextEncoder().encode("✻ a"), now);
+    // The response must also leave claude back at the idle REPL prompt — its
+    // footer's "to cycle" hint is what tells the parser the turn is genuinely
+    // done (issue #316); without it, quiet is correctly withheld.
+    feed(parser, new TextEncoder().encode("✻ a\nshift+tab to cycle"), now);
     expect(parser.sawByteSinceTurnStart).toBe(true);
 
     // Within the quiet window after that byte → still no quiet.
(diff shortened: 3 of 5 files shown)


TerrysPOV:
@claude /review

Issue #315, opened 2026-07-19
No graceful-restart guard: restarting the daemon silently kills an in-flight turn


Nibbler1250:
## Summary

Restarting the daemon (`systemctl restart`) while an agent turn is in flight terminates that turn with no guard and no signal. The reply the user was waiting on is dropped; recovery depends entirely on downstream nudges/watchdogs. There is no built-in "busy / turn-active" check, and no graceful drain on shutdown.

## Repro

1. Agent is mid-turn: a prompt has been received, tools are running, no `turn_end` has fired yet.
2. `systemctl restart <daemon>` (e.g. to deploy a change).

→ the in-flight turn is killed and the pending reply is lost. Nothing warns the operator that a turn was active.

## Impact

Any deploy or restart during activity risks a silent dropped turn. Operators currently work around this by polling the event log for recent `prompt` / `tool_use` events before restarting — fragile, per-operator, and easy to forget under pressure.

## Suggested direction (your call)

- Expose a lightweight **turn-active / busy** signal — a field on the existing health endpoint, or a small CLI check — so tooling can gate a restart on it.
- Optionally, a **graceful drain** on `SIGTERM`: stop accepting new prompts and let the current turn reach `turn_end` (bounded timeout) before exiting.

Angle: reliability + operability. Happy to draft the busy-signal half (a health-endpoint field + a `--check`-style guard) as a scoped PR if the direction is welcome.

PR #318, opened 2026-07-19
feat(health): pre-auth turn-active busy signal for graceful restart (refs #315)

Nibbler1250:
Refs #315.

## Problem

Restarting the daemon while an agent turn is in flight kills that turn silently — the pending reply is dropped. There is no signal an external deploy/restart tool can gate on to know a turn is active. (Filed as #315.)

## What this PR does — the busy-signal half

Exposes a **turn-active signal on the pre-auth `/api/health`** so external tooling can *drain-then-restart* instead of killing an in-flight turn:

- `busy: boolean` — true if any agent has a turn in flight (a `prompt` seen, no `response.turn_end` yet).
- `activeTurns: number` — the count, **only with the web token** (matches the #178 rule: the unauthenticated health payload carries boolean signals only; diagnostic detail requires the token).
- `ok` **stays `true`** while busy.

Plumbing: `BusCore` gains `activeTurnAgents(): string[]` (an aggregate sibling of the existing `isAgentTurnActive`), threaded to the web layer through the existing `BusWebUiBridge` (callback-shaped, so `server.ts` keeps no `BusCore` dependency). In legacy (non-bus) mode there is no bus, so the field is simply omitted.

## Design rationale

- **`busy` is orthogonal to health, not readiness.** A busy agent is not deadlocked and should not be evicted from routing, so `ok`/HTTP 200 is kept while busy; readiness-false / 503 is reserved for actual draining. (Kubernetes probe semantics; IETF `draft-inadarei-api-health-check` — unknown health fields are additive and MUST be ignored by consumers, so this is non-breaking.)
- **Gauge, not counter.** `activeTurns` is an integer gauge; naming avoids a `_total`/`_count` suffix (Prometheus conventions).
- **Consistent with #178.** Reuses the established pattern of putting a discriminator signal on the pre-auth `/api/health` with `ok` unchanged, boolean-only pre-auth.

## Consumer

An external restart guard reads `busy` (pre-auth, no token needed) and drains-then-waits before restarting — the two-phase graceful-shutdown pattern (flip-not-ready → wait for in-flight zero → bounded force). This PR is the signal half; a **graceful drain on `SIGTERM`** (stop accepting new prompts, let the current turn reach `turn_end` within a bounded timeout) is the natural follow-up and would let `stop()` itself honor the signal.

## Test plan

- **Unit/integration** (`src/ui/__tests__/api-auth-enforcement.test.ts`, 4 new cases): busy omitted with no bus (legacy); `busy:false` idle with `ok:true`; `busy:true` mid-turn with `ok:true`; pre-auth carries the boolean only while the count requires the token. Full `src/bus` + `src/ui` suites green (533 pass, 0 fail).
- **Live, end-to-end on a running bus-mode daemon**: `/api/health` reported `busy:false` at idle, `busy:true` continuously throughout an injected in-flight turn, and `busy:false` after `turn_end` — confirming the real wiring (`BusCore.activeTurnAgents()` → bridge → health), not just the stubbed test.
- No new lint errors; no new `tsc` errors (baseline diff on the touched files is unchanged).

Draft: opening for direction on the exposure surface (health field vs a dedicated status route) before finalizing — happy to adjust.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.197",
+      "version": "2.2.198",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.197",
+  "version": "2.2.198",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/core.ts
+++ b/src/bus/core.ts
@@ -242,6 +242,8 @@ export interface BusCore {
   /** True if the agent has a turn currently streaming (#222 reconciler reads
    *  this to skip restarting a deaf-but-mid-turn agent). */
   isAgentTurnActive(agentId: string): boolean;
+  /** #315: ids of all agents with an in-flight turn (aggregate of {@link isAgentTurnActive}). */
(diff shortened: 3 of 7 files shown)


Nibbler1250:

TerrysPOV:
@claude /review

Issue #313, opened 2026-07-18
Enhance skill-usage tracker: tag plugin + log agent spawns


txvicious:
## Enhance skill-usage tracker: tag plugin + log agent spawns

`hooks/track-skill-usage.sh` (wired as `PreToolUse[Skill]` in the operator's `~/.claude/settings.json`) logs one JSONL line per `Skill`-tool invocation to `~/.claude/claudeclaw/logs/skills-usage.jsonl`. It works, but two gaps make it unable to drive the plugin-cull decision (companion issue):

1. It logs the **skill name only** — not which **plugin** owns it. So it can't answer "is `pr-review-toolkit` used?"
2. It does **not** capture plugin **agent** spawns (Task/Agent tool), which are a large share of the per-session registry cost.

### Change
- Extend `track-skill-usage.sh` to also record a best-effort `plugin` field for each skill (resolved from the installed-plugins manifest / skill path).
- Add lightweight logging of agent-type spawns so agent usage is measurable alongside skills.
- Keep the hook **zero-token, non-blocking, always `exit 0`** (unchanged contract).

Repo copy is `hooks/track-skill-usage.sh`; the live copy at `~/.claude/claudeclaw/hooks/track-skill-usage.sh` is updated in lockstep (no auto-deploy exists between them).

### Why
Let the tracker accumulate ~1 week of real usage, then cull the "held for data" plugins (`ralph-loop`, `hookify`, `dev-browser`, `claude-mem`) from evidence rather than guesswork.

txvicious:
Created in error against the wrong repository (this is upstream). These items belong on the downstream fork and have been moved there. Apologies for the noise — closing.

Issue #312, opened 2026-07-18
Reduce first-load context: cull unused plugins (pr-review-toolkit, plugin-dev)


txvicious:
## Reduce first-load context: cull unused Claude Code plugins

First-load context for the operator session is ~36k tokens. Roughly half is the plugin/skill/agent **registry** — every enabled plugin's skill + agent-type descriptions load on *every* session, whether used or not.

### Change
Disabling the two clearly-unused plugins in the operator's `~/.claude/settings.json` `enabledPlugins`:

- **`pr-review-toolkit@claude-plugins-official`** — ships 7 agent types, each with verbose multi-example descriptions (the single largest registry contributor). The operator reviews via the `code-review` skill + external agents (codex), not these.
- **`plugin-dev@claude-plugins-official`** — only for authoring Claude Code plugins; not part of this operator's workflow.

Estimated saving: ~8–10k first-load tokens.

### Not touched (deliberately kept)
`caveman` (active), `superpowers` (carries `wayfinder`/`grill`/`tdd`/`research` — core workflow), `code-review`, `commit-commands`.

### Held for data
`ralph-loop`, `hookify`, `dev-browser`, `claude-mem` (⚠️ memory plugin — potential overlap with the file-based memory system) — pending skill-usage metrics (see companion tracker issue).

### Scope note
`enabledPlugins` lives in the operator's runtime `~/.claude/settings.json`, **not** in this repo — so this is a config change, tracked here for auditability only.

### Revert
Re-add the two keys to `~/.claude/settings.json` → `enabledPlugins`:
```json
"pr-review-toolkit@claude-plugins-official": true,
"plugin-dev@claude-plugins-official": true
```
and restart the session. Fully reversible, no data loss.

txvicious:
Created in error against the wrong repository (this is upstream). These items belong on the downstream fork and have been moved there. Apologies for the noise — closing.

PR #319, opened 2026-07-19
feat(telemetry): extensible host-telemetry producers + open stream namespace (refs #275)

Nibbler1250:
Refs #275.

## Problem

`buildHostTelemetryProvider()` composes a **hardcoded** list of producers, and `TelemetryStream` is a **closed** union of the built-in stream names. Together these mean an operator cannot compose their own telemetry source — or declare their own signal — without editing core / forking. The producer *interface* (`TelemetryProvider`) and the composite are already there; the missing piece is an extension point.

## What this does

- **`HostTelemetryConfig.extraProducers?: TelemetryProvider[]`** — operator-supplied producers, composed alongside the built-ins. Merged exactly like a built-in: `capabilities()` win the per-stream merge when available, `query()` results are concatenated. Composed **last**, so an explicit host source can upgrade a built-in stream that shipped unavailable (the merge is available-wins).
- **Widen `TelemetryStream`** to `KnownTelemetryStream | (string & {})` — a stream name is one of the built-ins **or** any operator-defined string, so a composed producer can advertise its own signal (custom producers SHOULD namespace, e.g. `custom.<name>`). The `& {}` preserves literal autocomplete for the known streams. `KnownTelemetryStream` is exported for code that still wants the closed set.

Backward-compatible: `extraProducers` is optional (omitting it is a no-op), and every existing call site / stream literal stays valid.

## Why

The extension point third-party operators need to instrument their own signals — and, in turn, to tune their own subjects — on the same rails as the built-ins, without a fork. Complements the consumer-side host-telemetry surface already shipped.

## Test plan

New cases in `host-telemetry-provider.test.ts`:
- an operator producer advertising a **custom** stream (`custom.demo`) is composed → shows up available in `capabilities()` and returns samples via `query()`;
- **no regression**: every built-in stream is still advertised;
- an available operator producer **upgrades** a built-in stream that shipped unavailable (available-wins merge);
- omitting `extraProducers` is a **no-op** (back-compat).

Full `src/tuner` + `src/skills-tuner` suites green (453 pass, 0 fail). The widening adds **no** new `tsc` errors (repo-wide count unchanged) and no lint errors — verified the two `switch (stream)` sites keep a safe default.

## Provenance

This extension point was surfaced by a third-party operator running the tuner (issues #312 / #313 by @txvicious): a downstream user instrumenting and culling their own plugins hit the wall that a host can't add its own telemetry source — or declare its own signal — without editing core. Crediting that report; it's the concrete "operator can't extend without forking" case this PR removes.

### DIFF
--- a/src/skills-tuner/core/telemetry.ts
+++ b/src/skills-tuner/core/telemetry.ts
@@ -49,7 +49,18 @@ export const TELEMETRY_STREAMS = [
   "memory_signal",
 ] as const;
 
-export type TelemetryStream = (typeof TELEMETRY_STREAMS)[number];
+/** The built-in streams this repo ships producers for. */
+export type KnownTelemetryStream = (typeof TELEMETRY_STREAMS)[number];
+/**
+ * A telemetry stream name. One of the built-in {@link KnownTelemetryStream}s,
+ * OR any operator-defined string — so a host-composed producer (see
+ * `buildHostTelemetryProvider`'s `extraProducers`) can advertise its own
+ * signal without editing this enum. The `& {}` preserves literal autocomplete
+ * for the known streams while widening to `string`. Custom producers SHOULD
+ * namespace their streams (e.g. `custom.<name>`) to avoid colliding with a
+ * future built-in.
+ */
+export type TelemetryStream = KnownTelemetryStream | (string & {});
 
 /**
  * The whole contract surface is versioned per host release for certification.
--- a/src/tuner/__tests__/wisecron/host-telemetry-provider.test.ts
+++ b/src/tuner/__tests__/wisecron/host-telemetry-provider.test.ts
@@ -741,3 +741,64 @@ describe("SessionJsonlTelemetryProducer", () => {
     expect(mem.every((s) => s.labels!.file.includes("mémoire"))).toBe(true);
   });
 });
+
+describe("buildHostTelemetryProvider — extraProducers (operator extension point)", () => {
+  // A minimal operator-supplied producer advertising its OWN (custom-namespaced)
+  // stream — the third-party instrumentation path this PR unlocks.
+  class CustomDemoProducer implements TelemetryProvider {
+    contractVersion(): string {
+      return "1.0.0";
+    }
+    capabilities() {
+      return [
+        { stream: "custom.demo" as TelemetryStream, schemaVersion: "1.0.0", available: true },
+      ];
+    }
+    async query(stream: TelemetryStream, _range: DateRange): Promise<MetricSample[]> {
+      if (stream !== "custom.demo") return [];
+      return [{ ts: new Date(NOW_MS), value: 42, labels: { source: "demo" } }];
+    }
+  }
+
+  it("composes an operator producer advertising a CUSTOM stream", async () => {
+    const provider = buildHostTelemetryProvider({ extraProducers: [new CustomDemoProducer()] });
+    const demo = provider.capabilities().find((c) => c.stream === "custom.demo");
+    expect(demo?.available).toBe(true);
+    const samples = await provider.query("custom.demo" as TelemetryStream, RANGE);
+    expect(samples).toHaveLength(1);
+    expect(samples[0]?.value).toBe(42);
+  });
+
+  it("still advertises every built-in stream (no regression)", () => {
+    const provider = buildHostTelemetryProvider({ extraProducers: [new CustomDemoProducer()] });
+    const streams = new Set(provider.capabilities().map((c) => c.stream));
+    for (const s of TELEMETRY_STREAMS) expect(streams.has(s)).toBe(true);
+  });
+
+  it("an available operator producer upgrades a built-in stream that shipped unavailable", async () => {
+    class UpgradeProducer implements TelemetryProvider {
+      contractVersion(): string {
+        return "1.0.0";
+      }
+      capabilities() {
+        return [
+          { stream: "memory_signal" as TelemetryStream, schemaVersion: "1.0.0", available: true },
+        ];
+      }
+      async query(): Promise<MetricSample[]> {
+        return [];
+      }
+    }
+    // Without `memorySignalHistoryPath` the built-in memory_signal ships unavailable;
+    // the operator producer (composed last) upgrades it via the available-wins merge.
+    const provider = buildHostTelemetryProvider({ extraProducers: [new UpgradeProducer()] });
+    const cap = provider.capabilities().find((c) => c.stream === "memory_signal");
+    expect(cap?.available).toBe(true);
+  });
+
+  it("omitting extraProducers is a no-op (back-compat)", () => {
+    const before = buildHostTelemetryProvider({}).capabilities().length;
+    const after = buildHostTelemetryProvider({ extraProducers: [] }).capabilities().length;
+    expect(after).toBe(before);
+  });
+});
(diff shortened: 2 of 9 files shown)


Nibbler1250:

TerrysPOV:
@claude /review

src/tuner/wisecron/host-telemetry-provider.ts
```diff
@@ -1094,5 +1103,8 @@ export function buildHostTelemetryProvider(
   if (cfg.tunerView) {
     providers.push(new TunerViewProvider(cfg.tunerView));
   }
+  // Operator-supplied producers last, so an explicit host source can upgrade
+  // a built-in stream that shipped unavailable (the merge is available-wins).
+  if (cfg.extraProducers?.length) providers.push(...cfg.extraProducers);
```
Nibbler1250, line 1134: `capabilities()` and `query()` disagree once an extraProducer shares a stream a built-in already serves as **available**.

extraProducers are appended last here, and the capabilities merge is first-available-wins (`CompositeTelemetryProvider.capabilities()`, ~L958-960), so the manifest reports **only the built-in** — the operator producer is invisible in `capabilities()`. But `query()` concatenates every provider (~L981), so `query('tool_call', …)` returns built-in samples **plus** operator samples. Net: fitness/telemetry consumers see doubled/mixed `MetricSample`s for that stream while the manifest claims the built-in is the sole source.

The "composed last so it can upgrade a built-in" framing is only true for an *unavailable* built-in (whose `query()` returns `[]`). For an available one the operator producer can't win the capability but still silently contributes query data. Worth either (a) de-duping query by the capability winner, or (b) documenting that extraProducers must own disjoint streams from available built-ins.

src/skills-tuner/core/telemetry.ts
```diff
@@ -49,7 +49,18 @@ export const TELEMETRY_STREAMS = [
...
+/**
+ * A telemetry stream name. One of the built-in {@link KnownTelemetryStream}s,
+ * OR any operator-defined string — so a host-composed producer (see
+ * `buildHostTelemetryProvider`'s `extraProducers`) can advertise its own
+ * signal without editing this enum. The `& {}` preserves literal autocomplete
+ * for the known streams while widening to `string`. Custom producers SHOULD
+ * namespace their streams (e.g. `custom.<name>`) to avoid colliding with a
+ * future built-in.
+ */
+export type TelemetryStream = KnownTelemetryStream | (string & {});
```
Nibbler1250, line 63: Widening to `KnownTelemetryStream | (string & {})` collapses to `string` in every value/record position, so existing internal maps silently lose tsc validation of built-in stream names.

Example: `SUBJECT_STREAM` in `src/tuner/wisecron/index.ts:63` is `Partial<Record<string, TelemetryStream>>`, and `MetricSource` (this file, ~L198) is `TelemetryStream | ARTIFACT_SOURCE`. Before this change a typo'd built-in (`cron: 'cron_runn'`) was a compile error against the strict union; now it compiles clean and the cron subject's activation gate just never finds telemetry.

Suggest widening only at the producer boundary and keeping internal maps/`MetricSource` keyed on `KnownTelemetryStream`, so custom streams stay possible without dropping the compiler check on the known ones.

src/tuner/wisecron/host-telemetry-provider.ts
```diff
@@ -1042,6 +1042,15 @@ export interface HostTelemetryConfig {
    * specialized page and falls back to the universal page like any plugin.
    */
   tunerView?: TunerViewSources;
+  /**
+   * Operator-supplied producers composed alongside the built-ins — the
+   * extension point for a host to add its own telemetry source (or a custom
+   * stream) without editing this function. Each is merged into the composite
+   * exactly like a built-in: its `capabilities()` win the per-stream merge
```
Nibbler1250: This line — "its `capabilities()` win the per-stream merge when available" — overstates what the merge does. extraProducers are appended last, and the merge is first-writer-wins with only unavailable→available upgrade (`capabilities()`, ~L958-960). So an extraProducer only wins over an **unavailable** built-in; an available built-in stays the owner. The commit message is precise ("upgrades an unavailable built-in"); this doc reads as if it overrides any built-in. A reader will expect their producer to replace a working built-in and be surprised (see the capabilities/query note on the push line).

src/tuner/__tests__/wisecron/host-telemetry-provider.test.ts
```diff
@@ -741,3 +741,64 @@ describe("SessionJsonlTelemetryProducer", () => {
     expect(mem.every((s) => s.labels!.file.includes("mémoire"))).toBe(true);
   });
 });
+
+describe("buildHostTelemetryProvider — extraProducers (operator extension point)", () => {
```
Nibbler1250, line 745: The four added tests cover disjoint/unavailable composition only: a custom-namespaced stream, no-regression of built-ins, upgrade of an *unavailable* built-in, and the empty no-op. None pins the actual footgun — an extraProducer contesting an **available** built-in stream. Neither the capabilities winner nor the `query()` concatenation (doubling) is asserted for that case, so a future merge/order change (or the duplication flagged on the push line) ships green. One extra test: built-in available + extraProducer available for the same stream → assert exactly one capability entry AND a defined query-sample count.

Nibbler1250 (commented):
High-effort local pass on the extensible-producers change. Feature is small and backward-compatible as advertised — no crash-class bug. The findings cluster around one real trap: the gap between what the compose-last merge actually does and what the docs claim it does.

**Headline (inline on the push line):** `capabilities()` is first-available-wins, but `query()` concatenates every provider. So an operator producer that contests a stream a built-in already serves *available* becomes an invisible-but-active query contributor — the manifest hides it, the samples still double. "Compose last to upgrade a built-in" only holds when the built-in shipped *unavailable*.

Other three are the type-safety cost of the `string & {}` widening, a doc line that overstates the merge, and the matching test gap. Details inline. Nothing here blocks — flagging for your call on the merge.

Nibbler1250:
Addressed my own review in e254ad3:

- **capabilities/query footgun** — `CompositeTelemetryProvider.capabilities()` now warns once per stream when two producers serve it `available` (query() concatenates, so the second would silently double-count). Kept concatenation as-is — no hot-path dedupe.
- **type-safety** — re-keyed `SUBJECT_STREAM` + `deriveCapabilitiesFromHealthChecks` on `KnownTelemetryStream`, so a typo'd built-in stream name is a tsc error again. `TelemetryStream`/`MetricSource` stay widened for real custom streams.
- **doc** — rewrote the `extraProducers` contract (disjoint streams; only upgrades an `unavailable` built-in) and dropped the "capabilities() win when available" wording.
- **tests** — pinned the contested-available case (one capability, warns once, query still doubles), the no-warn upgrade path, and custom-stream disjointness.

Leaving the merge to the maintainer.

Nibbler1250:
Good catch — you're right, and it's the path that actually matters. My earlier self-review stopped at the in-process `CompositeTelemetryProvider` and never traced through to the served MCP surface, so I missed exactly the gap you flagged: `telemetry__capabilities` advertises a custom `custom.<name>` stream `available: true`, but `QueryArgsSchema`'s `z.enum(TELEMETRY_STREAMS)` rejected it at query time and `McpTelemetryProvider.query` swallowed the error → available-but-always-empty.

Fixed in caaf598:

- `QueryArgsSchema.stream` widened `z.enum(TELEMETRY_STREAMS)` → `z.string().min(1).max(128)` (bounded since the surface is socket-exposed; unknown streams just return no samples, since each producer owns its own). Dropped the now-unused `TELEMETRY_STREAMS` import.
- Added an end-to-end `telemetry-mcp.test.ts` case that round-trips a custom stream through `registerHostTelemetryTools` + `McpTelemetryProvider` (the served path, not just the in-process provider). Confirmed it fails against the old enum schema and passes now.

This closes the #312/#313 operator path — a host-instrumented custom signal now reaches the tuner through MCP, not just via direct in-process composition.

PR #320, opened 2026-07-19
feat(tuner): finding-lifecycle ledger primitive (refs #275)

Nibbler1250:
Refs #275. Stacked on #319 (extensible producers).

## Problem

Finding-based subjects — a system advisory, a degraded signal, a low-ROI plugin — emit **instantaneous** findings each cycle. To act well, a subject needs the state **across runs**: has this finding just appeared, is it stably open, has it cleared? Today there is no such primitive: `fitness.ts` does windowed *measurement*, `proposals.ts` dedups proposal *signatures* — neither tracks a finding's lifecycle. Without it, a subject either re-alerts on every raw sighting or hand-rolls its own dedup/debounce (as I've done twice already off-tree).

## What this adds

`finding-ledger.ts` — a **pure, level-triggered** `reconcile(prior, current, nowIso, cfg)`:

- Identity = **fingerprint**. Each cycle takes the prior ledger + the full current finding set, returns the next ledger plus the transition edges.
- State machine mirrors Prometheus/Alertmanager: `pending → firing → resolving → resolved`. `forCycles` debounces the open edge (a finding must recur before it fires); `resolveAfterCycles` debounces the close edge (auto-resolves once the fingerprint stops recurring). This **suppresses flapping**.
- **Level-triggered** (Kubernetes-reconcile style): the full current set drives every transition, so a restarted producer just re-reads current state and resumes — no position to remember.
- **I/O-free**: takes a caller-supplied `nowIso`, returns `{ entries, newlyFiring, newlyResolved }`. Persistence (jsonl) is the caller's concern — keeps the core pure and testable, and lets the caller choose storage.

## Intended consumer

A finding-based subject (a system-advisory-style detector, or any subject debouncing its `degraded` signal) reconciles its per-cycle findings through the ledger and reacts to `newlyFiring` / `newlyResolved` — getting anti-flap + open/close edges for free instead of re-deriving them. Opening **draft** as the primitive; the first in-tree consumer follows (it pairs naturally with an operator producer composed via #319).

## Test plan

10 cases in `finding-ledger.test.ts` covering every transition: first-sighting is pending; fires after `forCycles`; `forCycles=1` fires immediately; firing persists without a duplicate open edge; single miss → resolving; `resolveAfterCycles` misses → resolved (emitted once, then dropped); re-fires when a resolving finding reappears; **flap suppression** (appear/miss/appear under the debounce never fires); independent multi-fingerprint tracking with `meta` carried through; `firstSeen` preserved while `lastSeen` advances. No new `tsc`/lint.

## Provenance & reference consumer — why the first consumer is Python

Part of the same initiative catalyzed by a third-party operator's usage (#312 / #313 by @txvicious): making the self-tuning layer robust and extensible for *any* operator, not just the reference host. Every finding-based subject would otherwise re-roll dedup/debounce (as happened twice off-tree) — hence this primitive.

The first reference consumer is a **host-side advisory producer written in Python — deliberately**. It began as a standalone probe collector for the *producers-first* phase: fast to iterate, runs on a timer, no build step, so it can accumulate real data before the TypeScript subject is committed. That producer uses a **faithful Python shim of this module** (parity-tested against the same transitions), so the local consumer gets lifecycle tracking today while this TS module stays the upstream home in wisecron. Validated end-to-end on a host across `pending → firing → resolving → resolved` with real findings (a transient failed unit + persistent open items).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.197",
+      "version": "2.2.200",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.197",
+  "version": "2.2.200",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/tuner/__tests__/wisecron/finding-ledger.test.ts
+++ b/src/tuner/__tests__/wisecron/finding-ledger.test.ts
@@ -1,0 +1,100 @@
+import { describe, it, expect } from "bun:test";
+import { reconcile, type Finding, type LedgerEntry } from "../../wisecron/finding-ledger.js";
+
+const T0 = "2026-07-19T00:00:00.000Z";
+const T1 = "2026-07-19T04:00:00.000Z";
+const T2 = "2026-07-19T08:00:00.000Z";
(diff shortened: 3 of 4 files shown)


Nibbler1250:

TerrysPOV:
@claude /review

PR #333, opened 2026-07-21
fix(security/discord): deny mention parsing on outbound sends (closes #323)

### LINKED ISSUE #323: security(discord): no `allowed_mentions` on outbound sends — agent text can ping @everyone
Surfaced during review of #321 (#301). Pre-existing and repo-wide, not specific to that change.

## Problem

`src/adapters/discord/rest-api.ts` POSTs `{content: chunk}` with no `allowed_mentions` field. Discord defaults to parsing every mention in `content`, so `@everyone`, `@here`, and role mentions in outbound text ping for real if the bot holds the permission.

`grep -rn "allowed_mentions" src/` returns nothing.

## Why it matters

Every outbound Discord path is affected — agent replies, permission prompts, `system.request_human`, and now `system.operator_alert`. The text is model-generated or interpolates arbitrary strings (tool names, `err.message`), so it is not operator-authored and not escaped anywhere.

Not currently known to have fired. Raising it because the mitigation is one field and the failure mode is loud and social rather than technical.

## Suggested fix

Add `allowed_mentions: { parse: [] }` to the send payload in `rest-api.ts`, opting in explicitly wherever a mention is genuinely wanted. Consider the equivalent for Slack.

## Related

Telegram has a milder version: `markdownToTelegramHtml` (`src/adapters/telegram/format.ts`) escapes `&<>` at step 4, but step 6 converts `[label](url)` into a real `<a href>` escaping only `"`, so link injection through interpolated text is possible. Same class, lower severity — worth folding into the same pass.

Nibbler1250:
Closes #323.

## Problem

Model-generated agent text (`@everyone`, `@here`, role mentions), plus interpolated strings (tool names, `err.message`), was POSTed to Discord with no `allowed_mentions`, so Discord parsed and fired those mentions for real. Never operator-authored, escaped nowhere.

There are **two** live Discord runtimes and both were affected:
- the Bus adapter — `src/adapters/discord/rest-api.ts`;
- the legacy `claudeclaw discord` command — `src/commands/discord.ts` (an `index.ts` entrypoint), including its **streaming edit** path, which is the primary render for streamed replies.

## Fix

Add `allowed_mentions: { parse: [] }` to every model-output send on both runtimes:

- **rest-api.ts** — chunked message POST + interaction callback.
- **commands/discord.ts** — `sendMessage`, image-caption sends + `uploadImageMessage` `payload_json`, the streaming `PATCH …/messages/{id}` edit, the webhook `@original` followup, and `respondToInteraction` (folded into the helper so every interaction response is covered uniformly).

Mentions become opt-in wherever genuinely wanted.

## Found by adversarial review

The first pass fixed only `rest-api.ts` (the file the issue named). An adversarial review pass caught that `commands/discord.ts` — a second live entrypoint with its own send sites and the streaming path — was left open, so #323's vector still fired through `claudeclaw discord`. This PR now covers both.

## Test plan

Two new cases in `rest-api.test.ts` assert `allowed_mentions: { parse: [] }` on both adapter send paths. `commands/discord.ts` has no unit harness today (pre-existing), so its sites are verified structurally: every content-bearing send now carries the field, `tsc` clean, and the full `src/adapters/discord` suite is green (68 pass, 0 fail). No new lint.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.197",
+      "version": "2.2.201",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.197",
+  "version": "2.2.201",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/adapters/discord/__tests__/rest-api.test.ts
+++ b/src/adapters/discord/__tests__/rest-api.test.ts
@@ -106,4 +106,21 @@ describe("DiscordRestApi", () => {
     const api = new DiscordRestApi({ token: "tok", logger: SILENT, fetchImpl: fake.fetchImpl });
     await expect(api.sendMessage("c1", "x")).rejects.toThrow(/403/);
   });
+
+  it("sendMessage denies all mention parsing (#323)", async () => {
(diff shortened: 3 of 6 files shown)


Nibbler1250:
@TerrysPOV took this one — small security fix, CI green. Slack/Telegram equivalents flagged as separate follow-ups. No rush.

Issue #325, opened 2026-07-20
adapters: Slack and the web UI do not render system.operator_alert (#301 follow-up)


TerrysPOV:
Follow-up to #301 / #321, which added the `system.operator_alert` topic.

## Problem

#321 introduced `system.operator_alert` as the daemon's "alert the operator on their surface" primitive, and wired it into the Discord and Telegram adapters. Two surfaces were left out:

- **Slack** — `src/adapters/slack/index.ts` subscribes to `response.text`, `channel.permission_request`, and `system.request_human`, but not `system.operator_alert`.
- **Web UI** — `src/bus/webui-bridge.ts` subscribes only to `response.text`.

An operator whose primary surface is Slack or the dashboard therefore still sees stall-watchdog alerts only in `journalctl` and `stall-kills.jsonl` — the exact log-only situation #301 set out to fix.

## Not a regression

Neither surface delivered these alerts before #301, so nothing got worse. #301's acceptance criteria named Discord specifically, and the scope was a deliberate call rather than an oversight. Filing so the gap is tracked rather than assumed closed by #301.

## Suggested approach

Slack is the cheaper of the two and mirrors Discord closely: add the topic to the subscription list and add a render branch alongside the existing `system.request_human` handler. Two constraints from #321 worth carrying over:

1. **Do not route it through the reply path.** The whole point of the dedicated topic is that it carries no turn semantics. On Telegram, routing an alert through `handleResponseText` would have closed the turn receipt as `turn_observed` and edited the alert over the agent's live reply message. Slack's reply handler should be checked for equivalent turn/thread state before reusing it.
2. **Bound the blast radius.** #321 narrowed the Discord alert to a single channel because `primaryChannelByAgent` containment is opt-in (#151) and the fallback fans out to every channel *and thread* mapped to the agent — fine for a periodic heartbeat, an alert storm for a stall alert. Slack has the same `primaryChannelByAgent` shape (`src/adapters/slack/types.ts`) and likely wants the same treatment.

For the web UI the question is different — an alert is not chat output there, so it may belong in a notifications surface rather than spliced into the conversation stream.

Issue #301, opened 2026-07-04
Stall watchdog: deliver the false-positive flag to Discord (operator-alert path)


TerrysPOV:
Follow-up from the stall-watchdog wiring (#300, completing PR 1/3 of #296).

## Current state
The stall watchdog's auto-discovery classifies each kill and, on a `suspected_false_positive`, produces an operator flag with a suggested ceiling raise (e.g. *"Stall-killed reg on `Bash` after 950s but it looked alive — raise `stallWatchdog.ceilings.bash.killSeconds` (suggest ~1920s)"*). Today that flag is surfaced via:
- the structured **daemon logger** (`logger.error("[stall-watchdog]", …)`), operator-visible in `journalctl`, and
- the durable **audit** at `.claude/claudeclaw/stall-kills.jsonl`.

## The gap
It does **not** reach the operator on **Discord** (or Telegram). A wrong kill — the exact case the operator most needs to see, so they can loosen the ceiling — currently requires reading server logs to notice. The `notify` dep in `runtime-mount.ts` is bound to `logger` only.

## The work
Route the `critical`-level stall-watchdog `notify` through the operator-alert / bus-delivery path so a `suspected_false_positive` flag (and a failed-restart alert) is delivered to the operator's channel, not just logged. Needs the "alert the operator on their surface" primitive that doesn't cleanly exist at the wiring point yet (same missing path #227 hit when it deferred summary-injection). Keep it best-effort (delivery failure must never block recovery) and rate-limited (respect the existing per-session cooldown so it can't storm the channel).

## Acceptance
- A `suspected_false_positive` kill delivers its flag + ceiling suggestion to the operator's chat surface (Discord), in addition to the log + `stall-kills.jsonl` audit.
- Delivery is best-effort and de-duplicated/cooled-down; a delivery error never affects recovery.

**Labels:** enhancement · Refs: #296, #300, #295

PR #334, opened 2026-07-21
feat(slack): render system.operator_alert (refs #325)

Nibbler1250:
Refs #325.

## Problem

#321 added `system.operator_alert` (the daemon's "alert the operator on their surface" primitive) and wired it into Discord and Telegram. Slack was left out: it subscribes to `response.text`, `channel.permission_request`, and `system.request_human`, but not `system.operator_alert`. An operator whose primary surface is Slack therefore still saw stall-watchdog alerts only in `journalctl` / `stall-kills.jsonl` — the log-only situation #301 set out to fix.

## What this does

Adds the `system.operator_alert` subscription and a `handleOperatorAlert` handler that mirrors Discord, carrying over both constraints from #321:

1. **Not routed through the reply path.** The alert carries no turn semantics, so it does not go through `handleResponseText` — it must not close a turn receipt or register a pending ask. Standalone `chat.postMessage`.
2. **Blast radius bounded to one channel.** It targets the agent's primary channel when set (`primaryChannelByAgent`), else the first resolved channel — never the fan-out that maps every channel *and thread* to the agent. Fine for a heartbeat; an alert storm for a stall alert. Matches Discord's and Telegram's single-target semantics, so one logical alert has the same reach on every surface.

## Scope

This is the Slack half of #325. The **web UI** half is deliberately left for a follow-up: as the issue notes, an alert there isn't chat output and likely belongs in a notifications surface rather than spliced into `webui-bridge`'s conversation stream — a small design call worth its own PR.

## Test plan

Three new cases in `slack.test.ts`: the alert is delivered as a standalone message; it is bounded to the primary channel (not fanned out); an empty alert is dropped. Updated the subscription-count assertion (3→4 topics per agent). Full `src/adapters/slack` suite green (74 pass, 0 fail); no new `tsc`/lint.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.197",
+      "version": "2.2.202",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.197",
+  "version": "2.2.202",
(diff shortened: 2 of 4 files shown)


Nibbler1250:
@TerrysPOV Slack half of #325 — mirrors Discord (off the reply path, one-channel blast radius). Left the web UI half as a follow-up since it is a notifications-surface design call, not a mirror. CI green.

Issue #327, opened 2026-07-20
tsc: 338 type errors repo-wide block a typecheck step; 14 are the it() timeout overload


TerrysPOV:
Found while fixing the `settings.cache` cluster of #304. Splitting it out because it blocks the typecheck step proposed in #324.

## The immediate 14

`bunx tsc --noEmit` reports 14 errors across the two MCP integration test files, **all the same one**:

```
src/__tests__/mcp-multiplexer-integration.test.ts(195,5): error TS2353:
  Object literal may only specify known properties, and 'timeout' does not
  exist in type '(done: (err?: unknown) => void) => void | Promise<unknown>'
```

They come from the three-argument `it()` form used for slow integration tests:

```ts
it("real SDK client lists tools and round-trips a tools/call over loopback HTTP", {
  timeout: 10000,
}, async () => { ... });
```

**This is a type-definition gap, not a bug.** Bun supports the options object at runtime — these tests pass (14/14 after the #304 cluster-1 fix). `bun-types` just doesn't declare the `(name, options, fn)` overload the runtime accepts.

Worth resolving deliberately rather than by silencing:

- confirm the supported signature for the pinned Bun version (the argument order may be `(name, fn, options)`), and use that form; or
- if the runtime genuinely accepts `(name, options, fn)` and the types are simply behind, upstream the overload or add a local declaration merge — not a blanket `@ts-expect-error`, which would hide real errors on those lines.

## The bigger number

Repo-wide, `bunx tsc --noEmit` currently reports **338 errors**. So "add a typecheck step to CI" (#324) is not a one-line change — the backlog has to be triaged first. From a sample, the categories are:

- test-runner type gaps like the above (no runtime impact);
- `vitest` imports in `src/__tests__/gateway/*` that resolve to nothing, so `describe`/`test`/`expect` are all undeclared;
- genuine type errors in `docs/spikes/probes/*` (missing required `IPtyForkOptions.name`, `string` assigned to `Uint8Array`);
- real drift in source, e.g. `handleKill`'s `Extract<>` collapsing to `never` (#324).

The last category is the one that matters — those are the errors a typecheck step would have caught before the tests in #304 rotted. Note `tsconfig.json` currently has no `exclude`, so `docs/spikes/` is being typechecked alongside `src/`.

## Suggested sequence

1. Triage the 338 into: real source errors, test-runner type gaps, and files that should be excluded (`docs/spikes/`).
2. Fix or exclude, category by category.
[...]

PR #335, opened 2026-07-22
fix(bus): split StallDecision so handleKill is type-checked (closes #324)

### LINKED ISSUE #324: bus: handleKill is not type-checked — Extract<StallDecision, {action:"kill"}> resolves to never
Deferred from #300 (note 1), re-surfaced during review of #321.

## Problem

`StallDecision` (`src/bus/stall-watchdog.ts:198-206`) is a two-member union whose second member carries a **combined** action:

```ts
export type StallDecision =
  | { action: "none" }
  | { action: "warn" | "kill"; tool: string; toolUseId: string; outstandingMs: number; ceiling: ToolCeiling };
```

`handleKill` takes `decision: Extract<StallDecision, { action: "kill" }>` (`:416`). Because `"warn" | "kill"` is not assignable to `"kill"`, the `Extract` matches nothing and the parameter type is `never` — so every property access inside `handleKill` is unchecked. `tsc --noEmit` reports these as errors today (11 in that file, all pre-existing).

## Why it matters now

#321 made `notify`'s third parameter (`agentId`) **required** and added it at five call sites, three of which are inside `handleKill`. In an un-type-checked function a missed argument would not be caught at build time, and there is no `tsc`/typecheck step in `package.json` or any workflow to catch it either. At runtime it would surface as `agentId === undefined` hitting the notifier's `if (!agentId) return;` guard — a silently swallowed alert.

All five call sites were verified by hand for #321 and are correct. This is about removing the class of failure, not fixing a live bug.

## Suggested fix

Split the union so `kill` and `warn` are separate members:

```ts
| { action: "warn"; tool: string; /* … */ }
| { action: "kill"; tool: string; /* … */ }
```

That restores `Extract` and type-checks `handleKill`. Worth pairing with a `typecheck` script and a CI step so the class stays closed — currently nothing type-checks the repo in CI.

Nibbler1250:
Closes #324.

## Problem

`StallDecision` carried `warn` and `kill` in a single combined member:

```ts
| { action: "warn" | "kill"; tool: string; toolUseId: string; outstandingMs: number; ceiling: ToolCeiling }
```

`handleKill` takes `Extract<StallDecision, { action: "kill" }>`. Because `"warn" | "kill"` is not assignable to `"kill"`, the `Extract` matched nothing and the parameter resolved to `never` — so every property access inside `handleKill` was unchecked. Concretely: #321 made `notify`'s `agentId` argument required and added it at three call sites inside `handleKill`; in an un-type-checked function a missed argument wouldn't be caught at build time (it would surface at runtime as `agentId === undefined` hitting the notifier's guard — a silently swallowed alert).

## Fix

Split `warn` and `kill` into distinct members over a shared `StallSignal` shape. `Extract<StallDecision, { action: "kill" }>` now resolves to the kill member and `handleKill`'s body is genuinely type-checked. Construction in `sweep()` stays type-safe by building each member with a **literal** action (the object is otherwise not assignable to the split union).

## Verification

- Repo-wide `tsc --noEmit` drops from **452 to 441 errors** — the 11 removed are exactly this file's, which go to **0**.
- Full `src/bus` suite green (503 pass, 0 fail).
- No new lint.

## Out of scope

The issue suggests pairing this with a `typecheck` script + CI step. That's worth doing but can't go green yet — 441 pre-existing repo-wide errors remain (tracked in #327). Adding the CI gate belongs in that cleanup, not here.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.197",
+      "version": "2.2.203",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.197",
+  "version": "2.2.203",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/stall-watchdog.ts
+++ b/src/bus/stall-watchdog.ts
@@ -195,15 +195,21 @@ export function newSessionState(
   return { agentId, sessionId, outstanding: new Map(), lastActivityAt: now };
 }
 
+/** The shared shape of an actionable stall (everything but the verb). */
+type StallSignal = {
+  tool: string;
+  toolUseId: string;
+  outstandingMs: number;
+  ceiling: ToolCeiling;
+};
+
+// `warn` and `kill` are DISTINCT members (not a combined `"warn" | "kill"`),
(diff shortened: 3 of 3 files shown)


Nibbler1250:
@TerrysPOV type-only fix — drops the file 11→0 tsc errors (repo-wide 452→441), bus suite green. Left the typecheck-CI pairing for the #327 cleanup since it cant go green with the pre-existing errors.

PR #337, opened 2026-07-22
fix(bus): give the warn-only critical escalation its own latch (closes #322)

### LINKED ISSUE #322: bus: shared `warned` latch suppresses the critical alert in stallWatchdog.action="warn" mode
Surfaced during review of #321 (#301). Pre-existing, not introduced there.

## Problem

`sweep()` has two branches that both gate on **and set** the same per-tool `warned` latch:

- `src/bus/stall-watchdog.ts:358-370` — the soft warn at the warn ceiling
- `src/bus/stall-watchdog.ts:374-386` — the `critical` "past kill ceiling, NOT restarting" alert, used when `stallWatchdog.action: "warn"`

`evaluateStall` returns `warn` first for a given `tool_use_id` (the ceilings are ordered warn-then-kill), so the first sweep past the warn ceiling sets `t.warned = true`. When the same tool later crosses the **kill** ceiling, the second branch finds `t.warned` already true and its `notify("critical", …)` never fires.

## Impact

In warn-only mode the operator gets the soft "outstanding for 120s" notice and **never** the critical escalation — which is the one that says the tool blew through the kill ceiling. Now that #301 delivers these to Discord/Telegram rather than just the log, the missing alert is more visible as a gap.

Not a problem in the default `action: "kill"` mode, where the kill path does not consult the latch.

## Suggested fix

Separate the two latches, e.g. `warned` and `criticalWarned` on `OutstandingTool`, so the escalation can fire once even after the soft warn has. Worth a test asserting both alerts arrive for one tool crossing both ceilings in warn-only mode.

Nibbler1250:
Closes #322.

## Problem

`sweep()` had two notify branches sharing one per-tool `warned` latch:
- the **soft warn** at the warn ceiling (`stall-watchdog.ts:359`);
- the **critical** "past kill ceiling, NOT restarting" escalation used in warn-only mode (`action: "warn"`).

`evaluateStall` returns `warn` before `kill` (ceilings are ordered warn-then-kill), so the first sweep past the warn ceiling sets `t.warned = true`. When the tool later crosses the **kill** ceiling, the critical branch finds `t.warned` already true and its `notify("critical", …)` never fires. In warn-only mode the operator gets the soft "outstanding for 120s" notice and **never** the critical escalation — the one that says the tool blew through the kill ceiling. Since #301 delivers these to chat surfaces (not just the log), the missing alert is a visible gap. Not a problem in the default `action: "restart"` mode, where the kill path never consults the latch.

## Fix

Give the critical escalation its own latch: add `criticalWarned: boolean` to `OutstandingTool`; the warn-only critical branch gates on `!t.criticalWarned` / sets it. Now both alerts fire once each for a tool that crosses both ceilings.

## Test

New case in `stall-watchdog.test.ts`: in warn-only mode a tool crosses the warn ceiling then the kill ceiling; asserts **both** a `warn` and a `critical` ("NOT restarting") notify arrive. It is a genuine regression test — it **fails without the src fix** (the critical is `undefined`) and passes with it (verified by stash-reverting only the source).

## Note

Found via an adversarial review pass that also caught a second `OutstandingTool` construction site (a test helper) needing the new field — fixed, so `tsc` is unchanged (repo-wide 452, no new errors). Full `src/bus` suite green (504 pass, 0 fail); verified the daemon still boots clean with the change.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.198",
+      "version": "2.2.204",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.198",
+  "version": "2.2.204",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/bus/__tests__/stall-watchdog.test.ts
+++ b/src/bus/__tests__/stall-watchdog.test.ts
@@ -27,7 +27,7 @@ function evt(
 
 function stateWith(tools: Array<{ id: string; name: string; startedAt: number }>) {
   const s = newSessionState("reg", "s1", 0);
-  for (const t of tools) s.outstanding.set(t.id, { ...t, warned: false });
+  for (const t of tools) s.outstanding.set(t.id, { ...t, warned: false, criticalWarned: false });
   return s;
 }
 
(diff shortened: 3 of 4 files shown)


Nibbler1250:
@TerrysPOV separate criticalWarned latch so the warn-only escalation fires. Regression test fails without the fix. Full gate incl. adversarial review (which caught a second construction site) + a clean-boot check. CI green.

PR #341, opened 2026-07-24
feat(bus): durable schedule_task tool + block native session-scoped cron (closes #342)

### LINKED ISSUE #342: Reminder scheduled via native CronCreate silently never fires under bus runtime
## What happened

A user asked the daemon (Discord) to "remind me to call the GP for Ginna at 8am tomorrow." The bot confirmed ("Done — I'll ping you at 8am tomorrow"), but the reminder never fired.

## Root cause

The agent scheduled it with Claude Code's **native `CronCreate`** tool (`cron: "0 8 24 7 *"`), not ClaudeClaw+'s own file-backed job scheduler. Native `CronCreate`'s wakeup is tied to the specific underlying Claude session that created it. Under `runtime: bus`, sessions rotate/churn independently of the daemon, so the session that owned the registration was gone by fire time and the wakeup silently died. The systemd daemon itself never restarted — confirmed via continuous uptime and no post-mortem in the window — so "the daemon crashed" is ruled out.

Verified against the live server's event log: the agent called `ToolSearch("select:CronCreate")` then `CronCreate(...)`. Nothing in ClaudeClaw+ defines or manages that native tool.

Secondary finding: an older standing reminder file (`reminder-ginna-gp.md`) was missing its required `schedule:` frontmatter field, so `parseJobFile()` returned null and it was never scheduled at all — a broken job produced by hand-writing the file directly.

## Fix

Make the agent always use ClaudeClaw+'s durable, file-backed cron for any scheduling (one-off or recurring), and remove the native scheduling tools so there is no wrong door. Give the agent a validated tool so it cannot produce a broken job file.

TerrysPOV:
## Problem (closes #342)

An overnight "remind me to call the GP for Ginna at 8am tomorrow" never fired. Root cause: the agent used Claude Code's **native `CronCreate`**, whose wakeup is tied to the live session — under the bus runtime, sessions rotate/churn independently of the daemon, so the wakeup silently died before fire time. A second, older reminder file was also broken (missing its `schedule:` field), so it never scheduled at all.

The ask wasn't to fix those files — it was to make the agent **always** use ClaudeClaw+'s own durable cron for any task, one-off or recurring.

## The change (three parts, wired for `runtime: bus`)

**A — block the broken door.** Every bus agent now spawns with `--disallowedTools CronCreate,CronDelete,CronList,ScheduleWakeup` (`src/bus/session-manager.ts`). These are architecturally broken under bus, not a preference, so it's an unconditional hardcoded floor — **takes effect on deploy, no server settings.json edit needed.**

**B — add a durable door.** New `schedule_task` / `list_scheduled_tasks` / `delete_scheduled_task` bus MCP tools (`src/bus/mcp-tools.ts`, handlers in `src/bus/mcp-server.ts`) backed by a new `src/bus/schedule-ops.ts`. They write file-backed cron jobs that survive restarts/session boundaries. `cron` is **required** in the tool schema, so the missing-`schedule:` bug class is structurally impossible through this path.

Writes are routed **over IPC to the daemon** (via `src/bus/core.ts`), not done in the MCP subprocess — the subprocess has the wrong cwd, so a direct write would land where `loadJobs()` never scans. This mirrors the existing `dispatch_job` pattern. The 30s hot-reload picks the job up with no restart. Default/global session → flat jobs dir (where its reminders already live); named agent → its own `agents/<id>/jobs/`.

**C — steer.** `schedule_task`'s description tells the agent to use it for any scheduling and never the native cron tools.

## Security review

An independent security pass found a **frontmatter-injection via the `model` field** (a newline could forge an `agent:` reroute or `enabled: false`). Fixed: `model` is validated against the allowed set, plus a belt-and-braces newline guard rejects newlines in every value written to frontmatter (also closes a `cron` internal-newline self-DoS). `label` (path-traversal), `prompt` (body-only), and `agentId` (daemon-stamped, not client-supplied) were confirmed safe.

**Deferred (LOW, follow-up):** no per-scope cap on job count — a runaway persona could bloat the jobs dir. Duplicate labels already throw; prompt is capped at 100KB.

## Tests

15 new tests in `src/bus/__tests__/schedule-ops.test.ts` — the load-bearing one is the **round-trip guard**: a task written by `scheduleTask` must re-parse via `loadJobs()` with a non-null `schedule` (directly reproduces the incident bug). Plus flat-vs-agent target, path-traversal / invalid-cron / empty-prompt / duplicate / model-injection / cron-newline rejection, list/delete, and a schema required-field guard. `session-manager.test.ts` gains a test that `buildClaudeArgs` emits the four blocked tools.

## Review status — please read

**Stage-8 code review has NOT run.** The repo's CI `Claude Code Review` workflow is `disabled_manually`, and `/code-review:code-review` is user-triggered so it couldn't be run automatically. This PR was opened with `ALLOW_PR_PUBLISH=1`, deferring the stage-8 review to a maintainer running the skill or re-enabling CI. Tests, lint, and security review are complete. **Do not merge without the code review.**

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.212",
+      "version": "2.2.213",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.212",
+  "version": "2.2.213",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/__tests__/runner.test.ts
+++ b/src/__tests__/runner.test.ts
@@ -502,3 +502,65 @@ describe("withCleanProcessEnv (oat01 exception wiring)", () => {
     }
   });
 });
+
+/* ───────────────────────────────────────────────────────────────────── */
+/* buildSecurityArgs — native scheduling-tool block on headless paths     */
+/* (#342: dispatch_job / any `claude -p` must not reach native cron)      */
+/* ───────────────────────────────────────────────────────────────────── */
+
+describe("buildSecurityArgs — native scheduling-tool block", () => {
+  const NATIVE = ["CronCreate", "CronDelete", "CronList", "ScheduleWakeup"];
+
+  function disallowedOf(args: string[]): string[] {
+    const i = args.indexOf("--disallowedTools");
+    return i >= 0 ? args[i + 1].split(",") : [];
+  }
+
+  it("blocks the four native tools at the default 'moderate' level", () => {
+    const args = runnerMod.buildSecurityArgs({
+      level: "moderate",
+      allowedTools: [],
+      disallowedTools: [],
+    });
+    const blocked = disallowedOf(args);
+    for (const t of NATIVE) expect(blocked).toContain(t);
+  });
+
+  it("keeps the strict denials AND adds the native tools at 'strict'", () => {
+    const args = runnerMod.buildSecurityArgs({
+      level: "strict",
+      allowedTools: [],
+      disallowedTools: [],
+    });
+    const blocked = disallowedOf(args);
+    for (const t of [...NATIVE, "Bash", "WebSearch", "WebFetch"]) {
+      expect(blocked).toContain(t);
+    }
+  });
+
+  it("merges operator-configured disallowedTools with the native block (single flag)", () => {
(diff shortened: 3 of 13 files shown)


TerrysPOV:
@claude /review

TerrysPOV:
Both findings addressed in 7e61086.

**🔴 Native-cron block missed the headless `dispatch_job` path** — correct, and it was the more dangerous case (a one-shot `claude -p` exits before any wakeup could fire). Fixed at the source rather than just `dispatch_job`: the four tool names are now a shared const (`config.ts`), and `buildSecurityArgs` always merges them into a single `--disallowedTools` — so every daemon-spawned `claude -p` path is covered, not just this one. `strict` keeps its `Bash,WebSearch,WebFetch` denials and operator-configured `disallowedTools` are merged in too (one flag, no duplicate `--disallowedTools`). `locked` already excludes them via its `--tools` allow-list. New tests in `runner.test.ts` cover moderate/strict/locked + the operator-config merge.

**🟡 TOCTOU on the duplicate-label check** — fixed with an atomic exclusive-create write (`writeFile(..., { flag: "wx" })`); the pre-check `existsSync` gap is gone, so two concurrent same-label calls can't both pass. `EEXIST` maps to the same friendly error, so the "never silently overwrites" guarantee now holds under concurrency.

Thanks for the review — the headless-path gap was a real hole in the central claim.

PR #379, opened 2026-09-07
fix(telegram): stop acking a 'discuss' tap as approved

### LINKED ISSUE #374: Telegram: a 'discuss' tap is acked as approved
Tapping **discuss** on a pending proposal answers `✅ Approuvé`.

`ackForResolution` in `src/commands/telegram.ts` handles `skip`, `reject` and `cancel` explicitly and falls through to `✅ Approuvé` for anything else. `discuss` is anything else — but it does not apply the proposal: the decision routes to the discussion handler, which opens a thread on it.

So the one tap that exists specifically to *not* commit to a proposal is the one reporting that it did. An operator who taps discuss is told the thing was approved.

```ts
const decLower = decision.toLowerCase();
if (decLower === "skip") return "⏸ Plus tard";
if (decLower === "reject" || decLower === "cancel") return "❌ Rejeté";
return "✅ Approuvé";   // ← discuss lands here
```

Small and self-contained: one branch in one function, plus a case-insensitive test per spelling (the decision arrives from a callback payload that is not normalised).

Branch: `fix/telegram-discuss-ack`

Nibbler1250:
Fixes #374.

A `discuss` tap on a pending action was acked as **approved**. The two
acknowledgement paths had drifted: `discuss` was handled in one and left lying
in the other, and an unrecognised decision fell through to the approval branch —
so the user was told their choice had been approved when it had not.

## What changed

Both paths now answer one rule, `classifyDecision`, instead of two deny-lists
that could disagree. An unrecognised decision is **reported** rather than
interpreted as an approval.

`normalizeDecision` is used both to classify and to echo, so the two paths also
render the same string: `ackForAlready` lower-cased while parsing while
`ackForResolution` echoed the raw callback value, which meant `Done` and `done`
came back differently. A mixed-case test pins that.

One existing assertion was deliberately reversed: an unknown decision used to
ack as approved, and that was the defect, not the contract.

## Verification

- `bun test src/commands/__tests__/telegram-pending-ack.test.ts` — 22 pass.
- Repo suite compared against the merge base: no failure the base did not have.
- Reviewed at the current head by two independent passes. The first found a real
  defect (the two unknown branches echoed differently-normalized values); it is
  fixed here and covered by a test. The re-review at this head found no defects.
- Live: the fix was deployed to a running daemon and a real `discuss` tap was
  acked as *sent to discussion*, not as approved.

Every decision value reaching these functions was enumerated from the code that
produces them, not from the ones the original report mentioned.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw-plus",
       "source": "./",
       "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw.",
-      "version": "2.2.238",
+      "version": "2.2.239",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw-plus",
-  "version": "2.2.238",
+  "version": "2.2.239",
   "description": "ClaudeClaw+ — governance, orchestration, persistent memory, and hardened web UI for Claude Code daemons. Sister project to moazbuilds/claudeclaw."
 }
--- a/src/commands/__tests__/telegram-pending-ack.test.ts
+++ b/src/commands/__tests__/telegram-pending-ack.test.ts
@@ -33,12 +33,50 @@ describe("ackForAlready (#314)", () => {
     );
   });
 
-  it("omits the timestamp when absent or unparseable, defaults to approved", () => {
+  it("omits the timestamp when absent or unparseable", () => {
     expect(ackForAlready("already:approve")).toBe("✅ Déjà approuvé");
     expect(ackForAlready("already:approve:not-a-date")).toBe("✅ Déjà approuvé");
-    // Unknown decision string → treated as the approve default (informative, not alarming).
+  });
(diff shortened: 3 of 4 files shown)


Nibbler1250:
Heads-up on the red check — it is not this PR's code.

`claude-review` fails at the checkout step, 3 seconds in:

> Refusing to check out fork pull request code from a 'pull_request_target' workflow. […] To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

`.github/workflows/claude-code-review.yml` triggers on `pull_request_target` and
then checks out `github.event.pull_request.head.repo.full_name` at the head sha.
GitHub's runner now refuses that combination by default, so the job dies before
it reads a single file.

This is not specific to this branch: it will fail the same way on **every** pull
request opened from a fork, which is every external contribution. Everything else
here is green.

Worth deciding deliberately rather than flipping the flag — the refusal is
pointing at a real "pwn request" shape: the job runs with the base repo's token
and secrets, and checking out fork code into that context is what the protection
exists to prevent. Two options that keep the review without the exposure:

- split it — a `pull_request` job that builds and reviews the untrusted code with
  no secrets, and a `pull_request_target` job that only posts the result; or
- keep `pull_request_target` but review the **diff** fetched through the API
  rather than checking out and executing the fork's tree.

Happy to send a PR for either if you say which shape you want.

Issue #362, opened 2026-08-29
Prompt delivery is confirmed from the rendered terminal, which cannot distinguish a submitted prompt from a repaint


Nibbler1250:
## Symptom

An auto-compaction that starts in the same second as an inbound prompt loses that
prompt. Nothing errors. The only signal is a receipt closing as `timeout` five
minutes later, and by then the user has been waiting in silence.

Latest occurrence, 2026-08-27:

```
14:02:15.474Z  prompt received
14:02:15       /compact starts — same second
14:04:43       "Compacted (ctrl+o to see full summary)" — 2m28s of compaction
14:04:47.206Z  turn_started_at  ─┐ delta 0 ms
14:04:47.206Z  stdin_written_at ─┘
               no end_turn anywhere in the session for that window
               prompt never answered
```

One compaction that day. One lost prompt.

## What the current guards do get right

The compaction latch works. The loop waited the full 2m28s without spending a
single submit nudge, and emitted no `delivery-confirm` warning. That half of
#359 behaves as designed.

## What they do not fix

Four seconds after the compaction ended, the loop concluded a turn had started.
No turn had. `turn_started_at` is stamped by the confirm loop, so its presence
records what the loop *believed*, not what happened.

The guard that should have covered this window keys on the literal `Compacted (`
still being present in the sampled tail. Once that string scrolls out of the
rolling buffer while the CLI keeps repainting the restored transcript, the next
window contains real text and no idle footer — which is exactly the shape the
loop treats as positive evidence of a turn.

This is not version drift: `to cycle`, `Compacting conversation`,
`Compacting at auto` and `Compacted` are all still present in the CLI binary
(checked against 2.1.247).

## Why this keeps recurring

Every probe in the delivery-confirm loop reads the **rendered terminal**. On that
surface, "the CLI submitted my text" and "the CLI drew something" are the same
bytes.

The consequence is that each guard trades one failure for its inverse rather than
removing the ambiguity:

| Guard | Buys | Costs |
| --- | --- | --- |
| Sticky compaction latch | waits out a spinner-only compaction | a marker in ordinary output stalls the write chain |
| Echo stripping | user text can't impersonate a status line | a prompt quoting the footer can erase the real one |
| Turn-evidence gate | no duplicate submit into a live turn | a post-compaction repaint reads as a turn and blocks the retype |

Three rounds of review (security, adversarial, Copilot) each found a real defect
[...]

Issue #239, opened 2026-06-07
bus: per-origin turn-state to fix silent-drop net interleaving on a shared agent (#217 follow-up)


Nibbler1250:
## Context

Follow-up from the #217 review (note 2, non-blocking). The silent-drop safety net keeps per-agent turn state in two single-slot maps — `currentTurnReplied` and `lastPromptOrigin` (routing) in `core.ts`.

## Problem

In the real deployment, one `default` agent serves multiple origins concurrently — telegram, webui-inject, cron/heartbeat. The single slot is last-write-wins per agent, so two prompts interleaving on the same agent race:

- A telegram prompt arriving mid-webui-turn resets the slot (and `lastPromptOrigin`), then the in-flight turn's `turn_end` synthesizes a reply routed to the wrong origin — the user can get unsolicited / misrouted ambient text.

The `CHANNEL_DRIVEN_ORIGINS` gate added in #217 (5f9e956) removes the cron/heartbeat false-positives but does **not** close the channel-vs-channel race (e.g. telegram during webui-inject). The webui-bridge mutex only serialises the webui path.

## Direction (not prescriptive)

Either serialize prompt→turn_end per agent across all channel-driven origins, or key the turn state by `origin_id` instead of a single per-agent slot so concurrent turns don't clobber each other's delivery/routing state.

Refs #217.

Nibbler1250:
@TerrysPOV — this needs your call on user-visible behaviour, not on code.

Two directions, incompatible:

- **(a) Serialize** prompt → `turn_end` per agent across all channel-driven origins. Small change; a shared agent becomes a queue, so a second chat waits for the first turn to finish — latency changes for everyone sharing an agent.
- **(b) Key the turn state by `origin_id`** (per-origin `currentTurnReplied` / `currentTurnFinalPublished` / nudge state / `lastPromptOrigin`). Concurrency stays; the #215/#217 net and every per-agent slot become maps — a wide change through `core.ts`.

#224 (merged today) added a third leg the issue did not have: a reply can name the chat it answers (`in_reply_to`), and finals are deduplicated per routed chat. With that in place, (b) is mostly about the *delivery net* (synthesizing a reply for the right chat when the agent forgets), not routing.

Queue, or concurrency? If (b), I'd take it in one PR with a live check on this host (one agent, Telegram + web UI interleaved); if (a), it's a short one.

Nibbler1250:
[...]

PR #404, opened 2026-09-15
fix(bus): surface every terminal stop_reason as turn_end, and release the operation slot once per message (#401, #405)

### LINKED ISSUE #401: bus: a turn that stops on stop_sequence/max_tokens emits no turn_end, so agentTurnActive stays set
`JsonlTailer` publishes `response.turn_end` only when an assistant line stops
with `end_turn` (`src/bus/jsonl-tailer.ts`, the `stop_reason === "end_turn"`
branch). A turn that ends with `stop_sequence` or `max_tokens` records no
`turn_end`, so `agentTurnActive` stays set until the next turn that does end
cleanly.

Everything downstream that waits on `agentTurnActive` inherits the stall:

- the deferred delivery verify from #400 waits for a `turn_end` that never
  comes, so a give-up during such a turn is re-checked only after the *next*
  completed turn;
- `mcp-reconciler` holds until the flag clears (bounded by its own timeout);
- the silent-drop net gets no terminator for that turn.

`markAgentReady` clears the flag at a session-generation boundary (#252), and
`replay_done` clears it on resume, but nothing clears it on a same-session
turn that simply stopped on another reason.

Noted as pre-existing in #400 ("Known bounds"); opening so it is not lost.
Reproduction is a transcript line with `stop_reason: "max_tokens"` followed by
no further assistant line — the flag then survives the whole idle period.

A fix is probably one of: publish `turn_end` with the actual `stop_reason` for
every terminal reason other than `tool_use` (consumers already read
`stop_reason` off the payload), or clear `agentTurnActive` on any
non-`tool_use` stop and keep `turn_end` semantics unchanged.

### LINKED ISSUE #405: bus: a multi-line assistant message repeats the terminal stop_reason on every line, so the tailer emits two turn_end (and two slot releases) per turn
`JsonlTailer.dispatchAssistant` publishes `response.turn_end` per assistant
*line*, and the CLI writes one line per content block of a message with the
message's `stop_reason` repeated on every line. A turn whose final message is
`thinking` + `text` therefore produces two `turn_end` events for one turn.

Measured on a live host across CLI 2.1.191 → 2.1.272: 1437/1437 multi-line
`end_turn` messages are `(thinking-line: end_turn, text-line: end_turn)`. Only
2.1.94 wrote `null` on the first line.

Consequences in `BusCore.ingestSessionEvent`:

- `handleTurnEnd` runs twice — harmless today (the second call sees
  `currentTurnFinalPublished`/`replyNudged`), but only by accident of ordering;
- `releaseTurn` runs twice (`core.ts`, the release after publish). The second
  release can free the slot of a *following* prompt that was counted in the
  meantime, so its events publish without a `promise_id` and its `turn_end`
  releases nothing.

The `stop_reason` widening in #401 does not add a new class of exposure (a
`max_tokens` cut on a thinking+text message repeats the same way), which is why
it is filed separately.

Fix shape: dedupe the boundary per `message.id` in `dispatchAssistant` — remember
the last message id for which a `turn_end` was published and skip a second line
carrying the same id — and pin it with a tailer test using the two-line shape
above. The text payload should then be the concatenation across the message's
lines, not of the last line only, which is a second, smaller defect of the
per-line emission.

Nibbler1250:
Closes #401. Closes #405.

## What

`JsonlTailer` published `response.turn_end` only when an assistant line stopped with `end_turn`. A turn that stopped on any other terminal reason (`max_tokens`, `stop_sequence`, …) produced no terminator, so in bus core `agentTurnActive` stayed set until the next clean turn — deferring the #400 delivery verify, holding the reconciler, and starving the silent-drop net.

The tailer now emits `response.turn_end` for every terminal `stop_reason`, carrying the real reason in the payload. `tool_use` and `pause_turn` resume the turn and stay silent; a null reason is a content-block line still streaming.

## The CLI's synthetic lines

On live transcripts every `stop_sequence` line is one the CLI wrote itself (`model: "<synthetic>"`), in two shapes:

- **API error** (`isApiErrorMessage: true` — "Prompt is too long …", "You've hit your session limit …", always followed by `last-prompt`): the turn *is* over. Surfaced as a boundary with `text: ""` and `synthetic: true`, so the flag and the operation slot release but the #215 net neither nudges an agent that cannot answer nor ships the raw error string as a reply — the error is already published as `system.api_error`, which now goes out *before* the boundary so it keeps the turn's `promise_id`.
- **Placeholder** ("No response requested.", no error): written for an SDK-enqueued prompt *before* that prompt's real turn starts. Not a boundary — emitting one would nudge the agent and free the slot while the real turn is about to begin. Skipped, as before.

Consumers (`core` flag + slot release, `runtime-mount` ×2, `stall-watchdog`, webui) key on the topic, not the reason, so their behaviour on the new boundaries is exactly their `end_turn` behaviour. `core.ts` changes are comments only.

## Evidence

- Replay of a real transcript (652 assistant lines) through the tailer, base vs branch: base emits 154 `end_turn` boundaries and misses the 2 API-error `stop_sequence` ones; branch emits 154 + 2 and ignores the 49 placeholders.
- Branch tailer run read-only against the live agent's active transcript for 3 min (file growing, daemon untouched): 2425 events, 0 malformed, 34/34 `end_turn` once each, 345 `tool_use` silent.
- Corpus check (1383 transcripts): the only stop reasons that occur are `tool_use`, `end_turn`, `stop_sequence`; every synthetic line is `stop_sequence`; no synthetic error is ever followed by a real-model line in the same turn (the CLI's internal retries write no line), so the early release is correct on all observed data.

## Tests

Tailer: emits on `max_tokens` / `stop_sequence` with the real reason; empty synthetic boundary for an API-error line with `system.api_error` ordered first; silent on the placeholder, on `pause_turn`, on a null reason. Core: contract pin that the flag clears on a `turn_end` whatever its reason.

## #405 — one slot release per message

The CLI repeats the terminal `stop_reason` on every content-block line of a message, so a thinking + text message already ended a turn twice on the base (1437/1437 multi-line `end_turn` messages across CLI 2.1.191→2.1.272), and core released the operation slot twice — the second release could free the slot of a prompt counted in between. Copilot flagged that the widening adds the `max_tokens` shape to that exposure, so it is closed here rather than left to a follow-up: every boundary now carries `message_id`, core releases on the first boundary of a message and ignores further lines of it. Both lines still reach the #215 net (the text line is the one it needs); a boundary without an id releases as before.

## Reviewer notes

`SCHEMA_VERSION` (tailer) is deliberately not bumped: it keys the schema-probe cache that re-validates *transcript line parsing* against the installed CLI, and this change alters what the tailer emits, not how it parses lines — the same reading as the five tailer changes since Sprint 2 (#215, #362, …), none of which bumped it.

Known and unchanged: if a prompt B is counted between the two boundary lines of A's message (they are milliseconds apart), A's second line is handled against B's per-turn flags/origin and stamped with B's id — the per-origin turn-state residual tracked in #239, identical on the base. This PR narrows it (the slot is no longer released twice) and does not widen it.

Version bump: 2.2.250 → 2.2.251.

## Review trail

Reviews this change went through before being marked ready (the SDC gate records each one against the exact commit under review):

- **Scope tests, lint and type delta** vs the merge base (T1).
- **Full suite ×3** vs the merge base, plus a live proof on the deployed host (T2).
- **Two-axis self review** (correctness, blast radius).
- **Security review** (new input surfaces, I/O, credentials).
- **Adversarial review** by independent agents (2 passes), and GitHub Copilot on 2 revisions.
- **Automated review** (Copilot on this PR, CodeRabbit on a mirror) on the final commit.
- **Live test** on the deployed host.

What those reviews changed, in one breath: the first adversarial pass split the CLI's synthetic lines (API errors become empty boundaries, placeholders stay silent) and the second reordered `system.api_error` ahead of the boundary so it keeps its `promise_id`; Copilot asked for the multi-line duplicate (#405) to be closed in the same PR, which is where the once-per-message slot release came from.


Nibbler1250:
@claude review

